PR #66, opened 2024-12-24
SQL Parser Test and functions list

Vedin:
Test for parser and preparation for end-to-end testing with SQL queries


Vedin:
close due to it not be usable in future

Issue #68, opened 2024-12-26
Add dateadd function  


osipovartem:
New function should support this signature
```sql
DATEADD( <date_or_time_part>, <value>, <date_or_time_expr> )
```
Detailed description here 
https://docs.snowflake.com/en/sql-reference/functions/dateadd
Basic implementation can be taken from
https://github.com/apache/datafusion/tree/main/datafusion/functions/src/datetime

DanCodedThis:
I looked into date_trunc, it cleared some things up. Still looking for the "addition op" as it was in this PR in datetime.rs line 202: https://github.com/apache/datafusion/pull/3110/files#diff-6acdb260dee24439d23b730937a21c579000adc11798f06b9d4517e2467c5dcb

Later removed here: https://github.com/apache/datafusion/pull/3153

I have a question, will we have a problem supporting "weeks", since I don't see a type for this here: https://docs.rs/datafusion/latest/datafusion/common/arrow/datatypes/enum.IntervalUnit.html

"Year" is supported through some internal conversion to MonthDayNano, if I'm not mistaken.

DanCodedThis:
With the last commit, I can say that if were to use the internal types, we need to extend the sql parser for this query "select dateadd('\''0 days'\'', 3,'\''2024-12-26'\'')" (now it works like this "3 days" => IntervalMonthDayNano(3)):

1. "days" => IntervalMonthDayNano(0), this one can be changed to use utf8 without sql parser extension (but it is probably right to extend specifically the parser)
2. "days, 3" =>  IntervalMonthDayNano(3) - may bring many horrible problems

Why do we want to use internal types? - Because the addition is built in, we can just add most of the variations and it will work out of the box

DanCodedThis:
Needs testing, seems to work

PR #70, opened 2024-12-27
least (copy), greatest (copy) and dateadd mock

osipovartem:
- Copied **least** and **greatest** UDF functions from this [PR](https://github.com/apache/datafusion/pull/13786) (already merged, but not released). After datafusion release we should remove this local copy
- Added mock for **dateadd** function to support sql syntax
- Changed logic for datafusion -> snowflake timestamps convertor for a while (still receive the error _252005: Failed to convert: field min_last_success: TIMESTAMP_NTZ::min_last_success, Error: invalid literal for int() with base 10: 'min_last_success'_)

### DIFF
--- a/crates/control_plane/src/sql/functions/common.rs
+++ b/crates/control_plane/src/sql/functions/common.rs
@@ -119,7 +119,7 @@ fn convert_timestamp_to_struct(column: &ArrayRef, unit: &TimeUnit) -> ArrayRef {
                 .unwrap();
             let epoch: Int64Array = array
                 .iter()
-                .map(|x| x.unwrap_or_else(now) / 1_000_000_000)
+                .map(|x| x.unwrap_or(now) / 1_000_000_000)
                 .collect();
             let fraction: Int32Array = array
                 .iter()
--- a/crates/control_plane/src/sql/functions/common.rs
+++ b/crates/control_plane/src/sql/functions/common.rs
@@ -1,10 +1,8 @@
 use crate::models::ColumnInfo;
-use arrow::array::{
-    Array, Int32Array, Int64Array, StructArray, TimestampMicrosecondArray,
-    TimestampMillisecondArray, TimestampNanosecondArray, TimestampSecondArray, UnionArray,
-};
+use arrow::array::{Array, Int32Array, Int64Array, TimestampMicrosecondArray, TimestampMillisecondArray, TimestampNanosecondArray, TimestampSecondArray, UnionArray};
 use arrow::datatypes::{Field, Schema, TimeUnit};
 use arrow::record_batch::RecordBatch;
+use chrono::Utc;
 use datafusion::arrow::array::ArrayRef;
 use datafusion::arrow::datatypes::DataType;
 use datafusion::common::Result as DataFusionResult;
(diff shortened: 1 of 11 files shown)


DanCodedThis:
For dateadd we can either use utf8 or it's variants or extend the sql parser to actualy work with underlying types, but the addition operator should work out of the box, more info in issue #68 (keep in mind there are two branhces now)

DanCodedThis:
Also when using nanoseconds, maybe I'm using to small of a value but I can't seem to see the change, or is it a part of that "thing" where timestamps don't work with them yet?

DanCodedThis:
And us (microsenconds) also

crates/control_plane/src/sql/functions/common.rs
```diff
@@ -81,13 +77,14 @@ pub fn convert_record_batches(
...
 fn convert_timestamp_to_struct(column: &ArrayRef, unit: &TimeUnit) -> ArrayRef {
+    let now = Utc::now().timestamp();
     let (epoch, fraction) = match unit {
         TimeUnit::Second => {
             let array = column
                 .as_any()
                 .downcast_ref::<TimestampSecondArray>()
                 .unwrap();
-            let epoch: Int64Array = array.clone().unary(|x| x);
+            let epoch: Int64Array = array.iter().map(|x| x.unwrap_or(now)).collect();
```
eadgbear: You might be able to use `map_or` here

eadgbear (dismissed):

crates/control_plane/src/sql/functions/common.rs
DanCodedThis: I'll look into it on Monday or Artem will

rampage644:
Closes #75

DanCodedThis:
i64 max is 9_223_372_036_854_775_807, and since we are adding nanoseconds even with years, if somebody decides to add 30 years to any timestamp/date/time we will have an overflow (in the value that is being added, not the timestamp), we are flying dangerously close to the sun for my liking, or are we?

I initially decided not to use MonthDayNano since days and month are i32, and even though we distinguish years, month, weeks, etc., anybody by mistake can write any value that may be bigger than i32 max, accidentally (since we always receive i64, ex.: was writing nanoseconds switched the form ns to yy) or is it just an edge case?

Plus the other problem is leap years, and "leap months", which should be handled internally by timestamps (didn't check yet, but it should be common sense), but because we are adding nanoseconds, our months are always 30 days and our years are always 365 days. 

I'm going to look into this, and after that the PR should be ready

DanCodedThis (approved):
Dateadd completed

DanCodedThis (dismissed):
Dateadd completed

crates/control_plane/src/sql/functions/common.rs
```diff
@@ -81,60 +82,116 @@ pub fn convert_record_batches(
...
-        }
+    let now = Utc::now();
+    let timestamps: Vec<_> = match unit {
+        TimeUnit::Second => column
+            .as_any()
+            .downcast_ref::<TimestampSecondArray>()
+            .unwrap()
+            .iter()
+            .map(|x| {
+                let ts = DateTime::from_timestamp(x.unwrap_or(now.timestamp()), 0).unwrap();
```
eadgbear, line 93: Is now a reasonable default if `x` can't be unwrapped?

eadgbear (approved):

crates/control_plane/src/sql/functions/common.rs
DanCodedThis: Maybe zero the unix 1970 timestamp?

DanCodedThis: For now let it be. Can't merge the base branch is protected

PR #106, opened 2025-01-03
Fix tables relations

osipovartem:
- added convert_timezone stub to unblock sql queries
- added additional logic to fix table names
```sql
CREATE OR REPLACE TABLE snowplow_web_base_sessions_lifecycle_manifest AS (SELECT * FROM (WITH new_events_session_ids_init AS (SELECT COALESCE(e.domain_sessionid, NULL) AS session_identifier, max(COALESCE(e.domain_userid, NULL)) AS user_identifier, min(collector_tstamp) AS start_tstamp, max(collector_tstamp) AS end_tstamp FROM embucket.datasets.events AS e WHERE dvce_sent_tstamp <= dateadd('day', 3, dvce_created_tstamp) AND collector_tstamp >= CAST('2022-08-19 00:00:00' AS TIMESTAMP) AND collector_tstamp <= CAST('2022-09-18 00:00:00' AS TIMESTAMP) AND true AND true GROUP BY 1), new_events_session_ids AS (SELECT * FROM new_events_session_ids_init AS e WHERE session_identifier IS NOT NULL AND NOT EXISTS (SELECT 1 FROM datasets.public_snowplow_manifest.snowplow_web_base_quarantined_sessions AS a WHERE a.session_identifier = e.session_identifier)), session_lifecycle AS (SELECT * FROM new_events_session_ids) SELECT sl.session_identifier, sl.user_identifier, sl.start_tstamp, least(dateadd('day', 3, sl.start_tstamp), sl.end_tstamp) AS end_tstamp FROM session_lifecycle AS sl) ORDER BY (to_date(start_tstamp)))
```
table identifier should be fixed from 
datasets.public_snowplow_manifest.snowplow_web_base_quarantined_sessions to **embucket.datasets.public_snowplow_manifest.snowplow_web_base_quarantined_sessions**
- disabled size limit for CSV files for upload

### DIFF
--- a/crates/control_plane/src/sql/sql.rs
+++ b/crates/control_plane/src/sql/sql.rs
@@ -85,7 +85,6 @@ impl SqlExecutor {
         let re = regex::Regex::new(r"(\w+)\[(\d+)]\.(\w+)").unwrap();
         let date_add = regex::Regex::new(r"(date|time|timestamp)(_?add)\(\s*([a-zA-Z]+),").unwrap();
 
-        // Replace what is not supported by sql parser
         let query = re
             .replace_all(query, "json_get(json_get($1, $2), '$3')")
             .to_string();
--- a/crates/control_plane/src/sql/sql.rs
+++ b/crates/control_plane/src/sql/sql.rs
@@ -1,5 +1,6 @@
 use crate::models::created_entity_response;
 use crate::sql::context::CustomContextProvider;
+use crate::sql::functions::convert_timezone::ConvertTimezoneFunc;
 use crate::sql::functions::date_add::DateAddFunc;
 use crate::sql::functions::greatest::GreatestFunc;
 use crate::sql::functions::least::LeastFunc;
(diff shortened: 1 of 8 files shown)


crates/control_plane/src/sql/sql.rs
```diff
@@ -390,7 +406,9 @@ impl SqlExecutor {
         mut table_name: Vec<Ident>,
         warehouse_name: &str,
     ) -> Vec<Ident> {
-        if warehouse_name.len() > 0 && !table_name.starts_with(&[Ident::new(warehouse_name)]) {
+        if warehouse_name.len() > 0
+            && !table_name.starts_with(&[Ident::new(warehouse_name)])
+            && table_name.len() > 1 {
```
Vedin, line 411: Are you sure that we won't face the situation where we need insert warehouse_name even before names without dots?

eadgbear (approved):
LGTM

crates/control_plane/src/sql/sql.rs
osipovartem: warehouse_name is required only if we specify the schema (namespace)

osipovartem: let's merge and see in future

PR #110, opened 2025-01-04
Fix schema builder after iceberg-rust version update

Vedin:
After updating iceberg-rust it no longer provides field id internally. This fixes a current broken behavior.

### DIFF
--- a/crates/control_plane/src/sql/sql.rs
+++ b/crates/control_plane/src/sql/sql.rs
@@ -30,6 +30,8 @@ use regex;
 use std::collections::hash_map::Entry;
 use std::collections::HashMap;
 use std::sync::Arc;
+use datafusion::error::DataFusionError;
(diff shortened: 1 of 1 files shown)


crates/control_plane/src/sql/sql.rs
```diff
@@ -132,10 +134,13 @@ impl SqlExecutor {
                 .await?
                 .collect()
                 .await?;
+            let fields_with_ids = StructType::try_from(&new_fields_with_ids(
+                plan.schema().as_arrow().fields(),
+                &mut 0)).map_err(|err| DataFusionError::External(Box::new(err)))?;
```
Vedin, line 139: I'm not sure about usage of `DataFusionError::External` here. Any suggestions?

eadgbear: It's our best option for now

osipovartem (approved):

PR #116, opened 2025-01-07
External table

osipovartem:
- added logic to create iceberg table with Iceberg Extension

### DIFF
--- a/crates/control_plane/src/service.rs
+++ b/crates/control_plane/src/service.rs
@@ -10,7 +10,6 @@ use arrow_json::WriterBuilder;
 use async_trait::async_trait;
 use base64::Engine;
 use bytes::Bytes;
-use datafusion::dataframe::DataFrameWriteOptions;
 use datafusion::execution::context::SessionContext;
 use datafusion::execution::SessionStateBuilder;
 use datafusion::prelude::{CsvReadOptions, SessionConfig};
@@ -18,12 +17,13 @@ use datafusion_iceberg::catalog::catalog::IcebergCatalog;
 use datafusion_iceberg::planner::IcebergQueryPlanner;
 use iceberg_rest_catalog::apis::configuration::Configuration;
 use iceberg_rest_catalog::catalog::RestCatalog;
-use object_store::local::LocalFileSystem;
+use icelake::TableIdentifier;
 use object_store::path::Path;
 use object_store::{ObjectStore, PutPayload};
 use rusoto_core::{HttpClient, Region};
 use rusoto_credential::StaticProvider;
 use rusoto_s3::{GetBucketAclRequest, S3Client, S3};
+use std::collections::HashMap;
 use std::sync::Arc;
 use url::Url;
 use uuid::Uuid;
@@ -232,7 +232,7 @@ impl ControlService for ControlServiceImpl {
         }
 
         let records: Vec<RecordBatch> = SqlExecutor::new(ctx)
-            .query(query, &catalog_name.clone().to_string())
+            .query(query, &warehouse)
             .await?
             .into_iter()
             .collect::<Vec<_>>();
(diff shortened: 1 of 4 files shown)


crates/control_plane/src/service.rs
```diff
@@ -207,8 +209,13 @@ impl ControlService for ControlServiceImpl {
             storage_profile.get_object_store_builder(),
         );
         let catalog = IcebergCatalog::new(Arc::new(rest_client), None).await?;
-        let ctx =
-            SessionContext::new_with_config(SessionConfig::new().with_information_schema(true));
+        let state = SessionStateBuilder::new()
+            .with_config(SessionConfig::new().with_information_schema(true))
+            .with_default_features()
```
DanCodedThis, line 214: Not important, but may not need ".with_default_features()", since in session state builder it is stated: "A builder to be used for building [SessionState](https://docs.rs/datafusion/latest/datafusion/execution/session_state/struct.SessionState.html)’s. Defaults will be used for all values unless explicitly provided." from: https://docs.rs/datafusion/latest/datafusion/execution/session_state/struct.SessionStateBuilder.html

DanCodedThis (commented):
Building on wsl ubuntu, to check explicitly, but otherwise looks good

DanCodedThis (approved):
Checked on my machine with wsl ubuntu, looks good

crates/control_plane/src/sql/planner.rs
```diff
@@ -60,7 +64,14 @@ where
...
             provider,
-            normalizer: Default::default(),
+            ident_normalizer: Default::default(),
+        }
+    }
+
+    pub fn statement_to_plan(&self, statement: DFStatement) -> Result<LogicalPlan> {
+        match statement {
+            DFStatement::Statement(s) => self.sql_statement_to_plan(*s),
+            _ => self.inner.statement_to_plan(statement),
```
Vedin, line 74: Can you please explain logic here? My understanding of this is that we need to distinguish statement types and can handle fully pattern match here. As I can see DFStatment is 
```rust
pub enum Statement {
    /// ANSI SQL AST node (from sqlparser-rs)
    Statement(Box<SQLStatement>),
    /// Extension: `CREATE EXTERNAL TABLE`
    CreateExternalTable(CreateExternalTable),
    /// Extension: `COPY TO`
    CopyTo(CopyToStatement),
    /// EXPLAIN for extensions
    Explain(ExplainStatement),
}
```
So why do we call self.inner.statement_to_plan for CreateExternalTable?

crates/control_plane/src/sql/sql.rs
```diff
@@ -103,21 +100,29 @@ impl SqlExecutor {
     pub async fn create_table_query(
         &self,
         statement: Statement,
-        warehouse_name: &String,
+        warehouse: &Warehouse,
     ) -> Result<Vec<RecordBatch>> {
         if let Statement::CreateTable(create_table_statement) = statement {
+            let warehouse_name = &*warehouse.name;
```
Vedin, line 106: nit: This `&*` looks a little bit weird to me. I don't know the right way to do this. The first thought is to use something like `(&warehouse.name).as_str()` if you want to have `&str` here to make it obvious why we want to extract the reference and take a reference at all.

crates/control_plane/src/sql/sql.rs
```diff
@@ -128,78 +133,58 @@ impl SqlExecutor {
...
+            // Create External table
+            match plan {
+                LogicalPlan::Ddl(DdlStatement::CreateMemoryTable(table)) => {
+                    let external_table_plan = LogicalPlan::Ddl(DdlStatement::CreateExternalTable(
+                        PlanCreateExternalTable {
+                            schema: table.input.schema().clone(),
+                            name: new_table_ref,
+                            location, // Specify the external location
+                            file_type: "ICEBERG".to_string(), // Specify the file type
+                            table_partition_cols: vec![], // Specify partition columns if any
```
Vedin, line 156: Not a part of this task but want to mention that we want to support this table_partition_cols in the nearest future.

crates/control_plane/src/sql/sql.rs
```diff
@@ -128,78 +133,58 @@ impl SqlExecutor {
...
                 .await?;
-            // self.ctx.sql(&insert_query).await?.collect().await?;
 
             // Drop InMemory table
             let drop_query = format!("DROP TABLE {new_table_name}");
             self.ctx.sql(&drop_query).await?.collect().await?;
-            return Ok(result);
-            // }
-            // Ok(created_entity_response())
+            Ok(created_entity_response())
```
Vedin, line 187: Do we need to keep `result` variable after this change? What is the purpose of having this separate function `created_entity_response` instead of using `result`? What happens if `self
                .execute_with_custom_plan(&insert_query, warehouse_name)
                .await?;` returns Err instead of Ok?

Vedin (approved):
lgtm

crates/control_plane/src/sql/planner.rs
osipovartem:
```rust
 /// Extension: `CREATE EXTERNAL TABLE`
    CreateExternalTable(CreateExternalTable),
```
CreateExternalTable means CREATE EXTERNAL TABLE, but we don't send such statement
We have only InMemory tables and then create LogicalPlan manually from InMemory table plan (copy schema, constrains, etc.))

crates/control_plane/src/sql/sql.rs
osipovartem: can be replaces just with warehouse.name.as_str()

crates/control_plane/src/sql/sql.rs
osipovartem: In case of any error we send this error into response. 
created_entity_response() is just a usual response for any table creation
`result` in this case can be different since we send INSERT instead of original CREATE TABLE

eadgbear (approved):

Issue #146, opened 2025-01-15
engine: support alter table statement


osipovartem:
Example
```sql
alter table datasets.public_snowplow_manifest.snowplow_web_base_sessions_lifecycle_manifest cluster by (to_date(start_tstamp));
```

PR #147, opened 2025-01-16
Implement MERGE INTO

### LINKED ISSUE #74: engine: Implement MERGE INTO statement

osipovartem:
Closes #74 
- Parse MERGE INTO statement as insert as select
- Fallback to NOT MATCH part as default insert
- Skip ALTER TABLE. The implementation will be added in a separate issue - #146
- removed some funcs which were copied from datafusion
- fixed query processing regex for **json_get**
- added full error message into response


crates/nexus/src/http/dbt/error.rs
```diff
@@ -57,9 +57,24 @@ impl IntoResponse for DbtError {
...
+            Self::RowParse { source } => format!("failed to parse row JSON: {source}"),
+            Self::MissingAuthToken | Self::MissingDbtSession | Self::InvalidAuthData => {
+                "session error".to_string()
+            }
+            Self::NotImplemented => "feature not implemented".to_string(),
+        };
+
         let body = Json(JsonResponse {
             success: false,
-            message: Some(self.to_string()),
```
eadgbear, line 62: `to_string` should be returning the display message for snafu, are we not getting the strings from that?

crates/nexus/src/main.rs
```diff
@@ -192,6 +192,7 @@ async fn print_request_response(
     Ok(res)
 }
 
+#[allow(clippy::future_not_send)]
```
eadgbear: Let's examine why the future is not `Send` before allowing this rule

crates/nexus/src/http/dbt/error.rs
osipovartem: We receive only "Internal Server Error" as a message text with current implementation
Looks like a status_code.to_string()

crates/nexus/src/main.rs
osipovartem: The issue was with missing Send trait
```rust
async fn buffer_and_print<B>(
    direction: &str,
    method: &String,
    uri: &String,
    body: B,
) -> Result<Bytes, (StatusCode, String)>
where
    B: axum::body::HttpBody<Data = Bytes> **+ Send**,
    B::Error: std::fmt::Display,
{
```

DanCodedThis (dismissed):
Looks good to me

Vedin (approved):
Let's merge it as is and create follow-up issues.

PR #150, opened 2025-01-16
Adding Snowflake syntax tests

eadgbear:
These tests are built using `insta`, a crate for snapshot testing.

For each query in the queries file, the query is:
* Parsed into an AST
* Turned into a Logical Plan
* Executed

The resulting snapshot is the results of each of those steps. The results are not validated, just that the query runs. These tests will tell us if we can parse certain Snowflake syntax and execute it.

To run these tests, simply run `cargo test` in the `runtime` crate. Then use `cargo insta review` to go over the diffs of the results. Not all the queries are deterministic yet so there may be some changes each time the tests are run.


osipovartem (approved):

rampage644:
@slyons Planning to merge?

PR #151, opened 2025-01-17
Implement QUAILFY

### LINKED ISSUE #78: engine: Implement `qualify` syntax

osipovartem:
Closes Embucket/control-plane-v2#78
- Added missing datatypes
-


crates/runtime/src/datafusion/execution.rs
```diff
@@ -509,6 +526,94 @@ impl SqlExecutor {
...
+        if let Some(with) = query.with.as_mut() {
+            for cte in &mut with.cte_tables {
+                self.update_qualify_in_query(&mut cte.query);
+            }
+        }
+
+        match query.body.as_mut() {
+            sqlparser::ast::SetExpr::Select(select) => {
+                if let Some(Expr::BinaryOp { left, op, right }) = select.qualify.as_ref() {
+                    if matches!(op, BinaryOperator::Eq) {
```
Vedin: Is it intended that we only support `BinaryOperator::Eq` here? I'm totally fine with this for the current solution.
In general, it can be any predicate. What do you think about extending it with some other logical operators? At least, `Lt` and `LtEq` can be useful for various use cases.

osipovartem: added

Vedin (approved):

PR #164, opened 2025-01-23
Basic support for create database

osipovartem:
Closes to Embucket/control-plane-v2#160

### DIFF
--- a/crates/catalog/src/models.rs
+++ b/crates/catalog/src/models.rs
@@ -193,7 +193,7 @@ pub struct DatabaseIdent {
 
 impl Display for DatabaseIdent {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}.{}", self.warehouse, self.namespace.to_url_string())
+        write!(f, "{}.{}", self.warehouse, self.namespace.join("."))
     }
 }
 
--- a/crates/runtime/src/datafusion/execution.rs
+++ b/crates/runtime/src/datafusion/execution.rs
@@ -83,7 +83,8 @@ impl SqlExecutor {
                 | Statement::Commit { .. }
                 | Statement::Insert { .. }
                 | Statement::ShowSchemas { .. }
-                | Statement::ShowVariable { .. } => {
+                | Statement::ShowVariable { .. }
+                | Statement::Update { .. } => {
                     return Box::pin(self.execute_with_custom_plan(&query, warehouse_name)).await;
                 }
                 Statement::Query(mut subquery) => {
@@ -291,8 +292,8 @@ impl SqlExecutor {
             self.update_tables_in_table_factor(&mut table, warehouse_name);
             self.update_tables_in_table_factor(&mut source, warehouse_name);
 
-            let (target_table, target_alias) = self.get_table_with_alias(table);
-            let (source_table, _source_alias) = self.get_table_with_alias(source.clone());
+            let (target_table, target_alias) = Self::get_table_with_alias(table);
+            let (_source_table, source_alias) = Self::get_table_with_alias(source.clone());
 
             let source_query = if let TableFactor::Derived {
                 subquery,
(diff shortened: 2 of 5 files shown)


DanCodedThis (dismissed):

crates/runtime/src/datafusion/execution.rs
```diff
@@ -437,6 +423,48 @@ impl SqlExecutor {
...
+        // Why are we checking if namespace exists as a single layer and then creating it?
+        // There are multiple possible Error scenarios here that are not handled
+        let should_create = if if_not_exists {
+            rest_catalog.load_namespace(&namespace).await.is_err()
+        } else {
+            true
+        };
+        if should_create {
+            #[allow(clippy::unwrap_used)]
+            let namespace = Namespace::try_new(&namespace_vec).unwrap();
```
eadgbear: Why are we using the vectorized namespace here instead of the `single_layer_namespace`?

eadgbear (dismissed):

crates/runtime/src/datafusion/execution.rs
DanCodedThis: > Why are we using the vectorized namespace here instead of the `single_layer_namespace`?

Was actually thinking the same

osipovartem: Single layer contains "." between layers, without it the namespace will be created as
db.warehouse_id.datasetssnowplow_web in catalog (instead of db.warehouse_id.datasets.snowplow_web
the key will be wrong since we use 
```rust
 write!(f, "{}.{}", self.warehouse, self.namespace.to_url_string())
 ```
 to format the key where to_url_string use **Unicode Character 'INFORMATION SEPARATOR ONE' (U+001F)** as a separator to split layers;
 
 **load_namespace** use under the hood use 
 ```rust
 let response = catalog_api_api::load_namespace_metadata(
    &self.configuration,
    self.name.as_deref(),
    &namespace.url_encode(),
)
```
but **create_namespace** with iceberg REST catalog use 
```rust
let response = catalog_api_api::create_namespace(
    &self.configuration,
    self.name.as_deref(),
    models::CreateNamespaceRequest {
        namespace: namespace.to_vec(),
        properties,
    },
)
```

Let me fix this

crates/catalog/src/models.rs
```diff
@@ -193,7 +193,7 @@ pub struct DatabaseIdent {
 
 impl Display for DatabaseIdent {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}.{}", self.warehouse, self.namespace.to_url_string())
+        write!(f, "{}.{}", self.warehouse, self.namespace.join("."))
```
osipovartem, line 196: FYI @slyons @DanCodedThis 
very important fix
databases will be saved with the key
`db.wh_uuid.db_layer1.db_layer2`
instead of
`db.wh_uuid.db_layer1\u001Fdb_layer2`
keys in the database should not depend on the iceberg catalog format

DanCodedThis: Great

DanCodedThis (dismissed):

crates/runtime/src/datafusion/execution.rs
```diff
@@ -389,42 +403,14 @@ impl SqlExecutor {
...
-                // There are multiple possible Error scenarios here that are not handled
-                if rest_catalog.load_namespace(&namespace).await.is_err() {
-                    #[allow(clippy::unwrap_used)]
-                    let namespace = Namespace::try_new(&namespace_vec).unwrap();
-                    rest_catalog
-                        .create_namespace(&namespace, None)
-                        .await
-                        .context(ih_error::IcebergSnafu)?;
-                }
+                self.create_database(warehouse_name, schema_name.clone(), if_not_exists)
```
rampage644, line 418: Is it just a typo and it should be `create_schema`?

DanCodedThis: Aren't `create_database` synonymous to `create_schema` (since `create_schema` internally calls `create_database`)?

DanCodedThis: > Is it just a typo and it should be `create_schema`?

Otherwise, it will just be a stack overflow recursion, since we'll be calling `create_schema` from `create_schema` forever, without an exit condition, if the signatures were the same

osipovartem: There's no typo, create schema calls fn to create database, the logic is the same for now

Vedin (dismissed):

DanCodedThis (approved):

PR #167, opened 2025-01-27
Handle result_scan

### LINKED ISSUE #161: engine: Implement workaround for `table(result_scan(last_query_id(-1)))` query
This seems a way for dbt-snowflake adapter to get information if certain table in a schema are Iceberg or not. Let's avoid supporting this in full and instead return a result of `show objects in SNOWPLOW_ICEBERG.PUBLIC_snowplow_manifest limit 10000;` (or whatever previous query is).

Consult `snowplow_web_amalgamation_workload_iceberg.test_slow` sql test in #Embucket/compatibility-test-suite

Perhaps return this `is_iceberg` table for both queries.

**last_query_id** - Returns the ID of a specified query in the current session. If no query is specified, the most recently-executed query is returned.
**result_scan** - Returns the result set of a previous command (within 24 hours of when you executed the query) as if the result was a table
**table** - create a table from result_scan

osipovartem:
Closes Embucket/control-plane-v2#161

Replace 
```sql
select all_objects.*, is_iceberg
      from table(result_scan(last_query_id(-1))) all_objects
      left join INFORMATION_SCHEMA.tables as all_tables
        on all_tables.table_name = all_objects."name"
        and all_tables.table_schema = all_objects."schema_name"
        and all_tables.table_catalog = all_objects."database_name";
```
by 
```sql
SELECT
   all_objects.* 
FROM
   (
      SELECT
         table_catalog AS 'database_name',
         table_schema AS 'schema_name',
         table_name AS 'name',
         CASE
            WHEN
               table_type = 'BASE TABLE' 
            THEN
               'TABLE' 
            ELSE
               table_type 
         END
         AS 'kind', NULL AS 'comment', 
         CASE
            WHEN
               table_type = 'BASE TABLE' 
            THEN
               true 
            ELSE
               false 
         END
         AS is_iceberg 
      FROM
         embucket.information_schema.tables
   )
   AS all_objects 
   LEFT JOIN
      embucket.INFORMATION_SCHEMA.tables AS all_tables 
      ON all_tables.table_name = all_objects.\"name\" 
      AND all_tables.table_schema = all_objects.\"schema_name\" 
      AND all_tables.table_catalog = all_objects.\"database_name\"
```

### DIFF
--- a/crates/runtime/src/datafusion/execution.rs
+++ b/crates/runtime/src/datafusion/execution.rs
@@ -83,7 +83,8 @@ impl SqlExecutor {
                 | Statement::Commit { .. }
                 | Statement::Insert { .. }
                 | Statement::ShowSchemas { .. }
-                | Statement::ShowVariable { .. } => {
+                | Statement::ShowVariable { .. }
+                | Statement::Update { .. } => {
                     return Box::pin(self.execute_with_custom_plan(&query, warehouse_name)).await;
                 }
                 Statement::Query(mut subquery) => {
@@ -291,8 +292,8 @@ impl SqlExecutor {
             self.update_tables_in_table_factor(&mut table, warehouse_name);
             self.update_tables_in_table_factor(&mut source, warehouse_name);
 
-            let (target_table, target_alias) = self.get_table_with_alias(table);
-            let (source_table, _source_alias) = self.get_table_with_alias(source.clone());
+            let (target_table, target_alias) = Self::get_table_with_alias(table);
+            let (_source_table, source_alias) = Self::get_table_with_alias(source.clone());
 
             let source_query = if let TableFactor::Derived {
                 subquery,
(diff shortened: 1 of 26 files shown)


crates/runtime/src/datafusion/execution.rs
```diff
@@ -634,6 +635,61 @@ impl SqlExecutor {
...
+            select.projection.retain(|field| {
+                if let SelectItem::UnnamedExpr(Expr::Identifier(ident)) = field {
+                    ident.value.to_lowercase() != "is_iceberg"
+                } else {
+                    true
+                }
+            });
+
+            // Replace result_scan with the select from information_schema.tables
+            for table_with_joins in &mut select.from {
```
DanCodedThis, line 660: How many cycles is the loop?

DanCodedThis (dismissed):
LGTM, only thing that caught my eye is the loop

crates/runtime/src/datafusion/execution.rs
osipovartem: there is a break for the first catch (usually 1-2)

eadgbear (dismissed):

crates/runtime/src/datafusion/execution.rs
```diff
@@ -634,6 +635,61 @@ impl SqlExecutor {
...
+                if let TableFactor::TableFunction {
+                    expr: Expr::Function(f),
+                    alias,
+                } = &mut table_with_joins.relation
+                {
+                    if f.name.to_string().to_lowercase() == "result_scan" {
```
rampage644, line 666: Can we short circuit the code and return "not supported" error if table function is now what we expect?
Like
```rust
if f.name.to_string().to_lowercase() != result_scan {
  return plan_err("blah-blah");
}
// continue with happy path flow

osipovartem: We update query in place, we cannot return anything here
Also this is a specific check for only result_scan func (rare case) and we have to skip all other Query statements (any select)

DanCodedThis (approved):

PR #168, opened 2025-01-27
Use snowflake dialect from env

### LINKED ISSUE #166: sqlparser: TokenizerError error when trying to insert records with single quote
Rows example
```csv
Korea (Democratic People's Republic of)
Côte d'Ivoire
Lao People's Democratic Republic
```
dbt preparing insert in format 

```sql
create table datasets.test (c varchar);
insert datasets.test (c) values('Lao People\'s Democratic Republic');
```
Last statement triggers 
```
Execution error: DataFusion error: SQL error: TokenizerError("Unterminated string literal at Line: 1, Column: 67"
```

osipovartem:
Closes Embucket/control-plane-v2#166

### DIFF
--- a/crates/control_plane/src/service.rs
+++ b/crates/control_plane/src/service.rs
@@ -22,6 +22,7 @@ use rusoto_core::{HttpClient, Region};
 use rusoto_credential::StaticProvider;
 use rusoto_s3::{GetBucketAclRequest, S3Client, S3};
 use snafu::ResultExt;
+use std::env;
 use std::sync::Arc;
 use url::Url;
 use uuid::Uuid;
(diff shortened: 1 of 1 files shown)


DanCodedThis (approved):

crates/control_plane/src/service.rs
```diff
@@ -227,8 +228,14 @@ impl ControlService for ControlServiceImpl {
             object_store,
         );
         let catalog = IcebergCatalog::new(Arc::new(rest_client), None).await?;
+        let sql_parser_dialect =
+            env::var("SQL_PARSER_DIALECT").unwrap_or_else(|_| "snowflake".to_string());
```
eadgbear, line 232: Not sure I agree with this. We're putting in so much effort to support Snowflake-only syntax, why bother supporting anything else?

osipovartem: let's switch to snowflake

rampage644: I would still have this supported even perhaps on the request-level, with a default for snowflake dialect. Previously, snowflake dialect was a bit stricter than generic.

osipovartem: So let's keep a flag from env with default to snowflake if not exist

rampage644 (approved):

PR #171, opened 2025-01-29
Add custom type planner

### LINKED ISSUE #131: Refactor datatypes in query planner
Remove all duplicate code from query planner by making certain methods in private datafusion fork public and keep only updated versions of `convert_data_type` method.

osipovartem:
Closes Embucket/control-plane-v2#131

- it's not possible to use session **with_type_planner** with custom SessionContext. Since there is no type_planner() method in SessionState (should be implemented in datafusion repo).
- Added custom type_planner to CustomContextProvider directly and removed default logic for other types


crates/runtime/src/datafusion/planner.rs
```diff
@@ -210,218 +165,6 @@ where
         Ok(Schema::new(fields))
     }
 
-    pub fn convert_data_type(&self, sql_type: &SQLDataType) -> Result<DataType> {
-        match sql_type {
-            SQLDataType::Array(
-                ArrayElemTypeDef::AngleBracket(inner_sql_type)
-                | ArrayElemTypeDef::SquareBracket(inner_sql_type, _),
```
DanCodedThis, line 254: A lot of methods were not moved, but removed, did we not use them?

DanCodedThis (dismissed):

crates/runtime/src/datafusion/planner.rs
osipovartem: We made them public in datafusion repo so we can remove duplicates

DanCodedThis (approved):

Vedin (approved):

crates/runtime/src/datafusion/context.rs
```diff
@@ -31,10 +31,12 @@ use datafusion::prelude::*;
 use datafusion::sql::planner::ContextProvider;
 use datafusion::sql::TableReference;
 use datafusion::variable::VarType;
+use datafusion_expr::planner::TypePlanner;
 
 pub struct CustomContextProvider<'a> {
```
rampage644: As mentioned previously: it does make sense to rethink if we need this `CustomContextProvider` at all

eadgbear (approved):

crates/runtime/src/datafusion/context.rs
osipovartem: Let's merge https://github.com/Embucket/datafusion/pull/9

PR #174, opened 2025-01-29
Icebucket(CatalogList/Catalog/CatalogProvider/MultiSchemaProvider)

Vedin:


crates/runtime/src/datafusion/data_catalog/extended_mirror.rs
```diff
@@ -0,0 +1,137 @@
...
+    error::Error as IcebergError,
+    object_store::Bucket,
+    spec::table_metadata::new_metadata_location,
+};
+use std::{collections::HashSet, sync::Arc};
+
+type NamespaceNode = HashSet<String>;
+
+#[derive(Debug)]
+enum Node {
```
rampage644: is it used anywhere?

crates/runtime/src/datafusion/data_catalog/extended_mirror.rs
```diff
@@ -0,0 +1,137 @@
...
+    Namespace(NamespaceNode),
+    Relation(Identifier),
+}
+
+#[derive(Debug)]
+pub struct ExtendedMirror {
```
rampage644: nitpick: Having a doc would help to understand what `mirror` means in this context

crates/runtime/src/datafusion/data_catalog/extended_mirror.rs
```diff
@@ -0,0 +1,137 @@
...
+        }
+    }
+
+    /// Lists all tables in the given namespace.
+    pub fn table_names(&self, namespace: &Namespace) -> Result<Vec<Identifier>, DataFusionError> {
+        let mut table_names = self.base.table_names(namespace)?;
+        let memory_tables: Vec<Identifier> = self
+            .memory_catalog_provider
+            .schema(namespace.to_string().as_str())
+            .unwrap()
```
rampage644: I believe clippy won't let use `unwrap()`, will it?

crates/runtime/src/datafusion/data_catalog/extended_mirror.rs
```diff
@@ -0,0 +1,137 @@
...
+    ) -> Result<Option<Arc<dyn TableProvider>>, DataFusionError> {
+        let iceberg_table = self.base.table(identifier.clone()).await?;
+        if iceberg_table.is_some() {
+            return Ok(iceberg_table);
+        }
+        let schema = self
+            .memory_catalog_provider
+            .schema(identifier.namespace().to_string().as_str());
+        schema
+            .expect("Schema does not exist")
```
rampage644: Same as with `unwrap()`: can we replace this with better error?

crates/runtime/src/datafusion/data_catalog/extended_mirror.rs
Vedin: No, it's redundant

crates/runtime/src/datafusion/data_catalog/extended_mirror.rs
Vedin: done

crates/runtime/src/datafusion/data_catalog/extended_mirror.rs
Vedin: done

crates/runtime/src/datafusion/data_catalog/extended_catalog.rs
```diff
@@ -0,0 +1,175 @@
...
+            .memory_catalog_provider
+            .schema(identifier.namespace().to_string().as_str())
+        {
+            schema.table(identifier.name()).await
+        } else {
+            Err(DataFusionError::Plan("Schema does not exist".to_string()))
+        }
+    }
+
+    #[allow(clippy::expect_used)]
```
Vedin, line 95: if somebody has any idea how can I get rid of this pls suggest. As I understood, I can't use `unwrap` or `match` or something similar here. Most likely I need use `map_or_else` somehow but I didn't fully get how use it in this case

Issue #157, opened 2025-01-22
api: remove dependency on warehouse_id in `query` endpoints


rampage644:
Presumably, this is an old artifact on previous code structure (catalog depends on `warehouse_id`) and a way to register a catalog for executor.

There should be no dependency on warehouse name in http handlers and service calls. Instead, we should parse FQN tables in the whole query, and register catalog based on that.
Alternatively, we can have a singleton SessionContext shareable between requests with all catalogs already registered.

YaroslavLitvinov:
I would divide this on two tasks:
- [ ] Basic functionality - get database, schema directly parsed from FQN;
- [ ] Extended functionality - infer database, schema from SessionContext if set before with USE DATABASE, USE SCHEMA” statements.

YaroslavLitvinov:
following `query` routes depend on warehouse_id :
```
/ui/warehouses/:warehouseId/databases/:databaseName/tables/:tableName/query

```
Following `query` functions depend on warehouse_id:
```
control-plane-v2/crates/nexus/src/http/ui/handlers/tables.rs :: query_table
control-plane-v2/crates/control_plane/src/service.rs :: ControlService::query_table
control-plane-v2/crates/control_plane/src/service.rs :: ControlService::query
control-plane-v2/crates/runtime/src/datafusion/execution.rs :: SqlExecutor::query
```

YaroslavLitvinov:
So far database & schema parsed from FQN tables looks good for simple sql statements, but for complicated ones which querying many tables in a cross database query in a single SQL statement - just some warehouse selected. By common sense - warehouse_id argument can be replaced only by one warehouse_id, despite how many candidates could be in sql statement. It's good to know which warehouse to choose. Probably it is not so important which warehouse will be chosen, since query planner should know how to run it, what/where to read and write.

YaroslavLitvinov:
After getting warehouse name instead of warehouse_id, we need to search warehouse by name in `ControlService::warehouse_repo`, though it is designed for key=`warehouse_id`. There are two implementations of `WarehouseRepository`, needed to be altered to search by warehouse name:
* WarehouseRepositoryDb - which uses SlateDb, where is searching by warehouse_name is quite problematic;
* InMemoryWarehouseRepository - which uses HashMap<uuid, Warehouse>.

rampage644:
`InMemoryWarehouseRepository` was originally added for testing purposes. 
Perhaps it's okay at this moment to do linear search over some in memory structure (cache?).

[...]

PR #176, opened 2025-01-29
remove warehouse_id argument in "query" function

YaroslavLitvinov:
This PR resolves issue #157.

Scalar functions supported, so `SELECT 1` query should work.
It provides warehouse_id parsed from query instead of removed function arg in `query` related functions, handlers.
It gets warehouse by doing linear search in both types of warehouses: persistent and in-memory.

### How it's tested
It seemed to me this test is broken `pytest tests/test_ui_endpoints.py`, so I used my simple tests to test /ui/query handler. Also in general functionality tested by simple slt test, which contains different types of queries to ensure parser correctly fetches table Idents:
``` bash
python -m icehut_sqllogictest   --config-file ~/.icehut/local-icehut.yaml --use-database tests2 --test-file tests/slt/sql/simple/test_fqn.slt 
```
tests/slt/sql/simple/test_fqn.slt:
```
statement ok
CREATE TABLE __DATABASE__.__SCHEMA__.simple_int(a INT);

statement ok
INSERT INTO __DATABASE__.__SCHEMA__.simple_int(a) VALUES (1),(2),(3),(4),(5);

statement ok
UPDATE __DATABASE__.__SCHEMA__.simple_int SET a=6 WHERE a=5;

statement ok
SELECT * FROM __DATABASE__.__SCHEMA__.simple_int;

statement ok
DROP TABLE __DATABASE__.__SCHEMA__.simple_int;
```


eadgbear:
Can we get some unit tests around this? Try to catch some edge cases.

crates/runtime/src/datafusion/execution.rs
```diff
@@ -723,6 +735,72 @@ impl SqlExecutor {
         }
     }
 
+    #[must_use]
+    #[allow(clippy::too_many_lines)]
+    pub fn get_table_path(&self, statement: &DFStatement) -> TablePath {
+        let table_path = |arr: &Vec<Ident>| -> TablePath {
```
osipovartem, line 758: Maybe better to use resolve_table_references and just iterate over it
```rust
let references = state
    .resolve_table_references(&statement)
```
[Link](https://github.com/Embucket/control-plane-v2/blob/main/crates/runtime/src/datafusion/execution.rs#L486-L503)

YaroslavLitvinov: Good catch, I missed that. It would be better reusing code vs adding new function, but accidentally  `.resolve_table_references(&statement)` returns no references at all for following query `DROP TABLE db.tests2.simple_int`. Also I tried to `preprocess_query()` before `resolve_table_references`, but without effect. Probably it's something known, but actually it's a bit strange for such an obvious bug in Datafusion. If no workaround for this maybe it's better to keep `get_table_path`. Anyway need more suggestions on this.

YaroslavLitvinov:
> Can we get some unit tests around this? Try to catch some edge cases.

I agree on this, I could add rust unit tests for parser. 
But running functional tests for "Snowflake endpoint" and for ui endpoint (which is not maintained if I don't miss something) would gave as even more cases tested, so we would check health of entire system as it always end up with successful statement execution.

YaroslavLitvinov:
What's new: Refactored ControlService::query function. Added unit test checking if scalar function is working.

crates/control_plane/src/error.rs
```diff
@@ -53,6 +53,9 @@ pub enum ControlPlaneError {
     #[snafu(display("Warehouse not found: {id}"))]
     WarehouseNotFound { id: Uuid },
 
+    #[snafu(display("Warehouse name not found: '{name}'"))]
+    WarehouseNameNotFound { name: String },
```
rampage644, line 57: nitpick: perhaps unify somehow with `WarehouseNotFound` error

crates/nexus/src/http/ui/handlers/tables.rs
```diff
@@ -265,14 +265,13 @@ pub async fn delete_table(
 #[tracing::instrument(level = "debug", skip(state), err, ret(level = tracing::Level::TRACE))]
 pub async fn query_table(
     State(state): State<AppState>,
-    Path((warehouse_id, database_name, table_name)): Path<(Uuid, String, String)>,
```
rampage644, line 268: Perhaps let's keep it for now to keep backward compatibility with UI? Will drop once UI has migrated

crates/nexus/src/http/ui/router.rs
```diff
@@ -56,10 +56,7 @@ pub fn create_router() -> Router<AppState> {
             "/warehouses/:warehouseId/databases/:databaseName/tables/:tableName",
             get(get_table).delete(delete_table),
         )
-        .route(
-            "/warehouses/:warehouseId/databases/:databaseName/tables/:tableName/query",
-            post(query_table),
-        )
+        .route("/query", post(query_table))
```
rampage644, line 63: Same as previous comment

rampage644 (dismissed):
Overall lgtm, please consider keeping current URL scheme to keep UI working

DanCodedThis (dismissed):

crates/nexus/src/http/ui/handlers/tables.rs
YaroslavLitvinov: For such compatibility it makes sense to keep only URL handler, since this function is internal

crates/control_plane/src/error.rs
YaroslavLitvinov: it's possible to do somehow if replace Uuid by String, but it's worse than this.

rampage644 (approved):

crates/nexus/src/http/ui/router.rs
YaroslavLitvinov: Added route `/warehouses/:warehouseId/databases/:databaseName/tables/:tableName/query` for compatibility with UI, though it's already requires FQN and ignores all the warehouseId, databaseName, tableName arguments in url.

PR #180, opened 2025-02-03
Implement copy into with stage

### LINKED ISSUE #178: engine: Implement COPY INTO
https://docs.snowflake.com/en/sql-reference/sql/copy-into-table

```sql
COPY INTO SNOWPLOW_ICEBERG.SNOWPLOW_WEB.SAMPLE_EVENTS_WEB_BASE FROM @~/snowplow_web_sample_stage
FILE_FORMAT = (TYPE=CSV field_delimiter=',' skip_header=1 FIELD_OPTIONALLY_ENCLOSED_BY='"');

```

osipovartem:
Closes Embucket/control-plane-v2#178
- Added support to create stages from csv files
- Implement copy into from stage


crates/control_plane/src/models/mod.rs
```diff
@@ -211,6 +212,18 @@ impl StorageProfile {
...
     }
 
+    pub fn get_object_store_endpoint_url(&self) -> ControlPlaneModelResult<Url> {
+        let storage_endpoint_url =
+            self.endpoint
+                .as_ref()
+                .ok_or_else(|| ControlPlaneModelError::InvalidEndpointUrl {
+                    url: self.endpoint.clone().unwrap_or_default(),
+                })?;
+        Url::parse(storage_endpoint_url).map_err(|_| ControlPlaneModelError::InvalidEndpointUrl {
```
rampage644: @slyons Am I correct this should be written with https://docs.rs/snafu/latest/snafu/#struct-style snafu `.context()` method to wrap an error?

crates/runtime/src/datafusion/execution.rs
```diff
@@ -154,10 +159,13 @@ impl SqlExecutor {
             .to_string();
         let query = date_add.replace_all(&query, "$1$2('$3',").to_string();
         // TODO implement alter session logic
-        query.replace(
-            "alter session set query_tag = 'snowplow_dbt'",
-            "SHOW session",
-        )
+        query
```
rampage644, line 174: Ok for now, but SQL text manipulation should be avoided going forward. I begin to think we might want to log warnings while we still do. I.e. 
 * receive query with `alter session` or others
 * override
 * log a warning

crates/runtime/src/datafusion/execution.rs
```diff
@@ -301,6 +308,142 @@ impl SqlExecutor {
         }
     }
 
+    pub async fn create_stage_query(
```
rampage644, line 330: nitpick: Since this is somewhat experimental let's please document this function to indicate this. 
```rust
// This is experimental CREATE STAGE support
// Current limitations 
// TODO
// - 1st
// - 2nd
// etc
pub async fn create_stage_query(
...
```

rampage644 (approved):
Good enough, minor comments

crates/control_plane/src/models/mod.rs
eadgbear: That's correct.

eadgbear (commented):
We should also get in the habit of adding tracing inside of functions. After the upload, after the data is processed, etc.

crates/control_plane/src/models/mod.rs
rampage644: @osipovartem Could we update it? I am seeing other code using `.context()` already and am curios to learn why it was different here :)

Issue #184, opened 2025-02-04
Add support for executing `CREATE SCHEMA` statements


rampage644:
Having merged #176 there is a way to automatically deduce warehouse and register corresponding Iceberg catalog provider for query execution. However, only FQN table statements are supported.

We need to update it such that schema - namespace queries are supported as well. Currently they fail with since there is no table in such queries and default "datafusion" catalog provider is used.

YaroslavLitvinov:
Close as done

PR #188, opened 2025-02-04
support create schema statement

YaroslavLitvinov:
Fix for issue #184


crates/control_plane/src/service.rs
```diff
@@ -226,8 +226,9 @@ impl ControlService for ControlServiceImpl {
 
         let table_path = executor.get_table_path(&statement);
         let warehouse_name = table_path.db;
+        // println!("warehouse_name: {warehouse_name}, query:");
```
osipovartem: let's remove print

osipovartem (approved):

PR #200, opened 2025-02-05
Skip iceberg related keywords

Vedin:
I decided to skip all iceberg-related keywords because we are not using them now. 
It unblocks the current "create iceberg table" behavior.

### DIFF
--- a/crates/runtime/src/datafusion/execution.rs
+++ b/crates/runtime/src/datafusion/execution.rs
@@ -1091,8 +1091,17 @@ impl SqlExecutor {
                         #[allow(clippy::unwrap_used)]
                         let mut query = create_table_statement.query.unwrap();
                         self.update_tables_in_query(&mut query, warehouse_name);
+                        // TODO: Removing all iceberg properties is temporary solution. It should be
+                        // implemented properly in future.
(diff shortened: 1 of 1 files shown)


crates/runtime/src/datafusion/execution.rs
```diff
@@ -1091,8 +1091,17 @@ impl SqlExecutor {
                         #[allow(clippy::unwrap_used)]
                         let mut query = create_table_statement.query.unwrap();
                         self.update_tables_in_query(&mut query, warehouse_name);
+                        // TODO: Removing all iceberg properties is temporary solution. It should be
+                        // implemented properly in future.
+                        // https://github.com/Embucket/control-plane-v2/issues/199
                         let modified_statement = CreateTableStatement {
                             query: Some(query),
+                            iceberg: false,
```
rampage644, line 1099: This should be fine given we currently treat CreateTableStatement as iceberg anyways, am I correct?

rampage644 (approved):

crates/runtime/src/datafusion/execution.rs
Vedin: Yep, I tested with some basic examples and it worked for me.

PR #205, opened 2025-02-06
Fix minor issues

osipovartem:
- added references check to use TableReference from datafusion
- added support for Merge statement warehouse check
- removed overlapping logic for table_path check 
- fixed convert_timezone return type
- moved common logic for show objects and schemas to a separate func

### DIFF
--- a/crates/runtime/src/datafusion/execution.rs
+++ b/crates/runtime/src/datafusion/execution.rs
@@ -42,13 +42,6 @@ use std::collections::HashMap;
 use std::sync::Arc;
 use url::Url;
 
-#[derive(Debug)]
-pub struct TablePath {
-    pub db: String,
-    pub schema: String,
-    pub table: String,
-}
-
 pub struct SqlExecutor {
     // ctx made public to register_catalog after creating SqlExecutor
     pub ctx: SessionContext,
@@ -477,16 +470,13 @@ impl SqlExecutor {
         warehouse_name: &str,
     ) -> IcehutSQLResult<Vec<RecordBatch>> {
         if let Statement::Merge {
-            mut table,
+            table,
             mut source,
             on,
             clauses,
             ..
         } = statement
         {
-            self.update_tables_in_table_factor(&mut table, warehouse_name);
-            self.update_tables_in_table_factor(&mut source, warehouse_name);
-
             let (target_table, target_alias) = Self::get_table_with_alias(table);
             let (_source_table, source_alias) = Self::get_table_with_alias(source.clone());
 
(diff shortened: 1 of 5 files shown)


crates/runtime/src/datafusion/execution.rs
```diff
@@ -42,13 +42,6 @@ use std::collections::HashMap;
 use std::sync::Arc;
 use url::Url;
 
-#[derive(Debug)]
```
osipovartem, line 45: Use datafusion TableReference instead of custom implementation

crates/runtime/src/datafusion/execution.rs
```diff
@@ -477,16 +470,13 @@ impl SqlExecutor {
...
         if let Statement::Merge {
-            mut table,
+            table,
             mut source,
             on,
             clauses,
             ..
         } = statement
         {
-            self.update_tables_in_table_factor(&mut table, warehouse_name);
```
osipovartem, line 487: We don't need this anymore since there is a check for warehouse

crates/runtime/src/datafusion/execution.rs
```diff
@@ -921,81 +911,36 @@ impl SqlExecutor {
...
-                        }
-                    }
-                    _ => table_path(&vec![]),
-                },
-                Statement::CreateTable(create_table) => table_path(&create_table.name.0),
-                Statement::Update { table, .. } => match table.relation {
-                    TableFactor::Table { name, .. } => table_path(&name.0),
-                    _ => table_path(&vec![]),
-                },
+                Statement::Drop { names, .. } => {
```
osipovartem, line 920: Keep only missing statements which are not supported by references visitor

crates/runtime/src/datafusion/functions/convert_timezone.rs
```diff
@@ -91,31 +90,26 @@ impl ScalarUDFImpl for ConvertTimezoneFunc {
...
+        internal_err!("return_type_from_args should be called")
     }
-    fn return_type_from_exprs(
-        &self,
-        args: &[Expr],
-        _schema: &dyn ExprSchema,
-        arg_types: &[DataType],
-    ) -> Result<DataType> {
-        match args.len() {
+    fn return_type_from_args(&self, args: ReturnTypeArgs) -> Result<ReturnInfo> {
```
osipovartem, line 95: return_type_from_exprs already deprecated and we should use return_type_from_args

crates/runtime/src/datafusion/planner.rs
```diff
@@ -239,102 +238,72 @@ where
     }
 
     fn show_objects_to_plan(&self, parent: &ObjectName) -> Result<LogicalPlan> {
+        if !self.inner.has_table("information_schema", "df_settings") {
```
osipovartem, line 241: This is a required check for Show* statements

crates/runtime/src/datafusion/planner.rs
```diff
@@ -239,102 +238,72 @@ where
...
-                    || plan_err!("Failed to parse SQL statement"),
-                    |statement| {
-                        if let DFStatement::Statement(s) = statement {
-                            self.sql_statement_to_plan(*s)
-                        } else {
-                            plan_err!("Failed to parse SQL statement")
-                        }
-                    },
-                )
+                let query = format!("SELECT {columns} FROM information_schema.tables");
```
osipovartem, line 261: removed catalog and schema check since on this step we don't have any records and dbt fails..

DanCodedThis (approved):

rampage644 (approved):

PR #222, opened 2025-02-10
Add partial support of arrow format for dbt/snowflake endpoint

### LINKED ISSUE #115: api: switch from JSON data type to ARROW IPC

YaroslavLitvinov:
Closes #115 

It has no breaking changes as "json" format is not replaced by "arrow" yet, as of known issues in arrow support.
This PR adds partial arrow support, that can be switched on using SERIALIZATION_FORMAT const.
In this PR UInt64 datatype replaced by Int64 in count batch of output schema.
Following queries return count batch that being fixed:
```
CREATE TABLE
CREATE SCHEMA
CREATE DATABASE
INSERT, UPDATE, DELETE
```

Datafusion repo updated as a part of fix: 
https://github.com/Embucket/datafusion/pull/17

Hash commit of change in datafusion propagated in repos too:
https://github.com/Embucket/datafusion-functions-json/pull/4
https://github.com/Embucket/iceberg-rust/pull/14

### DIFF
--- a/crates/nexus/src/http/dbt/handlers.rs
+++ b/crates/nexus/src/http/dbt/handlers.rs
@@ -25,7 +25,7 @@ use tracing::debug;
 use uuid::Uuid;
 
 // TODO: move out as a configurable parameter
-const SERIALIZATION_FORMAT: &str = "arrow"; // or "json"
+const SERIALIZATION_FORMAT: &str = "json"; // or "arrow"
 
 // https://arrow.apache.org/docs/format/Columnar.html#buffer-alignment-and-padding
 // Buffer Alignment and Padding: Implementations are recommended to allocate memory 
--- a/crates/nexus/src/http/dbt/handlers.rs
+++ b/crates/nexus/src/http/dbt/handlers.rs
@@ -24,7 +24,15 @@ use std::io::Read;
 use tracing::debug;
 use uuid::Uuid;
 
+// TODO: move out as a configurable parameter
 const SERIALIZATION_FORMAT: &str = "arrow"; // or "json"
+
+// https://arrow.apache.org/docs/format/Columnar.html#buffer-alignment-and-padding
+// Buffer Alignment and Padding: Implementations are recommended to allocate memory 
+// on aligned addresses (multiple of 8- or 64-bytes) and pad (overallocate) to a 
+// length that is a multiple of 8 or 64 bytes. When serializing Arrow data for interprocess 
+// communication, these alignment and padding requirements are enforced. 
+// For more info see issue #115
 const ARROW_IPC_ALIGNMENT: usize = 8;
 
 #[tracing::instrument(level = "debug", skip(state), err, ret(level = tracing::Level::TRACE))]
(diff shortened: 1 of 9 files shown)


DanCodedThis (changes requested):
Shouldn't we go from `UInt64` to `Int128`? Otherwise, could be potentially unsafe, since the maximum number of `UInt64` is not the maximum number of `Int64`

YaroslavLitvinov:
> Shouldn't we go from `UInt64` to `Int128`? Otherwise, could be potentially unsafe, since the maximum number of `UInt64` is not the maximum number of `Int64`

For result count it makes no sence.

DanCodedThis (approved):

crates/nexus/src/http/dbt/handlers.rs
```diff
@@ -5,16 +5,28 @@ use crate::http::dbt::schemas::{
...
+use base64::engine::general_purpose::STANDARD as engine_base64;
+use base64::prelude::*;
 use flate2::read::GzDecoder;
 use regex::Regex;
 use snafu::ResultExt;
 use std::io::Read;
+use tracing::debug;
 use uuid::Uuid;
 
+const SERIALIZATION_FORMAT: &str = "arrow"; // or "json"
```
rampage644: nitpick: sometime in the future this can be moved out to be a configuration parameter

crates/nexus/src/http/dbt/handlers.rs
```diff
@@ -5,16 +5,28 @@ use crate::http::dbt::schemas::{
...
+const ARROW_IPC_ALIGNMENT: usize = 8;
```
rampage644, line 36: Can we have a comment about why 8?

crates/nexus/src/http/dbt/handlers.rs
```diff
@@ -139,3 +196,120 @@ pub fn extract_token(headers: &HeaderMap) -> Option<String> {
     file.write_all(b"\n").await?;
     Ok(())
 }*/
+
+// mod tests {
```
rampage644: Just remove if not needed?

crates/runtime/Cargo.toml
```diff
@@ -12,15 +12,15 @@ async-trait = { workspace = true }
 
 chrono = { workspace = true }
 
-datafusion = { git = "https://github.com/Embucket/datafusion.git", rev = "a4c5f3f065b551afdf845000e3630012af27dc17" }
-datafusion-common = { git = "https://github.com/Embucket/datafusion.git", rev = "a4c5f3f065b551afdf845000e3630012af27dc17" }
-datafusion-expr = { git = "https://github.com/Embucket/datafusion.git", rev = "a4c5f3f065b551afdf845000e3630012af27dc17" }
-datafusion-functions-json = { git = "https://github.com/Embucket/datafusion-functions-json.git", rev = "9e9bd378b979a6b7081cf278c54a76eb94b00fc5" }
-datafusion_iceberg = { git = "https://github.com/Embucket/iceberg-rust.git", rev = "c5d3eb7a19c2777a8843a454a1a42cad3bec61d1" }
+datafusion = { git = "https://github.com/Embucket/datafusion.git", rev = "2b563bbfae1e6c33a8c5ced35cb6ccfb993b8b5c" }
```
rampage644: A bit worried this might bring irrelevant changes? Can we split into separate commit? Or changes there are related?

YaroslavLitvinov: providing those hashes in the same PR is mandatory (we do squash and rebase), they are a bit resorted as of multiple merge sessions.

crates/nexus/src/http/dbt/handlers.rs
YaroslavLitvinov: Probably we need it in future, as there are some things tested there,  but somehow cargo build didn't recognize imports related to mod tests and there were a lot of warnings while building main project.

crates/runtime/Cargo.toml
YaroslavLitvinov: Having those hashes shouldn't brake anything as it has all  the commits prior commit specified by hash

Issue #223, opened 2025-02-10
Improve `dateadd` UDF


rampage644:
During `dbt run` 9th model (`snowplow.PUBLIC_scratch.snowplow_web_sessions_this_run`) fails due to following statement:

```sql
...
    case when a.event_name = 'page_ping' then
        
    dateadd(second, -5, c.start_tstamp)::timestamp(6)

    else c.start_tstamp end as start_tstamp,
...
```

What is more interesting, there is no well formed message in the logs and service panics:
```
thread 'tokio-runtime-worker' panicked at /Users/ramp/.asdf/installs/rust/1.81.0/registry/src/index.crates.io-6f17d22bba15001f/arrow-array-54.1.0/src/array/primitive_array.rs:696:9:
Trying to access an element at index 0 from a PrimitiveArray of length 0
thread 'tokio-runtime-worker' panicked at /Users/ramp/.asdf/installs/rust/1.81.0/registry/src/index.crates.io-6f17d22bba15001f/arrow-array-54.1.0/src/array/primitive_array.rs:696:9:
Trying to access an element at index 0 from a PrimitiveArray of length 0
thread 'tokio-runtime-worker' panicked at /Users/ramp/.asdf/installs/rust/1.81.0/registry/src/index.crates.io-6f17d22bba15001f/arrow-array-54.1.0/src/array/primitive_array.rs:696:9:
Trying to access an element at index 0 from a PrimitiveArray of length 0
thread 'tokio-runtime-worker' panicked at /Users/ramp/.asdf/installs/rust/1.81.0/registry/src/index.crates.io-6f17d22bba15001f/arrow-array-54.1.0/src/array/primitive_array.rs:696:9:
Trying to access an element at index 0 from a PrimitiveArray of length 0
thread 'tokio-runtime-worker' panicked at /Users/ramp/.asdf/installs/rust/1.81.0/registry/src/index.crates.io-6f17d22bba15001f/arrow-array-54.1.0/src/array/primitive_array.rs:696:9:
Trying to access an element at index 0 from a PrimitiveArray of length 0
2025-02-10T19:49:26.255961Z ERROR request{method=POST uri=/queries/v1/query-request?requestId=b2a0321f-d2ba-40fe-925e-a32aaf67c4da&clientStartTime=1739216954476&retryCount=4&retryReason=500 version=HTTP/1.1}: tower_http::catch_panic: Service panicked: Trying to access an element at index 0 from a PrimitiveArray of length 0
[...]

PR #224, opened 2025-02-11
Fix dateadd function

rampage644:
This is a quick fix for `dateadd` UDF - it just makes it work such that its physical execution works for dbt snowplow 
workload.

 * Use interval instead of duration type for hour/minute/second
 * Handle empty array case

Ideally, it should be reimplemented differently:
 * Non-deprecated methods should be used
 * It should be designed to handle `ColumnarValue::Array` as a first class citizen (whereas scalar is an exception, not otherwise)
 * (Debatable) Perhaps even 2nd argument should be treated as a columnar value, not scalar

Good example is in [datafusion-examples directory](https://github.com/apache/datafusion/blob/main/datafusion-examples/examples/advanced_udf.rs).

Fixes - somewhat incorrect - #223

### DIFF
--- a/crates/runtime/src/datafusion/functions/date_add.rs
+++ b/crates/runtime/src/datafusion/functions/date_add.rs
@@ -226,6 +226,37 @@ mod tests {
             ),
         };
         match DateAddFunc::new().invoke_with_args(fn_args) {
+            Ok(ColumnarValue::Scalar(result)) => {
+                let expected = ScalarValue::TimestampMicrosecond(
+                    Some(1736600400000000i64),
+                    Some(Arc::from(String::from("+00").into_boxed_str())),
+                );
+                assert_eq!(&result, &expected, "date_add created a wrong value")
+            }
+            _ => panic!("Conversion failed"),
+        }
+    }
+    #[test]
+    fn test_date_add_days_timestamp_array() {
+        let args = vec![
+            ColumnarValue::Scalar(ScalarValue::Utf8(Some(String::from("days")))),
+            ColumnarValue::Scalar(ScalarValue::Int64(Some(5i64))),
+            ColumnarValue::Array(ScalarValue::TimestampMicrosecond(
+                Some(1736168400000000i64),
+                Some(Arc::from(String::from("+00").into_boxed_str())),
+            )
+            .to_array()
+            .unwrap()),
+        ];
+        let fn_args = ScalarFunctionArgs {
+            args: args,
+            number_rows: 0,
+            return_type: &arrow_schema::DataType::Timestamp(
+                arrow_schema::TimeUnit::Microsecond,
+                Some(Arc::from(String::from("+00").into_boxed_str())),
+            ),
+        };
+        match DateAddFunc::new().invoke_with_args(fn_args) {
             Ok(ColumnarValue::Array(result)) => {
                 let expected = ScalarValue::TimestampMicrosecond(
                     Some(1736600400000000i64),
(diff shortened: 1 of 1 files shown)


eadgbear (approved):

crates/runtime/src/datafusion/functions/date_add.rs
```diff
@@ -170,7 +171,12 @@ impl ScalarUDFImpl for DateAddFunc {
         };
         let date_or_time_expr = match &args[2] {
             ColumnarValue::Scalar(val) => val.clone(),
-            ColumnarValue::Array(array) => ScalarValue::try_from_array(&array, 0)?,
+            ColumnarValue::Array(array) => {
+                if array.len() == 0 {
+                    return Ok(ColumnarValue::Array(array.clone()));
+                }
+                ScalarValue::try_from_array(&array, 0)?
```
DanCodedThis: I think we can just copy the whole array and coerce everything to an array like in date diff probably

DanCodedThis: > ...
> * Non-deprecated methods should be used
> * It should be designed to handle `ColumnarValue::Array` as a first class citizen (whereas scalar is an exception, not otherwise)
> * (Debatable) Perhaps even 2nd argument should be treated as a columnar value, not scalar
>
> ...

I'll look into this

crates/runtime/src/datafusion/functions/date_add.rs
```diff
@@ -169,8 +170,8 @@ impl ScalarUDFImpl for DateAddFunc {
             _ => return plan_err!("Invalid value type"),
         };
         let date_or_time_expr = match &args[2] {
-            ColumnarValue::Scalar(val) => val.clone(),
-            ColumnarValue::Array(array) => ScalarValue::try_from_array(&array, 0)?,
+            ColumnarValue::Scalar(val) => val.to_array()?,
```
osipovartem: let date_or_time_expr = args[2].to_array(1)?;

DanCodedThis: > ```
>     let date_or_time_expr = args[2].to_array(1)?;
> ```

It maybe more than one, no?

Can't we for example give it a column of timestamps, that all need to add 5 days for example?

DanCodedThis:
- Arrays are first citizens now, and array length bug should be fixed now
- For now I implemented a simple fix for returning `Scalar` vs `Array`, from Artem's suggestion: https://github.com/Embucket/datafusion/blob/icehut/datafusion/functions/src/regex/regexpsubstr.rs#L111
- DF says for perf reasons it's better to have different impl's for them, so later we can explore that to speed it up (if it's slow)

crates/runtime/src/datafusion/functions/date_add.rs
```diff
@@ -208,11 +186,63 @@ impl ScalarUDFImpl for DateAddFunc {
...
+#[cfg(test)]
+#[allow(clippy::unwrap_in_result)]
+mod tests {
+    use super::DateAddFunc;
+    use datafusion_common::ScalarValue;
+    use datafusion_expr::{ColumnarValue, ScalarFunctionArgs, ScalarUDFImpl};
+    use std::sync::Arc;
+
+    #[test]
+    fn test_date_add_days_timestamp() {
```
rampage644, line 211: Could we please add a bit more tests here? Am I missing a test for ColumnarValue::Array (not seeing in this diff, but it is present)?

DanCodedThis: Added tests for arrays

Vedin (approved):
It's fine for now and I believe work pretty fast. I still don't like to create additional array for scalar values. We should consider using something similar to `evaluate_array_scalar` in the future.

rampage644:
@DanCodedThis Feel free to take over this (i.e. decision to merge is on you), feel free to remove all code I've written altogether.

Issue #175, opened 2025-01-29
Set current database, schema for the session


YaroslavLitvinov:
Unless statements are executed in the session are not fully qualified it fail if no current database, schema ever set.

Support `USE DATABASE`, `USE SCHEMA` statements.  
Support setting database, schema specified in connection settings, for instance in `snow sql`.

PR #232, opened 2025-02-12
Set variable, set entity within session

### LINKED ISSUE #173: Support setting variables with SET statement
Add support for setting variables using SET, like `SET timezone='UTC'` query.

More context:
Since issue https://github.com/Embucket/control-plane-v2/issues/157 been resolved, and `warehouse_id` argument is not anymore provided in `query` functions and handlers, it failing with error "Warehouse name not found". As no FQN present in query and warehouse name can't be interred.

osipovartem:
Closes #173 and #175

### DIFF
--- a/crates/runtime/src/datafusion/execution.rs
+++ b/crates/runtime/src/datafusion/execution.rs
@@ -35,7 +35,7 @@ use snafu::ResultExt;
 use sqlparser::ast::helpers::attached_token::AttachedToken;
 use sqlparser::ast::{
     BinaryOperator, GroupByExpr, MergeAction, MergeClauseKind, MergeInsertKind, ObjectType,
-    Query as AstQuery, Select, SelectItem,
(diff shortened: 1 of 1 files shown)


DanCodedThis (approved):

rampage644 (approved):

crates/runtime/src/datafusion/execution.rs
```diff
@@ -89,8 +89,56 @@ impl SqlExecutor {
         // etc
         if let DFStatement::Statement(s) = statement {
             match *s {
-                Statement::AlterSession { .. } => {
-                    return Box::pin(self.alter_session_query(*s)).await;
+                Statement::AlterSession {
+                    set,
+                    session_params,
+                } => {
+                    let params = session_params
```
eadgbear, line 96: The `session_params` are owned why are we cloning them? `into_iter` should give us an owned HashMap

PR #235, opened 2025-02-12
Fixing CORS to allow Cookies from UI

### LINKED ISSUE #236: Enable setting CORS `Allow-Origin` to enable cookies from UI
* With the new cookie-based client sessions, CORS needs to be modified from allow-all to a specified domain/value

eadgbear:
Closes #236


osipovartem (approved):

crates/nexus/src/http/layers.rs
```diff
@@ -41,3 +47,20 @@ pub async fn add_request_metadata(
...
+    Ok(CorsLayer::new()
+        .allow_origin(origin_value)
+        .allow_methods(vec![
+            Method::GET,
+            Method::POST,
+            Method::DELETE,
+            Method::HEAD,
+        ])
+        .allow_headers(vec![AUTHORIZATION, CONTENT_TYPE])
+        .allow_credentials(true))
```
osipovartem, line 65: Invalid CORS configuration: Cannot combine `Access-Control-Allow-Credentials: true` with `Access-Control-Allow-Origin: *`
with default env config

eadgbear:
I'll have to add something that disables CORS

crates/nexus/src/http/layers.rs
eadgbear: Fixed. There's now a flag to enable CORS in the first place. We'll have to update how we're running in prod

osipovartem (approved):

PR #248, opened 2025-02-14
Changing contributor script

eadgbear:
Changing to one that will issue a PR instead


eadgbear:
Bypassing review process just for this PR

PR #250, opened 2025-02-14
Make SP endpoint optional

osipovartem:


Vedin (approved):
lgtm, I'm surprised that URL with only schema and path is a valid URL.

DanCodedThis (approved):
LGTM, closes https://github.com/Embucket/control-plane-v2/issues/203

osipovartem:
> lgtm, I'm surprised that URL with only schema and path is a valid URL.

By this url datafusion decides which object store to use by comparing file path

PR #272, opened 2025-02-18
Control plane tests (batch 2)

osipovartem:
- fixed default timestamps precision
- added CP repository and utils tests
Related to #264

### DIFF
--- a/crates/control_plane/src/utils.rs
+++ b/crates/control_plane/src/utils.rs
@@ -5,13 +5,11 @@ use arrow::array::{
 };
 use arrow::datatypes::{Field, Schema, TimeUnit};
 use arrow::record_batch::RecordBatch;
-use chrono::{DateTime, Utc};
+use chrono::DateTime;
 use datafusion::arrow::array::ArrayRef;
 use datafusion::arrow::datatypes::DataType;
 use datafusion::common::Result as DataFusionResult;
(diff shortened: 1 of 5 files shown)


crates/control_plane/src/utils.rs
```diff
@@ -108,11 +107,10 @@ fn convert_timestamp_to_struct(column: &ArrayRef, unit: TimeUnit) -> ArrayRef {
             .unwrap()
             .iter()
             .map(|x| {
-                #[allow(clippy::unwrap_used)]
-                let ts =
-                    DateTime::from_timestamp_millis(x.unwrap_or_else(|| now.timestamp_millis()))
-                        .unwrap();
-                format!("{}.{}", ts.timestamp(), ts.timestamp_subsec_micros())
+                x.map(|ts| {
+                    let ts = DateTime::from_timestamp_millis(ts).unwrap();
```
Vedin, line 111: is it intended that this logic is changed? As I can see it set now previously in case of invalid unwrap.

Vedin (approved):

DanCodedThis (approved):

crates/control_plane/src/utils.rs
osipovartem: If the field is empty we shouldn't set default value there. It should be null.

PR #274, opened 2025-02-18
Control plane tests (batch 3)

osipovartem:
- tests for upload_data and query endpoint require Catalog and will be added separately
Related to #264

### DIFF
--- a/crates/control_plane/src/utils.rs
+++ b/crates/control_plane/src/utils.rs
@@ -1,4 +1,5 @@
-use crate::models::ColumnInfo;
+use crate::error::{ControlPlaneError, ControlPlaneResult};
+use crate::models::{ColumnInfo, Credentials, StorageProfile};
 use arrow::array::{
     Array, StringArray, TimestampMicrosecondArray, TimestampMillisecondArray,
     TimestampNanosecondArray, TimestampSecondArray, UnionArray,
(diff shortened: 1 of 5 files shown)


eadgbear:
Instead of mocking the S3 client, we can explore using Bollard to start up a Minio docker container.

osipovartem:
> Instead of mocking the S3 client, we can explore using Bollard to start up a Minio docker container.

This should be added as a part of integration tests, let's add unit tests first. We can add a separate workflow or mark some tests with specific tag to run them as integration. WDYT?

eadgbear:
> > Instead of mocking the S3 client, we can explore using Bollard to start up a Minio docker container.
> 
> This should be added as a part of integration tests, let's add unit tests first. We can add a separate workflow or mark some tests with specific tag to run them as integration. WDYT?

Fair. I just don't think that mocking an S3 client is necessary outside of Integration tests. Unit test should ideally be Memory only, then integration tests for File/S3-based tests, IMO

crates/control_plane/src/utils.rs
```diff
@@ -9,8 +10,70 @@ use chrono::DateTime;
...
+#[async_trait::async_trait]
+impl S3ClientValidation for S3Client {
+    async fn get_aws_bucket_acl(
+        &self,
+        request: GetBucketAclRequest,
+    ) -> ControlPlaneResult<GetBucketAclOutput> {
+        self.client
+            .get_bucket_acl(request)
+            .await
+            .map_err(|_| ControlPlaneError::InvalidCredentials)
```
rampage644: snafu `.context`?

osipovartem:
> > > Instead of mocking the S3 client, we can explore using Bollard to start up a Minio docker container.
> > 
> > 
> > This should be added as a part of integration tests, let's add unit tests first. We can add a separate workflow or mark some tests with specific tag to run them as integration. WDYT?
> 
> Fair. I just don't think that mocking an S3 client is necessary outside of Integration tests. Unit test should ideally be Memory only, then integration tests for File/S3-based tests, IMO

That's why I added a mock to not use any real s3 client calls. This mock works in memory only and it allows us to cover create_storage_profile and validate_credentials endpoints by unit tests without any external calls

DanCodedThis (approved):

rampage644 (approved):

Issue #276, opened 2025-02-18
Snow sql convertor displaying issues


YaroslavLitvinov:
```
select 
    a.*,
    row_number() over (partition by a.event_id order by a.collector_tstamp, a.dvce_created_tstamp) as row_num from snowplow.public.events_iceberg a limit 1;
```
Above query succesfully returning results in json format when executed via /ui/query API.
But the same query executed from snow sql gets following results:
```
No column converter found for type: FLOAT32
No column converter found for type: FLOAT32
No column converter found for type: FLOAT32
+-------------------------------------------------------------------------------
||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||
|+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||||
+-------------------------------------------------------------------------------
```

Instead when snow sql --format=json specified it returns mostly expected result with only difference how it represents timestamp.
In /ui/query:
`"etl_tstamp": "2022-08-20T00:39:32.000913"`

In snow sql --format=json:
`"etl_tstamp": "1660955972.913000"`

osipovartem:
It should be fixed in conversion logic, assign to me, I will fix

YaroslavLitvinov:
/ui/query result:
```
[{
    "app_id": "website",
    "platform": "web",
    "etl_tstamp": "1660955972.913000",
    "collector_tstamp": "1660955972.261000",
    "dvce_created_tstamp": "1660955972.178000",
    "event": "struct",
    "event_id": "000c0134-c96a-4fe1-bc10-3d0898003f54",
    "txn_id": null,
    "name_tracker": "biz1",
    "v_tracker": "js-3.5.0",
    "v_collector": "ssc-2.6.1-kinesis",
    "v_etl": "snowplow-stream-enrich-3.2.3-common-3.2.3",
    "user_id": null,
    "user_ipaddress": "8dbe5e9e1ea6a6061e54e5a1659a13fef17f187cbcf2d1b54e5a7177ccd39f14",
    "user_fingerprint": null,
    "domain_userid": "e94c4a06-6ba1-47fc-9967-a882431ba615",
    "domain_sessionidx": 1,
    "network_userid": "7b0388e4-6e1d-43f2-ba56-88ed849aa83f",
    "geo_country": "US",
    "geo_region": "MO",
[...]

PR #279, opened 2025-02-19
Fix timestamp displaying format

osipovartem:
Related #276

### DIFF
--- a/crates/control_plane/src/models/mod.rs
+++ b/crates/control_plane/src/models/mod.rs
@@ -454,7 +454,8 @@ impl ColumnInfo {
             | DataType::UInt8
             | DataType::UInt16
             | DataType::UInt32
-            | DataType::UInt64 => {
+            | DataType::UInt64
+            | DataType::Float32 => {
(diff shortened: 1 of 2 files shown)


crates/control_plane/src/models/mod.rs
```diff
@@ -454,7 +454,8 @@ impl ColumnInfo {
             | DataType::UInt8
             | DataType::UInt16
             | DataType::UInt32
-            | DataType::UInt64 => {
+            | DataType::UInt64
+            | DataType::Float32 => {
```
Vedin, line 458: is this change related to fixing timestamps or it's just a minor fix you want to merge here as well?

Vedin (approved):

crates/control_plane/src/models/mod.rs
osipovartem: Details are here 
https://github.com/Embucket/icebucket/issues/276

YaroslavLitvinov (approved):

Issue #67, opened 2024-12-26
Snowflake python connector cannot parse null timestamps


osipovartem:
Initial create query (empty table with defined column types)
```sql
CREATE OR REPLACE TABLE example AS (WITH prep AS (SELECT CAST(NULL AS TEXT) AS model, CAST('1970-01-01' AS TIMESTAMP) AS last_success) SELECT * FROM prep WHERE false)
```
Select query
```sql
 SELECT min(last_success) AS min_last_success, max(last_success) AS max_last_success, coalesce(count(*), 0) AS models FROM example WHERE model IN ('snowplow_web_page_views', 'snowplow_web_pv_scroll_depth', 'snowplow_web_pv_engaged_time', 'snowplow_web_page_views_this_run', 'snowplow_web_user_mapping', 'snowplow_web_sessions', 'snowplow_web_sessions_this_run', 'snowplow_web_users', 'snowplow_web_users_sessions_this_run', 'snowplow_web_users_this_run', 'snowplow_web_users_aggs', 'snowplow_web_users_lasts') 
```
The table is empty before the select query. Response contains rowset and rowtype.
```json
{"data": {"rowtype": [{"name": "min_last_success", "database": "", "schema": "", "table": "", "nullable": true, "type": "timestamp_ntz", "byteLength": null, "length": null, "scale": 9, "precision": 0, "collation": null}, {"name": "max_last_success", "database": "", "schema": "", "table": "", "nullable": true, "type": "timestamp_ntz", "byteLength": null, "length": null, "scale": 9, "precision": 0, "collation": null}, {"name": "models", "database": "", "schema": "", "table": "", "nullable": false, "type": "fixed", "byteLength": null, "length": null, "scale": 0, "precision": 38, "collation": null}], "rowsetBase64": null, "rowset": [{"min_last_success": null, "max_last_success": null, "models": 0}], "total": 1, "queryResultFormat": "json", "sqlState": "ok"}, "success": true, "message": "successfully executed", "code": "000200"}
```
Snowflake dbt response for the json above
```
13:40:43    252005: Failed to convert: field min_last_success: TIMESTAMP_NTZ::min_last_success, Error: invalid literal for int() with base 10: 'min_last_success'
```
[...]

PR #280, opened 2025-02-19
Fix IPC arrow format issue

### LINKED ISSUE #115: api: switch from JSON data type to ARROW IPC

osipovartem:
- the issue was in NULL type that we send for **NULL as comment**
- NULL type was fixed
- Arrow IPC format [expects](https://github.com/snowflakedb/snowflake-connector-python/blob/main/src/snowflake/connector/nanoarrow_cpp/ArrowIterator/CArrowChunkIterator.cpp#L268-L293) all timestamps in i64 format with scale, for example 1627846261233222 scale 6
- JSON format [expects](https://github.com/snowflakedb/snowflake-connector-python/blob/main/src/snowflake/connector/converter.py#L769-L781) all timestamps in string format with point delimiter with scale, for example "1627846261.233222" scale 6
- Moved DBT_FORMAT to env variable with json default

Closes #115 and #67

### DIFF
--- a/crates/control_plane/src/utils.rs
+++ b/crates/control_plane/src/utils.rs
@@ -1,6 +1,6 @@
 use crate::models::ColumnInfo;
 use arrow::array::{
-    Array, StringArray, TimestampMicrosecondArray, TimestampMillisecondArray,
+    Array, Int64Array, StringArray, TimestampMicrosecondArray, TimestampMillisecondArray,
     TimestampNanosecondArray, TimestampSecondArray, UnionArray,
 };
 use arrow::datatypes::{Field, Schema, TimeUnit};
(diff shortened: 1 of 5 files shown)


crates/control_plane/src/utils.rs
```diff
@@ -9,7 +9,45 @@ use chrono::DateTime;
...
+        match self {
+            Self::Arrow => write!(f, "arrow"),
+            Self::Json => write!(f, "json"),
+        }
+    }
+}
+
+impl SerializationFormat {
+    fn new() -> Self {
+        let var = env::var("DBT_SERIALIZATION_FORMAT").unwrap_or_else(|_| "json".to_string());
```
Vedin, line 44: do we need to use `to_lowercase()` here, or do we always expect it to be in lowercase already?

DanCodedThis (approved):
LGTM, you checked the https://github.com/Embucket/icebucket/issues/67?

crates/control_plane/src/utils.rs
```diff
@@ -87,58 +127,95 @@ pub fn convert_record_batches(
...
+fn convert_timestamp_to_struct(
+    column: &ArrayRef,
+    unit: TimeUnit,
+    ser: SerializationFormat,
+) -> ArrayRef {
+    match ser {
+        SerializationFormat::Arrow => {
+            let timestamps: Vec<_> = match unit {
+                TimeUnit::Second => column
+                    .as_any()
```
Vedin: I'm not sure what's a best practice here but personally don't like repeating:
```
.as_any()
.downcast_ref::<TimestampMillisecondArray>()
.unwrap()
.iter()
.collect(),
```
Maybe introduce some macro or generic type to avoid this? I'm talking about smth like this:
```rust
macro_rules! downcast_timestamp {
    ($column:expr, $array_type:ty) => {
        $column
            .as_any()
            .downcast_ref::<$array_type>()
            .unwrap()
            .iter()
    };
}
```

eadgbear: We could implement `TryInto` or something similar to skip the steps, but it would just be moving the code elsewhere

PR #285, opened 2025-02-19
Update query runner with session id

Vedin:
I found this script useful for quick debugging and the quickest way to try something like "select 1".
This PR adds support for session-id and changes the location param to env variable instead of argument. 
Also, minor changes to show the full response for errors.


rampage644:
I typically use `snow sql -c icebucket -q 'select count() from snowplow.public.events;'`

rampage644 (approved):

PR #286, opened 2025-02-19
Move DBT_SERIALIZATION_FORMAT to cli

### LINKED ISSUE #284: Refactor passing configuration to control plane service
PR #280 introduced a way to pass certain configuration parameters to `ControlPlaneImpl` object. This approach is difficult to generalize, test and maintain.

Instead, it should follow Rust idiomatic approaches with either Builder pattern or `new_with_opts` (or something like this) constructor.

All configuration parameters should be defined declaratively and readable from either cli args or env (or in some future from configuration file). All components should be parametrizable via constructors only.

osipovartem:
Closes #284

### DIFF
--- a/bin/bucketd/src/cli.rs
+++ b/bin/bucketd/src/cli.rs
@@ -130,10 +130,10 @@ pub struct IceBucketOpts {
         short,
         long,
         default_value = "json",
-        env = "DBT_SERIALIZATION_FORMAT",
-        help = "Serialization format for dbt endpoints"
+        env = "DATA_FORMAT",
+        help = "Data serialization format in Snowflake v1 API"
     )]
-    pub dbt_serialization_format: Option<String>,
+    pub data_format: Option<String>,
 }
 
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, ValueEnum)]
--- a/bin/bucketd/src/cli.rs
+++ b/bin/bucketd/src/cli.rs
@@ -125,6 +125,15 @@ pub struct IceBucketOpts {
         help = "CORS Allow Origin"
     )]
     pub cors_allow_origin: Option<String>,
+
+    #[arg(
+        short,
+        long,
+        default_value = "json",
+        env = "DBT_SERIALIZATION_FORMAT",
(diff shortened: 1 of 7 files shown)


bin/bucketd/src/cli.rs
```diff
@@ -125,6 +125,15 @@ pub struct IceBucketOpts {
         help = "CORS Allow Origin"
     )]
     pub cors_allow_origin: Option<String>,
+
+    #[arg(
+        short,
+        long,
+        default_value = "json",
+        env = "DBT_SERIALIZATION_FORMAT",
```
rampage644: Let's begin to remove `dbt` wording from everywhere (because we're trying to mimic  snowflake v1 api and not dbt), I suggest either something generic or very specific. I.e. `output-format`, `response-format`, `encoding`, `data-format` etc.

bin/bucketd/src/cli.rs
```diff
@@ -125,6 +125,15 @@ pub struct IceBucketOpts {
...
+        help = "Serialization format for dbt endpoints"
```
rampage644: "Data serialization format in Snowflake v1 API"

bin/bucketd/src/main.rs
```diff
@@ -53,8 +57,15 @@ async fn main() {
         Ok(object_store) => {
             tracing::info!("Starting 🧊🪣 IceBucket...");
 
-            if let Err(e) =
-                nexus::run_icebucket(object_store, slatedb_prefix, host, port, allow_origin).await
+            if let Err(e) = nexus::run_icebucket(
```
rampage644, line 60: Something to think about - once we have more parameters, how to pass them without ever increasing list of arguments for this function...

rampage644 (approved):
lgtm with minor updates

bin/bucketd/src/main.rs
eadgbear: We could theoretically pass the `Cli` object around, but that would require that `nexus` take a dependency on `bin/bucketd`, and that's a really bad idea. Instead it should be a config type that'll eventually live in `runtime` since that's the crate that will unite the rest.

PR #287, opened 2025-02-20
Issue 115 Arrow support: replace uints by ints in resulted RecordBatch

### LINKED ISSUE #115: api: switch from JSON data type to ARROW IPC

YaroslavLitvinov:
Closes #115

No impact on result when json format enabled.
When output arrow format is enabled it map all the unsigned int types to signed int types in result batch:
UInt64 -> Decimal128(38, 10)
UInt32 -> Int64
UInt16 -> Int32
UInt8 -> Int16 

Tested with dbt, and b using following query: 
https://github.com/Embucket/icebucket/issues/115#issuecomment-2669652559

### DIFF
--- a/crates/control_plane/src/utils.rs
+++ b/crates/control_plane/src/utils.rs
@@ -32,8 +32,8 @@ use rusoto_core::{HttpClient, Region};
 use rusoto_credential::StaticProvider;
 use rusoto_s3::{GetBucketAclOutput, GetBucketAclRequest, S3Client as ExternalS3Client, S3};
 use snafu::ResultExt;
-use std::fmt;
 use std::collections::HashMap;
+use std::fmt;
 use std::sync::Arc;
 
 pub struct Config {
@@ -186,12 +186,17 @@ pub fn convert_record_batches(
                     Arc::clone(&converted_column)
                 }
                 DataType::UInt64 | DataType::UInt32 | DataType::UInt16 | DataType::UInt8 => {
+                    let column_info = &column_infos[i];
                     convert_uint_to_int_datatypes(
                         &mut fields,
                         &field,
                         column,
                         metadata,
-                        serialization_format,
+                        data_format,
+                        (
+                            column_info.precision.unwrap_or(38),
+                            column_info.scale.unwrap_or(0),
+                        ),
                     )
                 }
                 _ => {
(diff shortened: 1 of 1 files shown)


osipovartem (approved):

crates/control_plane/src/utils.rs
```diff
@@ -276,11 +287,90 @@ fn convert_timestamp_to_struct(
...
+    ser: SerializationFormat,
+) -> ArrayRef {
+    match ser {
+        SerializationFormat::Arrow => {
+            match field.data_type() {
+                DataType::UInt64 => {
+                    fields.push(
+                        Field::new(
+                            field.name(),
+                            DataType::Decimal128(38, 10),
```
Vedin: Any reason to set the scale to 10 here? We have only whole positive numbers. I'd say we can use Decimal128(38,0) here.

Vedin (changes requested):

DanCodedThis (approved):
LGTM, only concern is what Denys mentioned

crates/control_plane/src/utils.rs
osipovartem: Also it should match with https://github.com/Embucket/icebucket/blob/main/crates/control_plane/src/models/mod.rs#L486-L492
So Decimal128(38,0) should be correct

YaroslavLitvinov: I thought it is a number base or so. I'll fix it.

YaroslavLitvinov:
I tried to replace `DataType::Decimal128(38, 10)`  by `DataType::Decimal128(38, 0)` that is used with Decimal128Array and I failed. I'm unable to create array of type: `PrimitiveArray<Decimal128(38, 0)>`.
Surprisingly Decimal128Array hardly ties on Decimal128(38, 10) datatype, and thus it's creating arrays of kind  `PrimitiveArray<Decimal128(38, 10)>`

If my thoughts are correct, in order to work with arrays like PrimitiveArray<Decimal128(38, 0)> it's not enough just putting scale=0, to get rid of fractional part. As it gets to compile errors that is not a data type but variable/constant. And it require creating own datatype with implementing/copy-pasting all the traits for `pub struct Decimal128Type {}` as it done for original Decimal128Array. And efforts for support our own datatype.
Another thought I have is about how arrow ipc will conduct witch such array type, will it encode/decode it - I don't know.
I wouldn't do such thing and would used how it is implemented now with `DataType::Decimal128(38, 10) `
Any thoughts on it? As I stuck with it.

YaroslavLitvinov:
@Vedin @osipovartem ^^

crates/control_plane/src/utils.rs
YaroslavLitvinov: > Also it should match with https://github.com/Embucket/icebucket/blob/main/crates/control_plane/src/models/mod.rs#L486-L492 So Decimal128(38,0) should be correct

I agree, for consistency column_info.precision  / column_info.scale and respective fields from metadata should be equal. And column info is source of truth here.

YaroslavLitvinov:
> I tried to replace `DataType::Decimal128(38, 10)` by `DataType::Decimal128(38, 0)` that is used with Decimal128Array and I failed. I'm unable to create array of type: `PrimitiveArray<Decimal128(38, 0)>`. Surprisingly Decimal128Array hardly ties on Decimal128(38, 10) datatype, and thus it's creating arrays of kind `PrimitiveArray<Decimal128(38, 10)>`
> 
> If my thoughts are correct, in order to work with arrays like PrimitiveArray<Decimal128(38, 0)> it's not enough just putting scale=0, to get rid of fractional part. As it gets to compile errors that is not a data type but variable/constant. And it require creating own datatype with implementing/copy-pasting all the traits for `pub struct Decimal128Type {}` as it done for original Decimal128Array. And efforts for support our own datatype. Another thought I have is about how arrow ipc will conduct witch such array type, will it encode/decode it - I don't know. I wouldn't do such thing and would used how it is implemented now with `DataType::Decimal128(38, 10) ` Any thoughts on it? As I stuck with it.

`with_precision_and_scale` helped as Vedin suggested.

YaroslavLitvinov:
[test / ubuntu / stable (pull_request)](https://github.com/Embucket/icebucket/actions/runs/13442899067/job/37561368371?pr=287)
Failing after 11m Test Error: Process completed with exit code 143.
On dev environement it stuck on:
 test service::tests::test_delete_session has been running for over 60 seconds

Vedin (approved):

PR #289, opened 2025-02-21
map float types to real logical_type

### LINKED ISSUE #115: api: switch from JSON data type to ARROW IPC

YaroslavLitvinov:
Closes #115

### DIFF
--- a/crates/control_plane/src/models/mod.rs
+++ b/crates/control_plane/src/models/mod.rs
@@ -796,7 +796,6 @@ mod tests {
                 assert_ne!(column_info.scale.unwrap(), scale);
             }
         }
-
     }
 
     #[tokio::test]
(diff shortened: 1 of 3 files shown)


crates/control_plane/src/models/mod.rs
```diff
@@ -485,12 +485,16 @@ impl ColumnInfo {
...
             | DataType::UInt16
             | DataType::UInt32
-            | DataType::UInt64
-            | DataType::Float32 => {
+            | DataType::UInt64 => {
                 column_info.r#type = "fixed".to_string();
                 column_info.precision = Some(38);
                 column_info.scale = Some(0);
             }
+            DataType::Float16 | DataType::Float32 | DataType::Float64 => {
```
osipovartem, line 493: Please add test for this case

YaroslavLitvinov: added

DanCodedThis (approved):

Issue #293, opened 2025-02-21
Add Construction date & time functions


osipovartem:
Function |Status
 -- | :---
[DATE_FROM_PARTS](https://docs.snowflake.com/en/sql-reference/functions/date_from_parts) |  <ul><li>- [ ] https://github.com/Embucket/icebucket/pull/299</li></ul>
[TIME_FROM_PARTS](https://docs.snowflake.com/en/sql-reference/functions/time_from_parts) |  <ul><li>- [ ] https://github.com/Embucket/icebucket/pull/298</li></ul>
[TIMESTAMP_FROM_PARTS](https://docs.snowflake.com/en/sql-reference/functions/timestamp_from_parts) |  <ul><li>- [ ] https://github.com/Embucket/icebucket/pull/297</li></ul>

https://github.com/apache/datafusion/blob/main/datafusion/functions/src/datetime/make_date.rs
can be used as a template for the list above

PR #297, opened 2025-02-24
Snowflake timestamp_from_parts udf

osipovartem:
https://docs.snowflake.com/en/sql-reference/functions/timestamp_from_parts
- added new UDF timestamp_from_parts

Related #293

### DIFF
--- a/crates/runtime/src/datafusion/functions/timestamp_from_parts.rs
+++ b/crates/runtime/src/datafusion/functions/timestamp_from_parts.rs
@@ -82,10 +82,6 @@ const UNIX_EPOCH_DAYS: i32 = 719_163;
         description = "An integer expression to use as a second for building a timestamp, usually in the 0-59 range."
     ),
     argument(
-        name = "second",
-        description = "An integer expression to use as a second for building a timestamp, usually in the 0-59 range."
-    ),
-    argument(
         name = "date_expr",
         description = "Specifies the date expression to use for building a timestamp
          where date_expr provides the year, month, and day for the timestamp."
--- a/crates/runtime/src/datafusion/functions/timestamp_from_parts.rs
+++ b/crates/runtime/src/datafusion/functions/timestamp_from_parts.rs
@@ -27,6 +27,7 @@ use arrow::datatypes::{
 use arrow_schema::DataType::{Date32, Int32, Int64, Time64};
 use arrow_schema::TimeUnit;
 use chrono::prelude::*;
+use chrono::{Duration, Months};
 use datafusion::logical_expr::TypeSignature::Coercible;
 use datafusion::logical_expr::TypeSignatureClass;
 use datafusion_common::types::{logical_int64, logical_string};
@@ -36,7 +37,7 @@ use datafusion_expr::{
 };
 use datafusion_macros::user_doc;
 
-const UNIX_EPOCH_DAYS: i32 = 719_163;
+const UNIX_DAYS_FROM_CE: i32 = 719_163;
 
 #[user_doc(
     doc_section(label = "Time and Date Functions"),
@@ -56,7 +57,6 @@ const UNIX_EPOCH_DAYS: i32 = 719_163;
             | 1740398450.65555555                                                                               |
             +---------------------------------------------------------------------------------------------------+
 ```",
-    standard_argument(name = "str", prefix = "String"),
     argument(
         name = "year",
         description = "An integer expression to use as a year for building a timestamp."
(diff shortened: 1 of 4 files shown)


DanCodedThis (approved):

crates/runtime/src/datafusion/functions/timestamp_from_parts.rs
```diff
@@ -0,0 +1,602 @@
...
+                        ]
+                        .concat(),
+                    ),
+                    // TIMESTAMP_FROM_PARTS( <date_expr>, <time_expr> )
+                    Coercible(vec![TypeSignatureClass::Date, TypeSignatureClass::Time]),
+                ],
+                Volatility::Immutable,
+            ),
+            aliases: vec![
+                String::from("timestamp_ntz_from_parts"),
```
Vedin, line 147: We should also support aliases without underscore to match with Snowflake. E.G. `TIMESTAMPFROMPARTS`. I think it should supported by default but I can be wrong. Can you confirm that it's supported or make alices without underscore else if I'm wrong?

Vedin (approved):
lgtm. I think all builtins are good option to use our snap test. It can be tested with 1 SQL query for every scenario. It'd be a good practice to cover all our functions with snap tests.

crates/runtime/src/datafusion/functions/timestamp_from_parts.rs
```diff
@@ -0,0 +1,602 @@
...
+    let hour = u32::try_from(hour)
+        .map_err(|_| _exec_datafusion_err!("hour value '{hour:?}' is out of range"))?;
+    let minute = u32::try_from(minute)
+        .map_err(|_| _exec_datafusion_err!("minute value '{minute:?}' is out of range"))?;
+    let seconds = u32::try_from(seconds)
+        .map_err(|_| _exec_datafusion_err!("seconds value '{seconds:?}' is out of range"))?;
+    let nano = u32::try_from(nanosecond.unwrap_or(0))
+        .map_err(|_| _exec_datafusion_err!("nanosecond value '{nanosecond:?}' is out of range"))?;
+
+    if let Some(date) = NaiveDate::from_ymd_opt(year, month, day) {
```
Vedin: Same issue with `from_ymd_opt` and `from_hms_nano_opt` as in `time_from_parts`. These functions check for out-of-range values that are supported by Snowflake. Something like `select timestamp_from_parts(2013, 15, 5, 12, 0, 0);` as well as `select timestamp_from_parts(2013, -15, -5, -12, 0, -3600);` should work correctly.

Vedin (changes requested):

crates/runtime/src/datafusion/functions/timestamp_from_parts.rs
osipovartem: Changed the logic with additional tests (compared with snowflake query api)

crates/runtime/src/datafusion/functions/timestamp_from_parts.rs
```diff
@@ -0,0 +1,599 @@
...
+    month: i32,
+    day: i32,
+    hour: i32,
+    minute: i32,
+    seconds: i32,
+    nano: Option<i64>,
+    timezone: Option<&str>,
+) -> Result<i64> {
+    let days = make_date(year, month, day)?;
+    let n_days = i64::from(days) * 86_400_000_000_000;
```
Vedin, line 354: nit: If it's reusable anywhere define constants. If it's only place then let's keep it as is.

crates/runtime/src/datafusion/functions/timestamp_from_parts.rs
```diff
@@ -0,0 +1,599 @@
...
+    let total_nanos = n_days + n_hour + n_minute + n_seconds + n_nano;
+
+    make_timestamp_from_nanoseconds(total_nanos, timezone)
+}
+
+fn make_timestamp_from_nanoseconds(nanoseconds: i64, tz: Option<&str>) -> Result<i64> {
+    let date_time = DateTime::from_timestamp_nanos(nanoseconds);
+    let timestamp = if let Some(timezone) = tz {
+        date_time
+            .with_timezone(&timezone.parse::<Tz>()?)
```
Vedin, line 368: Do we check that Tz is valid somewhere? In case of the invalid timezone we should return some understandable error.

Vedin (changes requested):

Vedin (approved):

PR #298, opened 2025-02-24
Snowflake time_from_parts udf

osipovartem:
https://docs.snowflake.com/en/sql-reference/functions/time_from_parts

added new UDF time_from_parts
Related https://github.com/Embucket/icebucket/issues/293


DanCodedThis (approved):

crates/runtime/src/datafusion/functions/time_from_parts.rs
```diff
@@ -0,0 +1,254 @@
...
+    syntax_example = "time_from_parts(<hour>, <minute>, <second> [, <nanosecond> ])",
+    sql_example = "```sql
+            > select time_from_parts(12, 34, 56, 987654321);
+            +-----------------------------------------------------------------+
+            | time_from_parts(Int64(12),Int64(34),Int64(56),Int64(987654321)) |
+            +-----------------------------------------------------------------+
+            | 1740398450.0                                                    |
+            +-----------------------------------------------------------------+
+```",
+    standard_argument(name = "str", prefix = "String"),
```
Vedin: can you please explain what's this used for? I don't get why it's "str" and prefix "String" here. Looks like it's no need in this at all because you cover all arguments below.

crates/runtime/src/datafusion/functions/time_from_parts.rs
```diff
@@ -0,0 +1,254 @@
...
+    let hour = u32::try_from(hour)
+        .map_err(|_| _exec_datafusion_err!("hour value '{hour:?}' is out of range"))?;
+    let minute = u32::try_from(minute)
+        .map_err(|_| _exec_datafusion_err!("minute value '{minute:?}' is out of range"))?;
+    let seconds = u32::try_from(seconds)
+        .map_err(|_| _exec_datafusion_err!("seconds value '{seconds:?}' is out of range"))?;
+    let nano = u32::try_from(nanosecond.unwrap_or(0))
+        .map_err(|_| _exec_datafusion_err!("nanosecond value '{nanosecond:?}' is out of range"))?;
+
+    if let Some(time) = NaiveTime::from_hms_nano_opt(hour, minute, seconds, nano) {
```
Vedin: This `from_hms_nano_opt` have these checks before transformation: 
```rust
 if (hour >= 24 || min >= 60 || sec >= 60)
            || (nano >= 1_000_000_000 && sec != 59)
            || nano >= 2_000_000_000
        {
            return None;
        }
```
So, if you pass smth like (0, 100, 0, 0) it'd return an error. Snowflake handles this correctly. Consider adding normalization before using this function or build time manually.

crates/runtime/src/datafusion/functions/time_from_parts.rs
```diff
@@ -0,0 +1,254 @@
...
+            ColumnarValue::Array(ScalarValue::from(v).to_array().unwrap())
+        }
+    }
+
+    #[allow(clippy::type_complexity)]
+    #[allow(clippy::unwrap_used)]
+    #[test]
+    fn test_time_from_parts() {
+        let args: [(i64, i64, i64, Option<i64>, i64); 4] = [
+            (12, 0, 0, None, 43_200_000_000_000),
```
Vedin: add some corner cases with out-range-values for hours/minutes/seconds to check the behavior described above.

Vedin (changes requested):

eadgbear (approved):

crates/runtime/src/datafusion/functions/time_from_parts.rs
osipovartem: updated this section

crates/runtime/src/datafusion/functions/time_from_parts.rs
osipovartem: Added

crates/runtime/src/datafusion/functions/time_from_parts.rs
```diff
@@ -0,0 +1,239 @@
...
+    syntax_example = "time_from_parts(<hour>, <minute>, <second> [, <nanosecond> ])",
+    sql_example = "```sql
+            > select time_from_parts(12, 34, 56, 987654321);
+            +-----------------------------------------------------------------+
+            | time_from_parts(Int64(12),Int64(34),Int64(56),Int64(987654321)) |
+            +-----------------------------------------------------------------+
+            | 1740398450.0                                                    |
+            +-----------------------------------------------------------------+
+```",
+    standard_argument(name = "str", prefix = "String"),
```
Vedin: it's still here. Please remove before merge

crates/runtime/src/datafusion/functions/timestamp_from_parts.rs
```diff
@@ -275,9 +275,9 @@ fn timestamps_from_components(
     let years = to_primitive_array::<Int32Type>(&years.cast_to(&Int32, None)?)?;
     let months = to_primitive_array::<Int32Type>(&months.cast_to(&Int32, None)?)?;
     let days = to_primitive_array::<Int32Type>(&days.cast_to(&Int32, None)?)?;
-    let hours = to_primitive_array::<Int32Type>(&hours.cast_to(&Int32, None)?)?;
-    let minutes = to_primitive_array::<Int32Type>(&minutes.cast_to(&Int32, None)?)?;
-    let seconds = to_primitive_array::<Int32Type>(&seconds.cast_to(&Int32, None)?)?;
+    let hours = to_primitive_array::<Int64Type>(&hours.cast_to(&Int64, None)?)?;
```
Vedin, line 278: due to cast to `Int64` instead of `Int32` we now support a wider range of values. E.G. Snowflake will fail on `time_from_parts(2147483648, 100, 0)` but I believe we won't. I'm fine with this it just worth to mention that we'd have different behaviour in this rare cases.

Vedin (approved):

crates/runtime/src/datafusion/functions/timestamp_from_parts.rs
osipovartem: To get nanoseconds we have to cast to i64
I changed this from Int32 to Int64 to not convert it twice. If the value will be out of range Datetime creation will fail with the value out of range error.

PR #299, opened 2025-02-24
snowflake date_from_parts udf

osipovartem:
https://docs.snowflake.com/en/sql-reference/functions/date_from_parts

added new UDF date_from_parts
Related https://github.com/Embucket/icebucket/issues/293

This UDF almost the same as **[make_date](https://github.com/apache/datafusion/blob/main/datafusion/functions/src/datetime/make_date.rs)** 
**but with an important difference, it supports negative values**

PR #288, opened 2025-02-20
Set warehouse location based on storage profile path

osipovartem:
Related to #268 
- Set full warehouse location instead of relative path prefix/id 
- fixed incorrect path if the prefix was empty
```python
{
    "id": "0e9dc17a-899d-4e2c-92c1-a4d792e71a6a",
    "name": "snowplow",
    "keyPrefix": "warehouse-prefix",
    "location": "warehouse-prefix/id",
}
->
{
    "id": "0e9dc17a-899d-4e2c-92c1-a4d792e71a6a",
    "name": "snowplow",
    "keyPrefix": "warehouse-prefix",
    "location": "s3://artemembucket",
}
{
    "id": "0e9dc17a-899d-4e2c-92c1-a4d792e71a6a",
    "name": "snowplow",
    "keyPrefix": "warehouse-prefix",
    "location": "file://<full_path>",
}



{
    "id": "0e9dc17a-899d-4e2c-92c1-a4d792e71a6a",
    "name": "snowplow",
    "keyPrefix": "",
    "location": "/id",
}
->
{
    "id": "0e9dc17a-899d-4e2c-92c1-a4d792e71a6a",
    "name": "snowplow",
    "keyPrefix": "",
    "location": "s3://artemembucket",
}
{
    "id": "0e9dc17a-899d-4e2c-92c1-a4d792e71a6a",
    "name": "snowplow",
    "keyPrefix": "",
    "location": "file://<full_path>",
}
```

PR #312, opened 2025-02-27
(WIP) Metastore + Unified Query 1.0

eadgbear:
- **Initial metastore crate**
- **Metastore implementation**
- **Upgrading Slatedb to 0.4, adding snapshot tests**
- **Checkpoint commit**
- **Checkpoint before crate re-org**
- **Fix timestamp displaying format (#279)**
- **in login request use database_name instead of warehouse (#277)**
- **Adding missing licenses + license check (#278)**
- **Fix IPC arrow format issue (#280)**
- **Control plane tests (batch 3) (#274)**
- **Move DBT_SERIALIZATION_FORMAT to cli (#286)**
- **Update query runner with session id (#285)**
- **Issue 115 Arrow support: replace uints by ints in resulted RecordBatch (#287)**
- **Set warehouse location based on storage profile path (#288)**
- **map float types to real logical_type (#289)**
- **Snowflake timestamp_from_parts udf (#297)**
- **Snowflake time_from_parts udf (#298)**
- **snowflake date_from_parts udf (#299)**
- **Implement extraction of the corresponding date part from a date or timestamp (#302)**
- **Add geospatial udfs (#304)**
- **Fix object store for custom endpoint (#295)**
- **Unified query, execution, catalog**


Vedin:
Are these data and metadata files used somewhere or are they here by accident? If it's used in some tests please make other PR for them. It's kinda impossible to load diff with such large data files.

.vscode/launch.json
YaroslavLitvinov, line 1: Added by accident?

osipovartem (approved):

data/sdb/manifest/00000000000000000086.manifest
YaroslavLitvinov, line 1: data, data2, prefix folders containing data

YaroslavLitvinov:
Please add related PR for iceberg-rust.

osipovartem (approved):

crates/metastore/src/metastore.rs
```diff
@@ -0,0 +1,1308 @@
...
+const KEY_DATABASES: &str = "databases";
+const KEY_DATABASE: &str = "db";
+const KEY_SCHEMAS: &str = "schemas";
+const KEY_SCHEMA: &str = "sch";
+const KEY_TABLES: &str = "tables";
+const KEY_TABLE: &str = "tbl";
+
+pub struct SlateDBMetastore {
+    db: Db,
+    object_store_cache: Mutex<HashMap<IceBucketVolumeIdent, Arc<dyn ObjectStore>>>,
```
Vedin: why not use concurrent hashmap implementation `DashMap` instead of `Mutex<HashMap>`? Are there any benefits I'm missing? 
Also, I didn't find anything about whether we control the size of this structure anywhere.

crates/metastore/src/metastore.rs
```diff
@@ -0,0 +1,1308 @@
...
+                let prefix = volume.prefix();
+                format!(
+                    "{prefix}/{}/{}/{}",
+                    ident.database, ident.schema, ident.table
+                )
+            };
+
+            let metadata_part = format!("metadata/{}", Self::generate_metadata_filename());
+            dbg!(&table_location);
+            /*let table_creation = {
```
Vedin: nit: remove this comment

crates/metastore/src/metastore.rs
```diff
@@ -0,0 +1,1308 @@
...
+        if let Some(_schema) = self.get_schema(&ident.clone().into()).await? {
+            let key = format!(
+                "{KEY_TABLE}/{}/{}/{}",
+                ident.database, ident.schema, ident.table
+            );
+
+            // This is duplicating the behavior of url_for_table,
+            // but since the table won't exist yet we have to create it here
+            let table_location = if table.is_temporary.unwrap_or_default() {
+                let volume_ident: String = table.volume_ident.as_ref().map_or_else(
```
Vedin, line 534: Please briefly explain how it'll work in case of existing volume. E.G. We create some iceberg table in `db.schema`. Then we try to create the temporary table in `db.schema`. Will we create a new volume in this case anyway?

crates/runtime/src/http/ui/models/mod.rs
Vedin, line 1: do we need this file in the future? Why is everything commented out?

Vedin (approved):
Awesome! 
I didn't deeply review everything. I tried to understand what changed and how it works now. I left some unblocking questions mostly to get better understanding

crates/metastore/src/models/mod.rs
```diff
@@ -0,0 +1,78 @@
...
+pub use volumes::*;
+
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, utoipa::ToSchema)]
+pub struct RwObject<T>
+where
+    T: Eq + PartialEq,
+{
+    #[serde(flatten)]
+    pub data: T,
+    pub created_at: NaiveDateTime,
```
YaroslavLitvinov, line 40: Still see this as a potential issue in future, when we need to use datetime as a key for slatedb.scan function. In addition to static keys from metastore. This datetime has no information about timezone, and assumptions about timezone must be made from server's localtime. Probably there expected  some tricky conversion into DateTime<Utc>. As RwObject is forced by design to be a logical parent of any data in metastore, and therefore datetime from there should be flexible enough and convenient to use.

crates/metastore/src/metastore.rs
YaroslavLitvinov: This hash map actually has from one to couple of items most of the time. It seemed to be a not an actual caching.

crates/metastore/src/metastore.rs
eadgbear: Done, thanks

crates/metastore/src/metastore.rs
eadgbear: I switched to Dashmap, thanks Denys.

It's caching because in the case of temporary tables we don't recreate the object store every time, that would destroy the files in the object store

crates/runtime/src/http/ui/models/mod.rs
eadgbear: They'll be added back in soon as I get this merged

crates/metastore/src/metastore.rs
eadgbear: Tables support overrides and individual volumes. So Table `x.y.z`, when temporary, has a separate volume than the volume for the database `x`

crates/metastore/src/models/mod.rs
eadgbear: Yep cool, going to close this as it's not relevant to the metastore.

osipovartem (approved):

data/sdb/manifest/00000000000000000086.manifest
eadgbear: Yep, my bad. Fixed

PR #314, opened 2025-02-28
Using preprocess_query as assosiated function

Vedin:
After recent changes preprocess_query become assosiated function instead of method. Fix usage of it in our service.


DanCodedThis (approved):
Strange, I fixed this in my PR, after Artem's PR

Issue #305, opened 2025-02-26
Geospatial geometry constructors


osipovartem:
https://github.com/geoarrow/geoarrow-rs/tree/main/rust/geodatafusion#geometry-constructors

PR #319, opened 2025-03-04
Add ST_MakePolygon, st_polygon

osipovartem:
- Added ST_MakePolygon, ST_Polygon
- Added and fixed ST_Makeline for handle both scalars and arrays
Related to #305 

Docs 
https://docs.snowflake.com/en/sql-reference/functions/st_makepolygon

### DIFF
--- a/crates/runtime/src/datafusion/functions/geospatial/constructors/line_string.rs
+++ b/crates/runtime/src/datafusion/functions/geospatial/constructors/line_string.rs
@@ -25,6 +25,8 @@ use datafusion::logical_expr::{
     ColumnarValue, ScalarUDFImpl, Signature, TypeSignature, Volatility,
 };
 use datafusion_common::{DataFusionError, Result};
+use datafusion_doc::Documentation;
+use datafusion_expr::scalar_doc_sections::DOC_SECTION_OTHER;
 use geo_traits::{LineStringTrait, MultiPointTrait, PointTrait};
 use geoarrow::array::{
     AsNativeArray, CoordType, LineStringArray, LineStringBuilder, MultiPointArray, PointArray,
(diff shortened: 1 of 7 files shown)


crates/runtime/src/datafusion/functions/geospatial/constructors/line_string.rs
```diff
@@ -133,54 +134,58 @@ unsafe fn make_line(args: &[ColumnarValue]) -> Result<ColumnarValue> {
...
-    builder
-        .linestring_begin(true, points.len(), 0)
-        .map_err(|e| DataFusionError::Execution(format!("Failed to start linestring: {e}")))?;
 
-    for p in points.iter_geo_values() {
+    for (idx, point) in points.iter_geo_values().enumerate() {
+        builder
+            .linestring_begin(true, 1, idx)
+            .map_err(|e| DataFusionError::Execution(format!("Failed to start linestring: {e}")))?;
         unsafe {
```
DanCodedThis, line 190: Unsafe?

DanCodedThis (commented):
LGTM, other than the `unsafe` stuff, not against it just curious

DanCodedThis (approved):

crates/runtime/src/datafusion/functions/geospatial/constructors/line_string.rs
osipovartem: Yes, since
    /// Push a raw coordinate to the underlying coordinate array.
    ///
    /// # Safety
    ///
    /// This is marked as unsafe because care must be taken to ensure that pushing raw coordinates
    /// to the array upholds the necessary invariants of the array.

Issue #306, opened 2025-02-26
api: worksheets UI handlers


YaroslavLitvinov:
Worksheets are embucket / ui entities, need to be created before running queries.

/worksheets POST, GET <- create get worksheet
/ui/worksheets/{wid} GET, PATCH, DELETE

PR #320, opened 2025-03-05
issue #306  woksheets queries endpoint

### LINKED ISSUE #104: api: Add query history endpoint
`/ui/worksheets/{wid}/queries` POST <- this is where /ui/query moves
`/ui/worksheets/{wid}/queries` GET <- access query history now worksheet related

old names are: `/ui/query` & `/ui/history`

YaroslavLitvinov:
Closes #104 #306

Separate history crate adds another pluggable store reusing the same single slatedb instance.


eadgbear (changes requested):
1. We don't need the `catalog` crate anymore
2. No modifications to the `control_plane` crate, it doesn't exist anymore

The changes to `utils` are good, though

eadgbear:
Is this small changes to my metastore branch or is it your query history changes? Let's not confound both

crates/runtime/src/execution/service.rs
```diff
@@ -38,14 +39,20 @@ use super::error::{self as ex_error, ExecutionError, ExecutionResult};
...
+    qhistory: Arc<dyn QHistoryApi>,
     df_sessions: Arc<RwLock<HashMap<String, Arc<IceBucketUserSession>>>>,
     config: Config,
 }
 
 impl ExecutionService {
-    pub fn new(metastore: Arc<dyn Metastore>, config: Config) -> Self {
+    pub fn new(
+        metastore: Arc<dyn Metastore>,
+        qhistory: Arc<dyn QHistoryApi>,
```
eadgbear: I wouldn't call it `QHistoryApi`, simply `QueryHistory`

crates/history/src/store.rs
```diff
@@ -0,0 +1,140 @@
...
+}
+
+pub type QHistoryResult<T> = std::result::Result<T, QHistoryError>;
+
+#[async_trait]
+pub trait QHistoryApi: Send + Sync {
+    async fn add_history_item(&self, item: HistoryItem) -> QHistoryResult<()>;
+    async fn query_history(
+        &self,
+        cursor: Option<String>,
```
eadgbear: `cursor` is a bit specific to slatedb. We should change this to be something more abstract, such as a timestamp.

Cargo.toml
```diff
@@ -3,6 +3,7 @@ default-members = ["bin/bucketd"]
 members = [
   "bin/bucketd",
   "crates/metastore",
+  "crates/history",
```
osipovartem, line 6: Do we need a separate crate for single endpoint?
Can we put the logic near /ui/query

crates/history/src/store.rs
YaroslavLitvinov: This cursor is introduced in QueryHistory store as a high level api element for ui handlers support, and cursor is needed for `cursor based pagination`. 
It's not specific to slatedb, but it could contains some timestamp underneath but interface doesn't expose such implementation details that can and will be changed in future.

Cargo.toml
YaroslavLitvinov: Main reason for having code in separate crate as it shortens development cycle - tests build faster. 
Logically it's the same thing as metastore, and it makes sense to be a separate crate too. 
That is the concern having more crates than less?

crates/history/src/store.rs
```diff
@@ -0,0 +1,140 @@
...
+use async_trait::async_trait;
+use bytes::Bytes;
+use icebucket_utils::iterable::IterableEntity;
+use icebucket_utils::Db;
+use snafu::prelude::*;
+#[cfg(test)]
+use std::sync::Arc;
+
+#[derive(Snafu, Debug)]
+pub enum QHistoryError {
```
eadgbear: Spell these names out fully
```suggestion
pub enum QueryHistoryError {
```

crates/runtime/Cargo.toml
```diff
@@ -48,6 +48,7 @@ icebucket_utils = { version = "0.1.0", path = "../utils" }
 indexmap = { version = "2.7.1" }
 itertools = "0.14.0"
 object_store = { workspace = true }
+history = { workspace= true }
```
eadgbear: Nit: This shouldn't be a separate crate, this should be a part of the runtime crate

crates/history/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "history"
```
eadgbear: If you're insisting on making this a separate crate, it should follow the naming convention of `icebucket_<crate_name>`

crates/history/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+version = "0.1.0"
+edition = "2021"
+license.workspace = true
```
eadgbear: Remove this line

crates/history/src/history_item.rs
```diff
@@ -0,0 +1,82 @@
...
+    pub end_time: DateTime<Utc>,
+    pub duration_ms: i64,
+    pub result_count: i64,
+    pub status_code: u16,
+    pub error: Option<String>,
+}
+
+impl HistoryItem {
+    #[must_use]
+    pub fn before_started(query: &str, id: Option<Uuid>, start_time: Option<DateTime<Utc>>) -> HistoryItem {
```
eadgbear: Rename this method
```suggestion
    pub fn query_start(query: &str, id: Option<Uuid>, start_time: Option<DateTime<Utc>>) -> HistoryItem {
```

crates/history/src/history_item.rs
```diff
@@ -0,0 +1,82 @@
...
+            start_time,
+            end_time: start_time,
+            status_code: 200,
+            duration_ms: 0,
+            result_count: 0,
+error: None,
+        }
+    }
+
+    pub fn set_finished(&mut self, result_count: i64, end_time: Option<DateTime<Utc>>) {
```
eadgbear:
```suggestion
    pub fn query_finished(&mut self, result_count: i64, end_time: Option<DateTime<Utc>>) {
```

crates/history/src/history_item.rs
```diff
@@ -0,0 +1,82 @@
...
+        self.end_time = Utc::now();
+        self.duration_ms = self.end_time
+            .signed_duration_since(self.start_time)
+            .num_milliseconds();
+        if let Some(end_time) = end_time {
+            self.end_time = end_time;
+        }
+    }
+
+    pub fn set_error(&mut self, error: String, error_code: u16) {
```
eadgbear:
```suggestion
    pub fn query_error(&mut self, error: String, error_code: u16) {
```

eadgbear: Also where does this error code come from?

crates/history/src/history_item.rs
```diff
@@ -0,0 +1,82 @@
...
+        self.error = Some(error);
+        self.status_code = error_code;
+    }
+}
+
+impl IterableEntity for HistoryItem {
+    type Cursor = i64;
+    const PREFIX: &[u8] = b"hi.";
```
eadgbear: This is slatedb specific, abstract it

crates/runtime/src/http/ui/handlers/history.rs
```diff
@@ -0,0 +1,140 @@
...
+use icebucket_utils::iterable::IterableEntity;
+use serde::{Deserialize, Serialize};
+use snafu::prelude::*;
+use snafu::ResultExt;
+use std::time::Instant;
+use utoipa::{OpenApi, ToSchema};
+
+#[derive(Snafu, Debug)]
+#[snafu(visibility(pub))]
+pub enum HistoryError {
```
eadgbear: Follow the conventions in the rest of the HTTP crate, create a separate `error` module for this

crates/runtime/src/http/ui/handlers/history.rs
```diff
@@ -0,0 +1,140 @@
...
+    tags = ["history"],
+    responses(
+        (status = 200, description = "Returns result of the history", body = HistoryResponse),
+        (status = 500, description = "Internal server error", body = ErrorResponse),
+    )
+)]
+#[tracing::instrument(level = "debug", skip(state), err, ret(level = tracing::Level::TRACE))]
+// Add time sql took
+pub async fn history(
+    DFSessionId(session_id): DFSessionId,
```
eadgbear: You don't need a session for this if you're not using a query history table

crates/runtime/src/http/state.rs
```diff
@@ -26,6 +27,7 @@ use crate::execution::service::ExecutionService;
 #[derive(Clone)]
 pub struct AppState {
     pub metastore: Arc<dyn Metastore + Send + Sync>,
+    pub qhistory: Arc<dyn QHistoryApi + Send + Sync>,
```
eadgbear:
```suggestion
    pub query_history: Arc<dyn QHistoryApi + Send + Sync>,
```

crates/history/src/history_item.rs
```diff
@@ -0,0 +1,88 @@
...
+        self.duration_ms = self
+            .end_time
+            .signed_duration_since(self.start_time)
+            .num_milliseconds();
+        if let Some(end_time) = end_time {
+            self.end_time = end_time;
+        }
+    }
+
+    pub fn set_finished_with_error(&mut self, error: String, error_code: u16) {
```
eadgbear:
```suggestion
    pub fn query_error(&mut self, error: String, error_code: u16) {
```

Also what is this error code? We're not using a `u16` for error code anywhere

crates/history/src/history_item.rs
```diff
@@ -0,0 +1,88 @@
...
+    pub end_time: DateTime<Utc>,
+    pub duration_ms: i64,
+    pub result_count: i64,
+    pub status_code: u16,
+    pub error: Option<String>,
+}
+
+impl HistoryItem {
+    #[must_use]
+    pub fn before_started(
```
eadgbear:
```suggestion
    pub fn query_started(
```

Cargo.toml
eadgbear: The metastore is a separate crate because it could conceivably operate independently of the runtime. If you think the history can operate independently of the UI, then keep it a separate crate. I don't think it's a separate crate.

crates/history/src/store.rs
eadgbear: A timestamp offset would be an easier abstraction and a clearer use-case from a user perspective (Ask yourself as a user: How do I know what the cursor is?). Change it to a timestamp offset.

eadgbear: Alternatively rename it to be clearer

crates/history/src/store.rs
```diff
@@ -0,0 +1,143 @@
...
+use icebucket_utils::iterable::IterableEntity;
+use icebucket_utils::Db;
+use snafu::prelude::*;
+#[cfg(test)]
+use std::sync::Arc;
+
+#[derive(Snafu, Debug)]
+pub enum QueryHistoryError {
+    #[snafu(display("Error adding query history: {source}"))]
+    QHistoryAdd { source: icebucket_utils::Error },
```
eadgbear:
```suggestion
    HistoryAdd { source: icebucket_utils::Error },
```

The context is already in the `enum` name

crates/history/src/store.rs
```diff
@@ -0,0 +1,143 @@
...
+
+    #[snafu(display("Error getting query history: {source}"))]
+    QHistoryGet { source: icebucket_utils::Error },
```
eadgbear:
```suggestion
    HistoryGet { source: icebucket_utils::Error },
```

The context is in the `enum` name

crates/runtime/src/execution/service.rs
```diff
@@ -87,14 +94,35 @@ impl ExecutionService {
...
+
+        match &res {
+            Ok(recs) => {
+                let result_count = i64::try_from(recs.0.len()).unwrap_or(0);
+                history_item.set_finished(result_count, None);
+                // TODO: add result records, perhaps using records_to_json_string
+            }
+            Err(err) => {
+                history_item.set_finished_with_error(err.to_string(), 0);
+                // TODO: http code
```
eadgbear: Query history shouldn't have any concept of the HTTP code. Queries don't care about HTTP, the HTTP is just an interface to run queries. Change the error state to include the string version of the error instead.

crates/runtime/src/http/state.rs
```diff
@@ -26,6 +27,7 @@ use crate::execution::service::ExecutionService;
 #[derive(Clone)]
 pub struct AppState {
     pub metastore: Arc<dyn Metastore + Send + Sync>,
+    pub qhistory: Arc<dyn QueryHistory + Send + Sync>,
```
eadgbear:
```suggestion
    pub query_history: Arc<dyn QueryHistory + Send + Sync>,
```

crates/runtime/src/http/ui/handlers/history.rs
```diff
@@ -0,0 +1,156 @@
...
+// for tracing logs
+impl Display for HistoryHandlerError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result<(), Error> {
+        self.0.fmt(f)
+    }
+}
+
+impl HistoryHandlerError {
+    #[must_use]
+    pub const fn map_error_to_http_error(&self) -> StatusCode {
```
eadgbear: What's the point of this?

crates/runtime/src/http/ui/handlers/history.rs
```diff
@@ -0,0 +1,156 @@
...
+use history::HistoryItem;
+use http::status::StatusCode;
+use icebucket_utils::iterable::IterableEntity;
+use serde::{Deserialize, Serialize};
+use std::fmt::{Display, Error, Formatter};
+use std::time::Instant;
+use tracing;
+use utoipa::{OpenApi, ToSchema};
+
+pub struct HistoryHandlerError(store::QueryHistoryError);
```
eadgbear: Please follow the convention for the other `http` modules. Create an `error` module, use `snafu`, etc.

crates/runtime/src/lib.rs
```diff
@@ -42,14 +43,15 @@ pub async fn run_icebucket(
...
             )
             .await
             .map_err(Box::new)?,
         ))
     };
 
-    let metastore = Arc::new(SlateDBMetastore::new(db));
-    let app = make_icebucket_app(metastore, &config.web)?;
+    let metastore = Arc::new(SlateDBMetastore::new(db.clone()));
+    let qhistory = Arc::new(QueryHistoryStore::new(db));
```
eadgbear: Do we want to use the same DB as the metastore? Or should history be under a different prefix?

eadgbear (dismissed):

crates/history/src/store.rs
YaroslavLitvinov: Cursor is a thing user gets from server and uses it as is. Cursor is an abstraction and user shouldn't care what's inside.

Cargo.toml
YaroslavLitvinov: Functionality and name of "history" crate is a subject to change. This implements a separate store that have it's own API.  If we would have a different crate something like "store", that would keep all the stores I would agree to move. But I don't see any sense moving it to runtime, at least don't understand why. What is the description of "runtime" after all?

crates/runtime/src/lib.rs
YaroslavLitvinov: The same database, as per our discussion in Eng chat

crates/runtime/src/http/ui/handlers/history.rs
YaroslavLitvinov: Errors just reused here, no need to define new Errors.

crates/utils/src/lib.rs
```diff
@@ -139,7 +146,7 @@ impl Db {
     /// Returns a `DbError` if the underlying database operation fails.
     /// Returns a `DeserializeError` if the value cannot be deserialized from JSON.
     #[allow(clippy::unwrap_used)]
-    pub async fn list_objects<T: for<'de> serde::de::Deserialize<'de>>(
+    pub async fn list_objects<T: Send + for<'de> serde::de::Deserialize<'de>>(
```
DanCodedThis, line 149: No sync, by design?

DanCodedThis (approved):
LGTM

crates/utils/src/lib.rs
YaroslavLitvinov: This was added to avoid clippy error about unsatisfied Sync trait

YaroslavLitvinov:
Tests not related to code in this branch:

failures:
    execution::dedicated_executor::tests::test_mock_store_copy_asserts
    execution::dedicated_executor::tests::test_mock_store_copy_if_not_exists_asserts
    execution::dedicated_executor::tests::test_mock_store_delete_asserts
    execution::dedicated_executor::tests::test_mock_store_get_asserts
    execution::dedicated_executor::tests::test_mock_store_list_asserts
    execution::dedicated_executor::tests::test_mock_store_list_with_delimiter_asserts
    execution::dedicated_executor::tests::test_mock_store_put_asserts
    execution::dedicated_executor::tests::test_mock_store_put_multipart_asserts
    execution::dedicated_executor::tests::test_that_default_executor_prevents_io
    http::tests::databases::test_ui_databases_metastore_update_bug
    http::tests::query::test_parallel_queries
    tests::queries::query_select_date_add_diff

test result: FAILED. 69 passed; 12 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.36s

rampage644 (approved):

PR #322, opened 2025-03-05
Add ST_PointN and ST_Endpoint

osipovartem:
https://docs.snowflake.com/en/sql-reference/functions/st_endpoint
ST_StartPoint already implemented

Related to #323

### DIFF
--- a/crates/runtime/src/datafusion/functions/geospatial/accessors/line_string.rs
+++ b/crates/runtime/src/datafusion/functions/geospatial/accessors/line_string.rs
@@ -1,0 +1,285 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
+// "License"); you may not use this file except in compliance
+// with the License.  You may obtain a copy of the License at
+//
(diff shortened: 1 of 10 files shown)


crates/runtime/src/datafusion/functions/geospatial/accessors/line_string.rs
```diff
@@ -0,0 +1,360 @@
...
+
+    fn signature(&self) -> &Signature {
+        &self.signature
+    }
+
+    fn return_type(&self, _arg_types: &[DataType]) -> Result<DataType> {
+        Ok(POINT2D_TYPE.into())
+    }
+
+    fn invoke(&self, args: &[ColumnarValue]) -> Result<ColumnarValue> {
```
rampage644, line 119: I think it `invoke` is deprecated since 42, `invoke_with_args` is recommended to use

crates/runtime/src/datafusion/functions/geospatial/accessors/line_string.rs
```diff
@@ -0,0 +1,360 @@
...
```
rampage644, line 178: same as above

rampage644 (approved):

crates/runtime/src/datafusion/functions/geospatial/accessors/line_string.rs
osipovartem: will fix all in the next PR

PR #324, opened 2025-03-05
Add St_x, st_y, st_srid

osipovartem:
https://docs.snowflake.com/en/sql-reference/functions/st_x
https://docs.snowflake.com/en/sql-reference/functions/st_y
https://docs.snowflake.com/en/sql-reference/functions/st_srid

Related to #323

### DIFF
--- a/crates/runtime/src/datafusion/functions/geospatial/accessors/point.rs
+++ b/crates/runtime/src/datafusion/functions/geospatial/accessors/point.rs
@@ -1,0 +1,236 @@
+// Licensed to the Apache Software Foundation (ASF) under one
+// or more contributor license agreements.  See the NOTICE file
+// distributed with this work for additional information
+// regarding copyright ownership.  The ASF licenses this file
+// to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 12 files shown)


crates/runtime/src/datafusion/functions/geospatial/accessors/point.rs
```diff
@@ -0,0 +1,236 @@
...
+
+    fn signature(&self) -> &Signature {
+        &self.signature
+    }
+
+    fn return_type(&self, _arg_types: &[DataType]) -> Result<DataType> {
+        Ok(Float64)
+    }
+
+    fn invoke(&self, args: &[ColumnarValue]) -> Result<ColumnarValue> {
```
rampage644, line 69: `invoke_with_args`

rampage644 (approved):

crates/runtime/src/datafusion/functions/geospatial/accessors/point.rs
osipovartem: Will fix all in the next PR

PR #329, opened 2025-03-07
Add St_area + st_distance

osipovartem:
https://docs.snowflake.com/en/sql-reference/functions/st_distance
https://docs.snowflake.com/en/sql-reference/functions/st_area

Related to #326


crates/runtime/src/datafusion/functions/geospatial/measurement/distance.rs
```diff
@@ -0,0 +1,282 @@
...
+            }
+            _ => {
+                return Err(DataFusionError::Execution(
+                    "ST_Distance does not support this rhs geometry type".to_string(),
+                ))
+            }
+        }
+    };
+}
+fn area(args: &[ColumnarValue]) -> Result<ColumnarValue> {
```
DanCodedThis: Accidentally naming of `fn distance` as `fn area`

DanCodedThis (changes requested):
Accidentally naming of `fn distance` as `fn area`

DanCodedThis (approved):
LGTM

PR #331, opened 2025-03-10
SlateDB range scans for metastore impl

DanCodedThis:


DanCodedThis:
- To add some clarity, for some reason the **utils** crate was having `.` as a separator between the naming of db, schemas, tbls, etc. 
- But the **metastore** had `/` as separator. I changed the **utils** to `/` so it would be the same. 
- You can change the separator however is needed, as long as you keep both **utils** and **metastore** in sync.

eadgbear:
@DanCodedThis Can you add to the unit tests? Just for my sanity

DanCodedThis:
> @DanCodedThis Can you add to the unit tests? Just for my sanity

- The checks for lists was already implimented in tests, in both **utils** and **metastore**
- I specifically tested with all the tests in **metastore** and **utils** which rely on this
- Fixed a lot of them to work with the new change
- All the tests pass and work correctly
- I don't think there is anything to add here, feel free to correct

eadgbear:
Nope that sounds good to me

crates/utils/src/lib.rs
```diff
@@ -134,55 +135,68 @@ impl Db {
...
-        .await
+    pub async fn list_objects<T: for<'de> serde::de::Deserialize<'de>>(
+        &self,
+        key: &str,
+    ) -> Result<Vec<T>> {
+        dbg!(&key);
+        let start = format!("{key}/");
+        let end = format!("{key}/\x7F");
+        let range = Bytes::from(start)..Bytes::from(end);
+        let mut test = self.0.scan(range).await.unwrap();
```
YaroslavLitvinov: Pls fix this clippy error: [clippy] crates/utils/src/lib.rs#L146

crates/metastore/src/metastore.rs
```diff
@@ -192,24 +188,28 @@ impl SlateDBMetastore {
...
-            .await
-            .context(metastore_error::UtilSlateDBSnafu)?
-            .unwrap_or_default();
-        let futures = keys.iter().map(|key| self.db.get(key)).collect::<Vec<_>>();
-        let results = futures::future::try_join_all(futures)
+            .list_objects(list_key)
             .await
             .context(metastore_error::UtilSlateDBSnafu)?;
-        let entities = results.into_iter().flatten().collect::<Vec<_>>();
+        // let keys = self
```
YaroslavLitvinov: Remove old implementations you commented, in this place and in other. We have it's in the history.

YaroslavLitvinov:
> > @DanCodedThis Can you add to the unit tests? Just for my sanity
> 
>     * The checks for lists was already implimented in tests, in both **utils** and **metastore**
>     * I specifically tested with all the tests in **metastore** and **utils** which rely on this
>     * Fixed a lot of them to work with the new change
>     * All the tests pass and work correctly
>     * I don't think there is anything to add here, feel free to correct

@DanCodedThis You can also check coverage for functionality you changed to find which execution branches are not covered by tests in order to have some formal criteria of acceptance. For instance following will generate a coverage report for whole project:
```
cargo install cargo-tarpaulin
cargo tarpaulin --out html
```

I also think if we cover metastore with enough amount of tests, then we don't need to cover new functions in utils crate by unit tests in this PR, as after merging query history hopefully we will generalize utilization of scan function, in order to have a single scan implementation we choose.

YaroslavLitvinov (approved):

PR #333, opened 2025-03-12
Updating bench branch with latest main branch code

eadgbear:


osipovartem (approved):

rest_examples.http
```diff
@@ -0,0 +1,166 @@
...
+# List all schemas
+# @name listSchemas
+
+GET {{ib_server}}/v1/metastore/databases/{{listDatabases.response.body.$[0].ident}}/schemas
+
+###
+# Create a schema
+# @name createSchema
+
+POST {{ib_server}}/v1/metastore/databases/{{listDatabases.response.body.$[0].ident}}/schemas
```
osipovartem, line 74: Do we need to duplicate listDatabases.response.body.$[0].ident in request path and body?

eadgbear: This is just how the schema objects are created currently. The `ident` object currently needs the database name and schema name. 

In the future we can improve the metastore API to only accept the necessary fields

PR #347, opened 2025-03-14
Schemas OpenAPI spec update

DanCodedThis:


DanCodedThis:
Will add this with tests for these endpoints later

PR #360, opened 2025-03-19
Fix iceberg catalog 

### LINKED ISSUE #349: Restore iceberg catalog endpoints

osipovartem:
- fixed iceberg-rust [optional table commit fields](https://github.com/Embucket/iceberg-rust/pull/22)
- fixed path for metadata location during iceberg table update
- fixed AssertRefSnapshotId check
- fixed panic with missing tokio runtime timers
- fixed missing object store for external tables
Closes #349

### DIFF
--- a/crates/runtime/src/execution/query.rs
+++ b/crates/runtime/src/execution/query.rs
@@ -910,23 +910,7 @@ impl IceBucketQuery {
     #[tracing::instrument(level = "trace", skip(self), err, ret)]
     pub async fn execute_with_custom_plan(&self, query: &str) -> ExecutionResult<Vec<RecordBatch>> {
         let plan = self.get_custom_logical_plan(query).await?;
-        let session = self.session.clone();
-        let stream = self
-            .session
-            .executor
-            .spawn(async move {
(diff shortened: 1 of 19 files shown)


Vedin (approved):

crates/runtime/src/execution/query.rs
```diff
@@ -910,23 +910,7 @@ impl IceBucketQuery {
...
-                    .execute_logical_plan(plan)
-                    .await
-                    .context(super::error::DataFusionSnafu)?
-                    .collect()
-                    .await
-                    .context(super::error::DataFusionSnafu)
-            })
-            .await
-            .context(super::error::JobSnafu)??;
-        Ok(stream)
```
DanCodedThis, line 929: Is this removal correct?

DanCodedThis (approved):
Other than that (which would be blocking), LGTM

crates/runtime/src/execution/query.rs
osipovartem: Yes, this code was duplicated

PR #367, opened 2025-03-21
/ui/databases-navigation

### LINKED ISSUE #359: api: /ui/databases-navigation
/ui/databases-navigation

DanCodedThis:
Closes #359


DanCodedThis:
- This is a simple impl
- Later on we can incorporate custom scan function to get everything at ones
- For example get all the `tbl. .. tbl.(last symbol del)`, and get the schema name and database name from just one loop, and get all the remaining empty schemas (without tables) and all the remaining empty databases (without schemas) through `scan iter seek` to seek afterward and before the table names or smth similar

crates/runtime/src/http/router.rs
```diff
@@ -60,6 +61,8 @@ pub fn create_app(state: AppState) -> Router {
         .merge_from(DatabasesApiDoc::openapi())
         // .merge_from(TableApiDoc::openapi())
         .merge_from(SchemasApiDoc::openapi())
+        .merge_from(QueryApiDoc::openapi())
```
YaroslavLitvinov, line 64: duplicate nesting

YaroslavLitvinov (changes requested):
Add tests for list_tables

YaroslavLitvinov:
> * This is a simple impl
> 
>     * Later on we can incorporate custom scan function to get everything at ones
> 
>     * For example get all the `tbl. .. tbl.(last symbol del)`, and get the schema name and database name from just one loop, and get all the remaining empty schemas (without tables) and all the remaining empty databases (without schemas) through `scan iter seek` to seek afterward and before the table names or smth similar

Actually you don't need a custom scan for this, just scan and full range iteration aka full scan. Read scans are cheap, to run it 3 times separately on group of keys: databases, schemas, tables. It's better do in separate scan than scanning all irrelevant query history items keys.

Regarding current implementation, probably when amount of databases, schemas will grow significantly then performance hit will be noticable, but for MVP it's ok.

BTW slatedb doesn't support backward seek.

YaroslavLitvinov (approved):

crates/runtime/src/http/router.rs
DanCodedThis: done

PR #371, opened 2025-03-24
Issues/366 UI worksheets queries bugfixes + runtime crate refactoring

### LINKED ISSUE #366: api: ui worksheets queries endpoint fixes
* Optional payload/response parameters are missing in openapi spec
* Irrelevant response code defined in openapi spec

YaroslavLitvinov:
Closes #366

* Worksheets fixes
* Refactored http/ui modules: group files by features rather than layers 
* Open api spec fixes for ui endpoints


crates/runtime/src/http/router.rs
```diff
@@ -56,17 +43,10 @@ pub fn create_app(state: AppState) -> Router {
...
-        // .merge_from(TableApiDoc::openapi())
-        .merge_from(SchemasApiDoc::openapi())
-        .merge_from(QueryApiDoc::openapi())
-        .merge_from(DatabasesNavigationApiDoc::openapi())
-        .merge_from(WorksheetsApiDoc::openapi());
-    if let Some(extra_spec) = load_openapi_spec() {
-        ui_spec = ui_spec.merge_from(extra_spec);
-    }
+    let ui_spec = ui_open_api_spec();
+    // if let Some(extra_spec) = load_openapi_spec() {
```
rampage644, line 47: Removing or planning to get back later? If remove, let's remove (and .yaml file as well); if planning to reenable, comment would be helpful

crates/runtime/src/http/ui/databases/errors.rs
```diff
@@ -0,0 +1,82 @@
...
+use icebucket_metastore::error::MetastoreError;
+use snafu::prelude::*;
+
+pub type DatabasesResult<T> = Result<T, DatabasesAPIError>;
+
+#[derive(Debug, Snafu)]
+#[snafu(visibility(pub(crate)))]
+pub enum DatabasesAPIError {
+    #[snafu(display("Create database error: {source}"))]
+    Create { source: MetastoreError },
```
rampage644, line 31: Why `MetastoreError`? Is it base error for all errors found in the project? If so, then just a nit: need better name

crates/runtime/src/http/ui/databases/errors.rs
```diff
@@ -0,0 +1,82 @@
+// Licensed to the Apache Software Foundation (ASF) under one
```
rampage644, line 1: I also wonder if we can consolidate UI errors in a single place vs. error per UI model

crates/runtime/src/http/ui/databases/errors.rs
rampage644: Oh, I think I get this one - UI handler use metastore to get actual models

crates/runtime/src/http/ui/queries/handlers.rs
```diff
@@ -0,0 +1,152 @@
...
+    responses(
+        (status = 200, description = "Returns result of the query", body = QueryResponse),
+        (status = 404, description = "Worksheet not found", body = ErrorResponse),
+        (status = 409, description = "Bad request", body = ErrorResponse),
+        (status = 422, description = "Unprocessable entity", body = ErrorResponse),
+        (status = 500, description = "Internal server error", body = ErrorResponse)
+    )
+)]
+#[tracing::instrument(level = "debug", skip(state), err, ret(level = tracing::Level::TRACE))]
+// Add time sql took
```
rampage644, line 64: nit: add `TODO`

crates/runtime/src/http/ui/queries/models.rs
```diff
@@ -0,0 +1,85 @@
...
+            context: None,
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, Validate, ToSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct QueryResponse {
+    pub id: QueryHistoryId,
+    pub worksheet_id: WorksheetId,
```
rampage644, line 47: Ideally, there should be queries with worksheet associated with it

rampage644 (approved):

crates/runtime/src/http/router.rs
YaroslavLitvinov: > Removing or planning to get back later? If remove, let's remove (and .yaml file as well); if planning to reenable, comment would be helpful

Not planning to get it back later, unless issues will happen. Here I just moved endpoints's specs closer to their place.

crates/runtime/src/http/ui/databases/errors.rs
YaroslavLitvinov: > Why `MetastoreError`? Is it base error for all errors found in the project? If so, then just a nit: need better name

This is underlying error from metastore crate, using metastore api for databases and etc

crates/runtime/src/http/ui/databases/errors.rs
YaroslavLitvinov: > I also wonder if we can consolidate UI errors in a single place vs. error per UI model

The issue arises from our previous implementation, where we reused all underlying errors without creating distinct error groups for each scenario. Introducing errors per API model was a way to add granularity to error handling, allowing us to map errors to specific status codes based on the handler. Since different handlers might return different status codes for the same error, creating dedicated error groups ensures that errors are mapped correctly to the appropriate status code depending on where the error occurred.

Currently  paradigm switched from grouping by layers to grouping by features. So each error group in its own bucket. If I wrong it can fix it later.

crates/runtime/src/http/ui/queries/models.rs
YaroslavLitvinov: > Ideally, there should be queries with worksheet associated with it

QueryResponse::query field serves such purpose

rampage644: My bad, I meant **without** worksheet

DanCodedThis:
Why did this branch fail to deploy?

PR #377, opened 2025-03-26
Update Snowflake response handling

rampage644:
Return 200 OK status codes for semantically valid errors that users can take action on, along with appropriate error messages in the response body.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -61,6 +61,10 @@ datafusion-macros = { git = "https://github.com/Embucket/datafusion.git", rev = 
 # issue #309
 chrono = { git = "https://github.com/chronotope/chrono.git", rev = "8b863490d88ba098038392c8aa930012ffd0c439" }
 
+# https://github.com/juhaku/utoipa/issues/1345
+utoipa = { git = "https://github.com/juhaku/utoipa.git", rev = "1d32f0524a8680d79537ae92aa9ced6ba9d2b630" }
+utoipa-swagger-ui = { git = "https://github.com/juhaku/utoipa.git", rev = "1d32f0524a8680d79537ae92aa9ced6ba9d2b630" }
+
 [workspace.lints.clippy]
 all = { level = "deny", priority = -1 }
 pedantic = { level = "warn", priority = 1 }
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -61,6 +61,10 @@ datafusion-macros = { git = "https://github.com/Embucket/datafusion.git", rev =
 # issue #309
 chrono = { git = "https://github.com/chronotope/chrono.git", rev = "8b863490d88ba098038392c8aa930012ffd0c439" }
 
+# https://github.com/juhaku/utoipa/issues/1345
```
rampage644, line 64: TODO: this should be removed once utoipa version is bumped

crates/metastore/src/models/volumes.rs
```diff
@@ -197,7 +197,11 @@ impl IceBucketVolume {
                     .context(metastore_error::ObjectStoreSnafu)
             }
             IceBucketVolumeType::File(_) => {
-                Ok(Arc::new(object_store::local::LocalFileSystem::new()) as Arc<dyn ObjectStore>)
+                let prefix = self.prefix();
```
rampage644: Unrelated, but fixes file-based volume

crates/runtime/src/execution/error.rs
```diff
@@ -101,13 +101,3 @@ pub enum ExecutionError {
 }
 
 pub type ExecutionResult<T> = std::result::Result<T, ExecutionError>;
-
-impl IntoResponse for ExecutionError {
```
rampage644, line 105: This has been moved closer to where actual response is generated such that it stays in a single place vs. spread in the modules. More work to be done.

crates/metastore/src/models/volumes.rs
```diff
@@ -212,7 +216,7 @@ impl IceBucketVolume {
                 .bucket
                 .as_ref()
                 .map_or_else(|| "s3://".to_string(), |bucket| format!("s3://{bucket}")),
-            IceBucketVolumeType::File(volume) => format!("file://{}", volume.path),
+            IceBucketVolumeType::File(volume) => volume.path.clone(),
```
osipovartem: This change break FS volume create/update table

rampage644: Thanks for noting, will check and redo

osipovartem (approved):

PR #387, opened 2025-03-27
Build and deploy only for labeled PRs

osipovartem:
To build and deploy docker image we need to set **deploy** label

Or we can build on each merge, but deploy only labeled PRs
This will reduce amount of deploys for non-urgent PRs


rampage644 (approved):
We need more discussion on this, I am not sure this is the best approach to take, approving this one as a temporary solution.

PR #402, opened 2025-04-03
UI Restore Schemas Get List

### LINKED ISSUE #399: api: UI Restore Schemas Get List
- `/ui/databases/{databaseName}/schemas` GET LIST
- Add pagination
- Investigate created_at, updated_at for schemas

DanCodedThis:
Closes #399 

- [x] `/ui/databases/{databaseName}/schemas` GET LIST
- [x] Add pagination
- [x] Investigate created_at, updated_at for schemas

### DIFF
--- a/crates/runtime/src/http/ui/databases/handlers.rs
+++ b/crates/runtime/src/http/ui/databases/handlers.rs
@@ -21,8 +21,8 @@ use crate::http::{
     metastore::handlers::QueryParameters,
     ui::databases::error::{DatabasesAPIError, DatabasesResult},
     ui::databases::models::{
-        DatabaseCreatePayload, DatabaseCreateResponse, DatabaseResponse, DatabaseUpdatePayload,
-        DatabaseUpdateResponse, DatabasesResponse,
+        Database, DatabaseCreatePayload, DatabaseCreateResponse, DatabaseResponse,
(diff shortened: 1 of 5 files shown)


YaroslavLitvinov (approved):
skip() function in pagination is just a logical abstraction, as regardless provided cursor it is doing a full scan anyway. Since metastore won't use slatedb iterator, it loads a full list of available keys instead. It's better to switch to a slatedb iterator along with fixing known metastore keys issue.

DanCodedThis:
> skip() function in pagination is just a logical abstraction, as regardless provided cursor it is doing a full scan anyway. Since metastore won't use slatedb iterator, it loads a full list of available keys instead. It's better to switch to a slatedb iterator along with fixing known metastore keys issue.

**_SlateDB_**'s iterator is simplistic, it doesn't have any of the methods we have in the **_magical rust iterators_**.

It only has `next` and `seek`.

Maybe it will have more methods in the future, however last time I checked, to make a rust iterator you only had to provide the `next` method. So it's probably the underline mechanism that is blocking them from implementing the full rust iterator on their's.

YaroslavLitvinov:
This is pure practical question, nobody cares about how many features their iterator implements and those are implemented are pure rustish or not, It allows iterating over a specified range of keys, instead of doing a full scan.

PR #403, opened 2025-04-04
UI Get List Pagination + refactor full scan in slatedb

### LINKED ISSUE #401: api: UI all `GET LIST` requests to have pagination
- Investigate pagination for `NavigationTrees`
- Investigate pagination for `Schemas`
- Investigate pagination for `DB's`

Create a sub-issue for each if needed

DanCodedThis:
Closes #401 +

- In utils crate now we have the ability to use a cursor and limit.
- next_cursor is name of the last object + `\x00` - Nul ASCII char.
- Example: `tbl.{...}.aa < tbl.{...}.aa\x00 < tbl.{...}.ab`

### DIFF
--- a/crates/utils/src/lib.rs
+++ b/crates/utils/src/lib.rs
@@ -30,6 +30,7 @@ use snafu::prelude::*;
 use std::ops::RangeBounds;
 use std::string::ToString;
 use std::sync::Arc;
+use futures::stream_select;
 use uuid::Uuid;
 
 #[derive(Snafu, Debug)]
@@ -149,15 +150,21 @@ impl Db {
     pub async fn list_objects<T: Send + for<'de> serde::de::Deserialize<'de>>(
         &self,
         key: &str,
+        cursor: Option<String>,
+        limit: Option<usize>,
     ) -> Result<Vec<T>> {
-        let start = format!("{key}/");
+        let start = cursor.map_or(format!("{key}/"), |cursor| format!("{key}/{cursor}\x00"));
         let end = format!("{key}/\x7F");
         let range = Bytes::from(start)..Bytes::from(end);
+        let limit = limit.unwrap_or(usize::MAX);
         let mut iter = self.0.scan(range).await.context(ScanFailedSnafu)?;
         let mut objects: Vec<T> = vec![];
         while let Ok(Some(value)) = iter.next().await {
             let value = de::from_slice(&value.value).context(DeserializeValueSnafu)?;
             objects.push(value);
+            if objects.len() >= limit {
+                break;
+            }
         }
         Ok(objects)
     }
@@ -257,7 +264,7 @@ pub trait Repository {
     }
 
     async fn _list(&self) -> Result<Vec<Self::Entity>> {
-        let entities = self.db().list_objects(Self::collection_key()).await?;
+        let entities = self.db().list_objects(Self::collection_key(), None, None).await?;
         Ok(entities)
     }
 
@@ -295,12 +302,12 @@ mod test {
             .await
             .expect("Failed to put entity");
         let get_after_put = db.get::<TestEntity>("test/abc").await;
-        let list_after_append = db.list_objects::<TestEntity>("test").await;
+        let list_after_append = db.list_objects::<TestEntity>("test", None, None).await;
         db.delete("test/abc")
             .await
             .expect("Failed to delete entity");
         let get_after_delete = db.get::<TestEntity>("test/abc").await;
-        let list_after_remove = db.list_objects::<TestEntity>("test").await;
+        let list_after_remove = db.list_objects::<TestEntity>("test", None, None).await;
 
         insta::assert_debug_snapshot!((
             get_empty,
--- a/crates/utils/src/lib.rs
+++ b/crates/utils/src/lib.rs
@@ -30,7 +30,6 @@ use snafu::prelude::*;
 use std::ops::RangeBounds;
 use std::string::ToString;
 use std::sync::Arc;
-use futures::stream_select;
 use uuid::Uuid;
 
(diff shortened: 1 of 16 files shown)


DanCodedThis:
Dependency has a broken test

eadgbear:
Not to get too pedantic, but ideally `list_volumes`, or any of the `list_` functions should return an iterator that can be configured to skip and paginate :) Something to do in the future.

DanCodedThis:
> Not to get too pedantic, but ideally `list_volumes`, or any of the `list_` functions should return an iterator that can be configured to skip and paginate :) Something to do in the future.

Makes sense, then we could remove the limit and cursor from their signatures

crates/utils/src/lib.rs
```diff
@@ -149,15 +149,22 @@ impl Db {
     pub async fn list_objects<T: Send + for<'de> serde::de::Deserialize<'de>>(
         &self,
         key: &str,
+        cursor: Option<String>,
+        limit: Option<usize>,
     ) -> Result<Vec<T>> {
-        let start = format!("{key}/");
+        let start =
+            cursor.map_or_else(|| format!("{key}/"), |cursor| format!("{key}/{cursor}\x00"));
```
YaroslavLitvinov, line 156: Why start_key needs \x00  ?

DanCodedThis:
> > Not to get too pedantic, but ideally `list_volumes`, or any of the `list_` functions should return an iterator that can be configured to skip and paginate :) Something to do in the future.
> 
> Makes sense, then we could remove the limit and cursor from their signatures

However, if we do this, we would get back to full scan. We have to maybe extend the **_SlateDBIterator_** or at least expose it higher in the API

YaroslavLitvinov:
> > > Not to get too pedantic, but ideally `list_volumes`, or any of the `list_` functions should return an iterator that can be configured to skip and paginate :) Something to do in the future.
> > 
> > 
> > Makes sense, then we could remove the limit and cursor from their signatures
> 
> However, if we do this, we would get back to full scan. We have to maybe extend the **_SlateDBIterator_** or at least expose it higher in the API

Range Iterator iterating over a range you provided, since you are

> > > Not to get too pedantic, but ideally `list_volumes`, or any of the `list_` functions should return an iterator that can be configured to skip and paginate :) Something to do in the future.
> > 
> > 
> > Makes sense, then we could remove the limit and cursor from their signatures
> 
> However, if we do this, we would get back to full scan. We have to maybe extend the **_SlateDBIterator_** or at least expose it higher in the API

I think it's mainly depends on how you would implement skip() - at least no full scan if used DbIterator::seek(), according to API reference: https://docs.rs/slatedb/latest/slatedb/struct.DbIterator.html.

As for me it would a better alternative and a bit more rust-ish when returning iterator exposing skip() and limit functions versus adding None, None across the code.

Either exposing slatedb iterator, if it's considered usable enough.

crates/utils/src/lib.rs
```diff
@@ -295,12 +305,12 @@ mod test {
...
             .expect("Failed to put entity");
         let get_after_put = db.get::<TestEntity>("test/abc").await;
-        let list_after_append = db.list_objects::<TestEntity>("test").await;
+        let list_after_append = db.list_objects::<TestEntity>("test", None, None).await;
         db.delete("test/abc")
             .await
             .expect("Failed to delete entity");
         let get_after_delete = db.get::<TestEntity>("test/abc").await;
-        let list_after_remove = db.list_objects::<TestEntity>("test").await;
+        let list_after_remove = db.list_objects::<TestEntity>("test", None, None).await;
```
YaroslavLitvinov, line 313: Makes sense to test pagination works here in lower level.

YaroslavLitvinov (approved):

YaroslavLitvinov:
> Dependency has a broken test

https://github.com/juhaku/utoipa/pull/1363/files

crates/utils/src/lib.rs
DanCodedThis: Because it's will be the smallest next key after the last one. We take the last key and append a NUL char, this means we exclude the last key itself and open up any other key after the last one to be consumed.

crates/utils/src/lib.rs
DanCodedThis: We'll add more tests later, when we move to a different key system. Since this is essentially just a patch for now.

DanCodedThis:
> > > > Not to get too pedantic, but ideally `list_volumes`, or any of the `list_` functions should return an iterator that can be configured to skip and paginate :) Something to do in the future.
> > > 
> > > 
> > > Makes sense, then we could remove the limit and cursor from their signatures
> > 
> > 
> > However, if we do this, we would get back to full scan. We have to maybe extend the **_SlateDBIterator_** or at least expose it higher in the API
> 
> Range Iterator iterating over a range you provided, since you are
> 
> > > > Not to get too pedantic, but ideally `list_volumes`, or any of the `list_` functions should return an iterator that can be configured to skip and paginate :) Something to do in the future.
> > > 
> > > 
> > > Makes sense, then we could remove the limit and cursor from their signatures
> > 
> > 
> > However, if we do this, we would get back to full scan. We have to maybe extend the **_SlateDBIterator_** or at least expose it higher in the API
> 
> I think it's mainly depends on how you would implement skip() - at least no full scan if used DbIterator::seek(), according to API reference: https://docs.rs/slatedb/latest/slatedb/struct.DbIterator.html.
> 
> As for me it would a better alternative and a bit more rust-ish when returning iterator exposing skip() and limit functions versus adding None, None across the code.
> 
> Either exposing slatedb iterator, if it's considered usable enough.

Definitely need to expose it. It's either we get the point across though the signature or through using the iterator directly. Either way, we have to provide the cursor and the limit to the underline impl.

PR #404, opened 2025-04-06
WIP Uploading data via in memory table

rampage644:
This PR partially restores CTAS capabilities and updates upload method for execution service. It currently breaks table creation with TIMESTAMP columns.

I added an failure expected test to fix that in the next PR.

### DIFF
--- a/crates/runtime/src/execution/catalog/mod.rs
+++ b/crates/runtime/src/execution/catalog/mod.rs
@@ -69,6 +69,7 @@ pub struct IceBucketDFMetastore {
     pub metastore: Arc<dyn Metastore>,
     pub mirror: Arc<CatalogProviderCache>,
     pub table_object_store: Arc<DashMap<String, Arc<dyn ObjectStore>>>,
+    pub catalogs: DashMap<String, Arc<dyn CatalogProvider>>,
 }
 
 impl IceBucketDFMetastore {
@@ -79,6 +80,7 @@ impl IceBucketDFMetastore {
             metastore,
             mirror: Arc::new(DashMap::new()),
             table_object_store: Arc::new(table_object_store),
+            catalogs: DashMap::default(),
         }
     }
 
@@ -260,18 +262,26 @@ impl CatalogProviderList for IceBucketDFMetastore {
 
     fn register_catalog(
         &self,
-        _name: String,
-        _catalog: Arc<dyn CatalogProvider>,
+        name: String,
+        catalog: Arc<dyn CatalogProvider>,
     ) -> Option<Arc<dyn CatalogProvider>> {
-        // This is currently a NOOP because we don't support registering new catalogs yet
-        None
+        self.catalogs.insert(name, catalog)
     }
 
     fn catalog_names(&self) -> Vec<String> {
-        self.mirror.iter().map(|e| e.key().clone()).collect()
+        let mut catalog_names = self
+            .catalogs
+            .iter()
+            .map(|c| c.key().clone())
+            .collect::<Vec<String>>();
+        catalog_names.extend(self.mirror.iter().map(|e| e.key().clone()));
+        catalog_names
     }
 
     fn catalog(&self, name: &str) -> Option<Arc<dyn CatalogProvider>> {
+        if let Some(catalog) = self.catalogs.get(name) {
+            return Some(catalog.value().clone());
+        }
         if !self.mirror.contains_key(name) {
             return None;
         }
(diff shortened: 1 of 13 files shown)


crates/runtime/src/execution/catalog/mod.rs
```diff
@@ -260,18 +262,26 @@ impl CatalogProviderList for IceBucketDFMetastore {
...
     fn register_catalog(
         &self,
-        _name: String,
-        _catalog: Arc<dyn CatalogProvider>,
+        name: String,
+        catalog: Arc<dyn CatalogProvider>,
     ) -> Option<Arc<dyn CatalogProvider>> {
-        // This is currently a NOOP because we don't support registering new catalogs yet
-        None
+        self.catalogs.insert(name, catalog)
```
osipovartem, line 268: why not use 
```rust
pub mirror: Arc<CatalogProviderCache>,
```

crates/runtime/src/execution/service.rs
```diff
@@ -134,77 +137,73 @@ impl ExecutionService {
...
-        // We construct this URL so we can unwrap it
-        #[allow(clippy::unwrap_used)]
-        user_session.ctx.register_object_store(
-            ObjectStoreUrl::parse(&upload_path).unwrap().as_ref(),
-            object_store.clone(),
-        );
+        let src_table_ref = "tmp_db.tmp_schema.tmp_table";
+        let inmem_catalog = MemoryCatalogProvider::new();
+        inmem_catalog
+            .register_schema("tmp_schema", Arc::new(MemorySchemaProvider::new()))
```
osipovartem: Just use inmemory object store to load csv data and read add with already prepared schema schema

```rust
    let store = InMemory::new();
            store.put(&Path::from(file_name), data.into()).await.context(ex_error::ObjectStoreSnafu)?;
            let url = Url::parse("memory://").unwrap();
            user_session.ctx.register_object_store(&url, Arc::new(store));

        // Here we can get schema of csv file
        let df = user_session
            .ctx
            .read_csv("memory:///", CsvReadOptions::new())
            .await
            .context(ex_error::DataFusionSnafu)?
            .collect()
            .await
            .context(ex_error::DataFusionSnafu)?;

        let table =
            MemTable::try_new(df[0].schema(), vec![df]).expect("Failed to create table");
                .collect::<Vec<_>>();
                        let src_table_ref = "tmp_db.tmp_schema.tmp_table";
        let inmem_catalog = MemoryCatalogProvider::new();
        inmem_catalog
            .register_schema("tmp_schema", Arc::new(MemorySchemaProvider::new()))
            .context(ex_error::DataFusionSnafu)?;
        user_session
            .ctx
            .register_catalog("tmp_db", Arc::new(inmem_catalog));

        user_session
            .ctx
            .register_table(
                TableReference::full("tmp_db", "tmp_schema", "tmp_table"),
                Arc::new(table),
            ).context(ex_error::DataFusionSnafu)?;
```
            
also as far I remember we don't support create as select statements (we have to create a table manually)

YaroslavLitvinov: Using datafusion's read_csv we can infer some schema, and using LogicalPlanBuilder::insert_into it creates a table from the schema. So theoretically create table before upload is not a hard prerequisite.

crates/runtime/src/execution/catalog/mod.rs
rampage644: This TBD, I consider `mirror` to be a cache and something to be treated very differently. We still lack temp tables design (and thus catalogs and schemas).

crates/runtime/src/execution/service.rs
rampage644: Finally updated the code to read the csv and infer schema automatically.

camuel (approved):
blind approve

osipovartem (approved):

crates/runtime/src/execution/service.rs
```diff
@@ -125,86 +129,90 @@ impl ExecutionService {
...
+            source_table.catalog().unwrap_or_default(),
+            Arc::new(inmem_catalog),
         );
+
+        // Here we create an arrow csv reader that infers the schema from first 10 rows
+        // and then builds a record batch
+        // TODO: This partially duplicates what Datafusion does with `CsvFormat::infer_schema`
+        let (schema, _) = Format::default()
+            .with_header(true)
+            .infer_schema(data.clone().reader(), Some(100))
```
YaroslavLitvinov, line 166: Cloning all buffer for 10 rows, in case of csv can be replaced by reading just required amount of lines, to get slice of data. Not sure about Parquet.

YaroslavLitvinov (approved):

rampage644:
Rebased on recent main

PR #416, opened 2025-04-10
Add s3 tables catalog (initial impl)

osipovartem:


crates/runtime/src/execution/session.rs
```diff
@@ -91,6 +99,65 @@ impl IceBucketUserSession {
...
+
+            let catalog = DataFusionIcebergCatalog::new(Arc::new(catalog), None)
+                .await
+                .context(ex_error::DataFusionSnafu)?;
+            let catalog_provider = Arc::new(catalog) as Arc<dyn CatalogProvider>;
+
+            let _catalog_name = volume.bucket.clone().unwrap_or_default().to_string();
+
+            // TODO set correct catalog name (s3 tables bucket contains "_" which is not supported by datafusion)
+            ctx.register_catalog("s3_tables", catalog_provider);
```
rampage644, line 156: I would suggest expanding volume (perhaps creating different type to differentiate from regular bucket), adding name to it explictly to be used as a catalog name.

osipovartem: It was in initial version, but since it was just for tests I tried to avoid additional structs, will fix

rampage644: Makes sense

crates/metastore/src/models/volumes.rs
```diff
@@ -115,14 +117,15 @@ pub struct IceBucketFileVolume {
...
+    S3Tables(IceBucketS3Volume),
     File(IceBucketFileVolume),
     Memory,
 }
 
 impl Validate for IceBucketVolumeType {
     fn validate(&self) -> Result<(), ValidationErrors> {
         match self {
-            Self::S3(volume) => volume.validate(),
+            Self::S3(volume) | Self::S3Tables(volume) => volume.validate(),
```
YaroslavLitvinov, line 128: consider using
`Self::S3(volume) | Self::S3Tables(volume) | Self::File(volume) => volume.validate(),`

YaroslavLitvinov (approved):

crates/metastore/src/models/volumes.rs
osipovartem: Self::S3(volume) | Self::S3Tables(volume) | Self::File(volume) => volume.validate(),
    |                      ------                   ^^^^^^ expected `&IceBucketS3Volume`, found `&IceBucketS3TablesVolume`
    |                      |
    |                      first introduced with type `&IceBucketS3Volume` here

PR #419, opened 2025-04-10
UI Tables Get List

### LINKED ISSUE #405: api: UI Tables Get List (with pagination)
- `/database/${databaseName}/schemas/${schemaName}/tables` Get List
- Snowflake response fields: **_NAME_**, _**TYPE**_, _**OWNER**_, **_ROWS_**, **_BYTES_**, **_CREATED / UPDATED_**

DanCodedThis:
Closes #405 

- [x] `/database/${databaseName}/schemas/${schemaName}/tables` Get List with Snowflake response fields: **_NAME, TYPE, OWNER, ROWS, BYTES, CREATED / UPDATED_**
- [x] Pagination with cursor
- [x] Tests

### DIFF
--- a/crates/runtime/src/http/ui/tables/handlers.rs
+++ b/crates/runtime/src/http/ui/tables/handlers.rs
@@ -21,8 +21,9 @@ use crate::http::session::DFSessionId;
 use crate::http::state::AppState;
 use crate::http::ui::tables::error::{TablesAPIError, TablesResult};
 use crate::http::ui::tables::models::{
-    TableColumnInfo, TableColumnsInfoResponse, TablePreviewDataColumn, TablePreviewDataParameters,
-    TablePreviewDataResponse, TablePreviewDataRow, TableStatistics, TableStatisticsResponse,
+    Table, TableColumnInfo, TableColumnsInfoResponse, TablePreviewDataColumn,
+    TablePreviewDataParameters, TablePreviewDataResponse, TablePreviewDataRow, TableStatistics,
+    TableStatisticsResponse, TablesParameters, TablesResponse,
 };
 use arrow_array::{Array, StringArray};
 use axum::extract::Query;
@@ -31,7 +32,7 @@ use axum::{
     Json,
 };
 use icebucket_metastore::error::MetastoreError;
-use icebucket_metastore::IceBucketTableIdent;
+use icebucket_metastore::{IceBucketSchemaIdent, IceBucketTableIdent};
 use utoipa::OpenApi;
 
 #[derive(OpenApi)]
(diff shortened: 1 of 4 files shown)


crates/runtime/src/http/ui/tables/handlers.rs
```diff
@@ -264,3 +265,76 @@ pub async fn get_table_preview_data(
...
     }))
 }
+
+#[utoipa::path(
+    get,
+    path = "/ui/databases/{databaseName}/schemas/{schemaName}/tables",
+    params(
+        ("databaseName" = String, description = "Database Name"),
+        ("schemaName" = String, description = "Schema Name"),
+        ("tableName" = String, description = "Table Name")
```
YaroslavLitvinov: table_name not used here

crates/runtime/src/http/ui/tables/handlers.rs
```diff
@@ -264,3 +265,76 @@ pub async fn get_table_preview_data(
...
+                .iter()
+                .last()
+                .map_or(String::new(), |rw_table| rw_table.ident.table.clone());
+
+            let items = rw_tables
+                .into_iter()
+                .map(|rw_table| {
+                    let mut total_bytes = 0;
+                    let mut total_rows = 0;
+                    if let Ok(Some(latest_snapshot)) = rw_table.metadata.current_snapshot(None) {
```
YaroslavLitvinov, line 308: why not use if let { (total_bytes, total_rows) } else { (0, 0) }
instead of let mut ?

YaroslavLitvinov (changes requested):

crates/runtime/src/http/ui/tables/handlers.rs
DanCodedThis: I think this is a little bit harder to read (since it's more common to see the **_if-not_** (or however it's called) pattern): 

```Rust
let (total_bytes, total_rows) = if let Ok(Some(latest_snapshot)) = rw_table.metadata.current_snapshot(None) {
                        (latest_snapshot
                            .summary()
                            .other
                            .get("total-files-size")
                            .and_then(|value| value.parse::<i64>().ok())
                            .unwrap_or(0),
                        latest_snapshot
                            .summary()
                            .other
                            .get("total-records")
                            .and_then(|value| value.parse::<i64>().ok())
                            .unwrap_or(0))
                    } else {
                        (0, 0)
                    };
```

Especially since we need to have an `unwrap_or(0)` there anyway. I think the top level if is not really connected to the else statement, so it's more logical with `mut`.

P.S. This logic was taken almost as is from Artem's statistics endpoint before **_Metastore_**.

Let me know what'd you think?

YaroslavLitvinov: This can be done without mut, so using mut here is excessive.

YaroslavLitvinov (approved):

PR #422, opened 2025-04-11
UI Search by name for Get List handlers

### LINKED ISSUE #421: api: UI Search by name for Get List handlers
- [x] Preliminary impl for search in **_crates/utils_**
- [x] Refactor to use `ListConfig` wrapper instead of `None, None, None` in `fn` signatures
- [x] Update **_Metastore_** impl accordingly
- [x] Update all Get List handlers in **_UI_** (excluding: **_Worksheets_** and **_Queries_**)
- [x] Update all other Get List handlers to accommodate the change (ex: **_Catalog, Metastore,_** etc)
- [x] Update and Add new tests

DanCodedThis:
Closes #421 

- Impl for search in **_crates/utils_** + refactor the signature of `fn list_objects` and others to use `ListConfig` wrapper instead of `None, None, None`
- Updated **_UI_** handlers for `Databases`, `Schemas` and `Tables` + Added tests
- Updated **_Metastore_** impl + fixed handlers for it, **_Catalog_** handlers, etc to accommodate the change

### DIFF
--- a/crates/metastore/src/metastore.rs
+++ b/crates/metastore/src/metastore.rs
@@ -18,6 +18,8 @@
 use std::{collections::HashMap, sync::Arc};
 
 use crate::error::{self as metastore_error, MetastoreResult};
+#[allow(clippy::wildcard_imports)]
+use crate::models::*;
 use async_trait::async_trait;
 use bytes::Bytes;
 use chrono::Utc;
@@ -25,14 +27,12 @@ use dashmap::DashMap;
 use futures::{StreamExt, TryStreamExt};
 use iceberg_rust::catalog::commit::apply_table_updates;
 use iceberg_rust_spec::table_metadata::{FormatVersion, TableMetadataBuilder};
+use icebucket_utils::list_config::ListConfig;
 use icebucket_utils::Db;
 use object_store::{path::Path, ObjectStore, PutPayload};
 use serde::de::DeserializeOwned;
 use snafu::ResultExt;
 use uuid::Uuid;
-use icebucket_utils::list_config::ListConfig;
-#[allow(clippy::wildcard_imports)]
-use crate::models::*;
 
 #[async_trait]
 pub trait Metastore: std::fmt::Debug + Send + Sync {
--- a/crates/metastore/src/metastore.rs
+++ b/crates/metastore/src/metastore.rs
@@ -30,7 +30,7 @@ use object_store::{path::Path, ObjectStore, PutPayload};
 use serde::de::DeserializeOwned;
 use snafu::ResultExt;
 use uuid::Uuid;
-
+use icebucket_utils::list_config::ListConfig;
 #[allow(clippy::wildcard_imports)]
 use crate::models::*;
 
@@ -38,8 +38,7 @@ use crate::models::*;
 pub trait Metastore: std::fmt::Debug + Send + Sync {
     async fn list_volumes(
         &self,
-        cursor: Option<String>,
-        limit: Option<usize>,
+        list_config: ListConfig,
     ) -> MetastoreResult<Vec<RwObject<IceBucketVolume>>>;
     async fn create_volume(
         &self,
@@ -67,8 +66,7 @@ pub trait Metastore: std::fmt::Debug + Send + Sync {
 
     async fn list_databases(
         &self,
-        cursor: Option<String>,
-        limit: Option<usize>,
+        list_config: ListConfig,
     ) -> MetastoreResult<Vec<RwObject<IceBucketDatabase>>>;
     async fn create_database(
         &self,
@@ -93,8 +91,7 @@ pub trait Metastore: std::fmt::Debug + Send + Sync {
     async fn list_schemas(
         &self,
         database: &IceBucketDatabaseIdent,
-        cursor: Option<String>,
-        limit: Option<usize>,
+        list_config: ListConfig,
     ) -> MetastoreResult<Vec<RwObject<IceBucketSchema>>>;
     async fn create_schema(
         &self,
@@ -119,8 +116,7 @@ pub trait Metastore: std::fmt::Debug + Send + Sync {
     async fn list_tables(
         &self,
         schema: &IceBucketSchemaIdent,
-        cursor: Option<String>,
-        limit: Option<usize>,
+        list_config: ListConfig,
     ) -> MetastoreResult<Vec<RwObject<IceBucketTable>>>;
     async fn create_table(
         &self,
@@ -200,15 +196,14 @@ impl SlateDBMetastore {
     async fn list_objects<T>(
         &self,
         list_key: &str,
-        cursor: Option<String>,
-        limit: Option<usize>,
+        list_config: ListConfig,
     ) -> MetastoreResult<Vec<RwObject<T>>>
     where
         T: serde::Serialize + DeserializeOwned + Eq + PartialEq + Send + Sync,
     {
         let entities = self
             .db
-            .list_objects(list_key, cursor, limit)
+            .list_objects(list_key, list_config)
             .await
             .context(metastore_error::UtilSlateDBSnafu)?;
         Ok(entities)
@@ -294,10 +289,9 @@ impl SlateDBMetastore {
 impl Metastore for SlateDBMetastore {
     async fn list_volumes(
         &self,
-        cursor: Option<String>,
-        limit: Option<usize>,
+        list_config: ListConfig,
     ) -> MetastoreResult<Vec<RwObject<IceBucketVolume>>> {
-        self.list_objects(KEY_VOLUME, cursor, limit).await
+        self.list_objects(KEY_VOLUME, list_config).await
     }
 
     async fn create_volume(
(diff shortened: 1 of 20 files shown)


crates/metastore/src/metastore.rs
```diff
@@ -18,28 +18,27 @@
...
-#[allow(clippy::wildcard_imports)]
-use crate::models::*;
-
 #[async_trait]
 pub trait Metastore: std::fmt::Debug + Send + Sync {
     async fn list_volumes(
         &self,
-        cursor: Option<String>,
-        limit: Option<usize>,
+        list_config: ListConfig,
```
rampage644, line 41: I like this approach better than previous one. I was also thinking if listing options/paramters should be part of the interface as well, but have no strong opinion at the moment

crates/runtime/src/http/ui/tests/databases.rs
```diff
@@ -277,4 +278,95 @@ async fn test_ui_databases() {
         "test3".to_string(),
         databases_response.items.first().unwrap().name
     );
+
+    // Create database with another name, Ok
+    let expected_another = DatabaseCreatePayload {
```
rampage644, line 283: Not a feedback for this specific PR, and more of a generic testing approach: I wonder if `insta` snapshot based testing approach is a better fit here. 

Are there interesting approaches to http enpoints testing in axum/actix/other rust web frameworks?

crates/runtime/src/http/ui/tests/tables.rs
```diff
@@ -382,4 +382,112 @@ async fn test_ui_tables() {
     let tables: TablesResponse = res.json().await.unwrap();
     assert_eq!(2, tables.items.len());
     assert_eq!("tested3".to_string(), tables.items.first().unwrap().name);
+
+    //Create a table with another name
+    let query_payload = QueryCreatePayload {
+        query: format!(
+            "create or replace Iceberg TABLE {}.{}.{}
```
rampage644: Use `CREATE TABLE` instead of Iceberg table - there are currently subtle bugs with iceberg statements.

crates/runtime/src/http/ui/tests/databases.rs
DanCodedThis: I'll look into it

crates/utils/src/list_config.rs
```diff
@@ -0,0 +1,39 @@
...
+//
+// Unless required by applicable law or agreed to in writing,
+// software distributed under the License is distributed on an
+// "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied.  See the License for the
+// specific language governing permissions and limitations
+// under the License.
+
+pub struct ListConfig {
+    pub cursor: Option<String>,
```
rampage644, line 26: I would suggest better naming (and/or comments) here. It is not quite clear what is the cursor and search strings. 

Perhaps `cursor` -> `starts_with`. Can we make limit non optional but having sane default?

DanCodedThis: - `search` -> `starts_with`
- `cursor` probably needs to stay, can't think of a better name (`from_where`?)
- limit `Option<usize>` -> `usize::Max` (we need to return all of the items if there is no limit, especially for non UI handlers)

What do you think? I would keep the `Option<usize>` for simplicity inside UI handlers (since their Query params are always options)

rampage644: Typically you want either search/starts_with (indicating starting iteration at certain key) or cursor/token (indicating get next page/batch/chunk of items).

rampage644: I.e. here is how `lakekeeper` [implements it](https://github.com/lakekeeper/lakekeeper/blob/main/crates/iceberg-catalog/src/implementations/postgres/namespace.rs)

crates/metastore/src/metastore.rs
rampage644: The reason I am not a fan of having `ListConfig` here is that it forces one to use transport-level API pagination mechanism in Rust. 

Probably, it's more elegant/interesting/idiomatic to rely in streaming/iterator/whatever in internal traits and terminate token/cursor/pagination on http/ui level.

DanCodedThis: Yeah, I agree completely, it seems like UI is a dripping roof, and the water gets deeper than it should. We can implement our own thing on top of the **_SlateDB iter_**, since their iterator only has two methods, however I saw that they are planing to add filter to their iterator in the near future.

DanCodedThis: This probably should be a distinct PR (for iter)

crates/utils/src/list_config.rs
DanCodedThis: I'll look into it

crates/metastore/src/metastore.rs
DanCodedThis: Our own iterator should or trait or whatever it will be, should just reimagine the utils crate from just a wrapper to a full experience, make it more ergonomic, unify the whole thing to a simple-to-use interface with everything that you expect from an iterator. (is unify the right word?)

DanCodedThis: Actually scratch that, I think, maybe (not sure if it would work well) I have another a lot simpler idea

DanCodedThis: ## The problem
 All the magic happens inside the scan function, not the **_SlateDB_** iterator
 
## Possible solution
Something like a builder pattern: `state.metastore.list().volumes().get()` - and you can do things like add a `cursor`, `limit` or `search` by function and then with `get` you build the iterator and return the `Vec<T>` you need

### For normal uses:
- `state.metastore.list().volumes().get()`
- `state.metastore.list().databases().get()`
- `state.metastore.list().schemas(ident).get()` ( **_ident_** like in the impl right now)
- `state.metastore.list().tables(ident).get()` ( **_ident_** like in the impl right now)

### More sophisticated example 
`state.metastore.list().tables(ident).cursor(cursor).search(token).limit(10).get()` 

## Methods explanation
We can just leverage the rusts typesystem to not be able to call `fn cursor`, `fn get`, `fn limit` or `fn search` if no key
is chosen via `fn volumes`, `fn databases`, `fn schemas` or `fn tables` so these function can return like a `KeyfullListBuilder` with like a concrete type (we also can have `type DatabaseListBuidler = ListBuidler<IceBucketDatabase>`  maybe)
 
```Rust
KeylessListBuilder {
    key: Option<String>
}
KeyfullListBuilder {
key: String,
cursor: Option<String>,
limit: Option<usize>,
token: Option<String>
}

//In Metastore
fn list(&self) -> KeylessListBuilder {
    return KeylessListBuiler::new()
}
 
 
impl KeylessListBuilder {
    fn volumes(&self) -> KeyfullListBuilder //takes a key for volumes
    fn databases(&self) -> KeyfullListBuidler //takes key for databases
    fn schemas(&self, ident: IceBucketSchemaIdent) -> KeyfullListBuilder //takes key for schemas and formats it with the ident
    fn tables(&self, ident: IceBucketTableIdent) -> KeyfullListBuilder //takes key for tables and formats it with the ident
}
 
impl KeyfullListBuilder { 
//maybe use an option in all `fn` paramas as all `Query` parameters are options
    fn cursor(&self, cursor: String) -> KeyfullListBuilder //adds a cursor
    fn limit(&self, limit: usize) -> KeyfullListBuilder //add a limit
    fn search(&self, token: String) -> KeyfullListBuilder //add search token
    fn get(&self) -> Vec<T> //similar to `fn list_objects` in utils crate lib.rs file 
}
```
 
 ## Alternatively
 We can just return from `list_volumes`, `list_databases`, etc - `ListBuilder::new(key)` with a chosen key like (more simple approach, however if more entities will be added will need to add them inside **_Metastore_**, instead of the **_list_builder.rs_** file). The first approach should be more idiomatic.
 
 ```Rust
ListBuilder {
    key: String,
    cursor: Option<String>,
    limit: Option<usize>,
    token: Option<String>
}

impl ListBuidler { ... } //all methods here, are the same, only new is different (added to siganture `key: String`)
 ```

DanCodedThis: Made an issue for this: https://github.com/Embucket/icebucket/issues/423

crates/utils/src/list_config.rs
DanCodedThis: Changed `search` -> `token`

DanCodedThis:
For clarity sake: I reverted the Scan iterator changes and made a separate branch for it (not done yet)

crates/runtime/src/http/ui/tests/databases.rs
YaroslavLitvinov: @rampage644 I think in this particular case when we want to test database creation itself it's ok to use payload and responses structs. but when it comes to testing in overall it looks like we better don't use  a custom code for creating databases, schemas. We could achieve it by using some convention that some standard volumes and databases already exists.

YaroslavLitvinov: using sql

YaroslavLitvinov (approved):

PR #429, opened 2025-04-14
S3 tables integration

osipovartem:
- updated S3 tables logic
- enabled default in-memory catalog **icebucket** with default **public** schema since it should be used for memory tables and other system defaults
- added new volume type for s3 tables where ARN and catalog name are required
- what is work for now
 - ```create table s3_tables.datasets.example_table (vendor_id INT, trip_id INT, store_and_fwd_flag int) ;```
 - ```insert into s3_tables.datasets.example_table values (1,2,3)```
 - ```create table s3_tables.datasets.example_table_2 as select * from s3_tables.datasets.example_table;```
 - local table ```create table database.public.example as select * from s3_tables.datasets.example_table;```
 - ```drop table s3_tables.datasets.example_table_2``` - works with Iceberg extension planner

NOTES
- SchemaProvider **register_table** is not **async** so we have to create local thread / block_on to create a table. During table creation we call object store to put metadata files, etc. **There is additional block_on call and we make a deadlock for the running thread, by this reason we cannot use SchemaProvider table register as is for now**
- To support other type of catalogs we need to handle all such calls by our code since part of DF traits is not flexible for our use cases since we use iceberg format which expects catalog calls
- For s3 tables we use iceberg-rest implementation which is based on IcebergCatalog trait


DanCodedThis (approved):
LGTM

crates/metastore/src/models/volumes.rs
```diff
@@ -81,6 +81,14 @@ pub struct IceBucketS3Volume {
     pub metadata_endpoint: Option<String>,
     #[validate(required, nested)]
     pub credentials: Option<AwsCredentials>,
+}
+
+#[derive(Validate, Serialize, Deserialize, Debug, Clone, PartialEq, Eq, utoipa::ToSchema)]
+#[serde(rename_all = "kebab-case")]
+pub struct IceBucketS3TablesVolume {
+    pub volume: IceBucketS3Volume,
```
rampage644: This feels recursive, I would include all required fields from regular s3 volume directly here even if it causes some code duplication.

Also, I believe it's better to avoid having structs with all `Option<>` fields (is it even valid object to have?).  [Rust newtypes](https://www.howtocodeit.com/articles/ultimate-guide-rust-newtypes) and [Parse dont validate](https://www.rustfinity.com/blog/parse-dont-validate) explain this idea much better. 

I found it very unfortunate we have domain models mixed with serialization.

crates/metastore/src/models/volumes.rs
```diff
@@ -81,6 +81,14 @@ pub struct IceBucketS3Volume {
...
+    #[validate(length(min = 1), custom(function = "validate_bucket_name"))]
+    pub catalog: Option<String>,
```
rampage644: Let's just go with `name`: `catalog: Option<String>` -> `name: String`. Having s3 table bucket volume without name is invalid and shouldn't be allowed.

crates/runtime/src/execution/session.rs
```diff
@@ -150,10 +150,8 @@ impl IceBucketUserSession {
                 .context(ex_error::DataFusionSnafu)?;
             let catalog_provider = Arc::new(catalog) as Arc<dyn CatalogProvider>;
 
-            let _catalog_name = volume.bucket.clone().unwrap_or_default().to_string();
-
-            // TODO set correct catalog name (s3 tables bucket contains "_" which is not supported by datafusion)
-            ctx.register_catalog("s3_tables", catalog_provider);
+            let catalog_name = volume.catalog.unwrap_or_else(|| "s3_tables".to_string());
```
rampage644: Not having a name should be an error

crates/runtime/src/execution/catalog/mod.rs
```diff
@@ -429,15 +445,35 @@ impl SchemaProvider for IceBucketDFSchema {
 #[derive(Debug)]
 pub struct IceBucketIcebergBridge {
     pub metastore: Arc<dyn Metastore>,
-    pub ident: IceBucketTableIdent,
+    pub database: String,
     pub object_store: Arc<dyn ObjectStore>,
 }
 
+impl IceBucketIcebergBridge {
+    fn new(metastore: Arc<dyn Metastore>, database: String) -> MetastoreResult<Self> {
```
rampage644, line 453: Can you elaborate why move object_store inside constructor? It looks like previously it was simply defined by the caller code

crates/runtime/src/execution/query.rs
```diff
@@ -456,6 +440,73 @@ impl IceBucketQuery {
...
+    #[allow(unused_variables)]
+    #[tracing::instrument(level = "trace", skip(self), err, ret)]
+    pub async fn create_iceberg_table(
+        &self,
+        catalog: Arc<dyn CatalogProvider>,
+        table_location: Option<String>,
+        ident: IceBucketTableIdent,
+        plan: LogicalPlan,
+    ) -> ExecutionResult<()> {
+        let iceberg_catalog =
```
rampage644, line 471: Am I correct this scary if/else statement could be avoided in we can somehow rely on CatalogProvider (or actually SchemaProvider) `.register_table()` method to actually create table?

osipovartem: Register table will not work since it is non-async. We have to write data with object store there in a separate thread. Register_table can create it's own block_on thread, but then deadlock appears during new thread creation when uploading data files.

crates/runtime/src/execution/catalog/mod.rs
osipovartem: Object store is required for most of Catalog methods in iceberg. It should be created once for catalog (database). Previously it was created per table to load tables. But in case we don't have table yet and need to create then object store will be empty, so it was moved to constructor since database can have just single object store.

crates/runtime/src/execution/session.rs
osipovartem: Will add validation

crates/runtime/src/execution/query.rs
rampage644: Okay, it took me quite a while to understand what this piece of code does and why it does it that way. Dumping all my thoughts here.

The code below relies on `iceberg_rust::catalog::Catalog` trait for different iceberg catalog implementations (REST, S3 table buckets, or anything else). 

In case this is built-in datafusion's `MemoryCatalog` we shortcut and execute logical plan and rely on its implementation to actually create a table.

Otherwise, code tries to downcast catalog to `iceberg_rust::catalog::Catalog` and if successful, calls its API to call external Iceberg catalog to do the job with `CreateTableBuilder.build()`.

@osipovartem I highly suggest we at least put this explanation here with comments.

rampage644 (approved):
Okay to merge with several assumptions
 * Still need several implementations to consolidate the logic somewhere (maybe stop relying on non-official `iceberg-rust` that much)
 * Do not try to coerce DDL commands/statement execution into `Catalog/Schema/TableProviders` API
 * Perhaps can simplify the logic and only support Iceberg REST. S3 tables now have such interface, and we can use our own catalog like this. Yes, again make http calls within same binary as we did previously.

osipovartem:
> Okay to merge with several assumptions
> 
> * Still need several implementations to consolidate the logic somewhere (maybe stop relying on non-official `iceberg-rust` that much)
> * Do not try to coerce DDL commands/statement execution into `Catalog/Schema/TableProviders` API
> * Perhaps can simplify the logic and only support Iceberg REST. S3 tables now have such interface, and we can use our own catalog like this. Yes, again make http calls within same binary as we did previously.

We can use just IcebergCatalog instead of our own catalog with single downcast

crates/metastore/src/models/volumes.rs
osipovartem: we have some funcs like get_s3_builder which work with IceBucketS3Volume only. In case we have separate volume type we have to duplicate all of them

osipovartem: Fixed

crates/runtime/src/execution/query.rs
osipovartem: Added comments

PR #444, opened 2025-04-16
Simplify naming for all icebucket entities (icebucket -> embucket)

rampage644:


osipovartem (approved):

bin/bucketd/Cargo.toml
```diff
@@ -7,7 +7,7 @@ license-file = { workspace = true }
 [dependencies]
 clap = { version = "4.5.27", features = ["env", "derive"] }
 dotenv = { version = "0.15.0" }
-icebucket_runtime = { path = "../../crates/runtime" }
+embucket_runtime = { path = "../../crates/runtime" }
```
DanCodedThis, line 10: Can we also remove embucket from `utils`, `runtime`, `metastore`, `history`? Is there a reason we keep it in the crate name?

DanCodedThis (approved):
LGTM, finally we removed the prefix!

bin/bucketd/Cargo.toml
rampage644: I would like to and also tried - found it more difficult since `runtime` and `utils` are very common names, and even with `metastore` there are already name collisions

rampage644:
Rebased on recent main, merging shortly

PR #446, opened 2025-04-16
Use custom logical plan to get missing schema

osipovartem:
- fixed drop table/view (check if exists)
- fixed BEGIN, StartTransaction since we call plan if the schema is empty (by this we call DF directly and don't take into account our own LogicalPlans)
- added status response entity for all SYSTEM sql like ALTER, BEGIN, SET, etc.
- recovered DBT seed

Related to #439

### DIFF
--- a/crates/runtime/src/execution/query.rs
+++ b/crates/runtime/src/execution/query.rs
@@ -80,6 +80,7 @@ pub struct QueryContext {
 pub struct UserQuery {
     pub metastore: Arc<dyn Metastore>,
     pub query: String,
+    pub parsed_query: String,
     pub session: Arc<UserSession>,
     pub query_context: QueryContext,
 }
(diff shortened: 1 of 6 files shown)


YaroslavLitvinov (approved):

rampage644 (approved):

crates/runtime/src/execution/query.rs
```diff
@@ -84,6 +84,7 @@ pub enum IceBucketQueryState {
 pub struct IceBucketQuery {
     pub metastore: Arc<dyn Metastore>,
     pub query: String,
+    pub parsed_query: String,
```
rampage644: We need better and more clear naming here - and comments about what is what and why. How about `raw_query`, `original_query` etc?

osipovartem: pub query: String, -> raw_query
pub parsed_query: String,-> query?

PR #450, opened 2025-04-17
Issues/426 global query history

### LINKED ISSUE #426: api: query history. Change GET behaviour: iterate over global query history If no worksheet_id is provided
Currently: we use default worksheet_id = 0 if no worksheet_id argument provided. So if no worksheet_id defined we add such queries into a default worksheet, and for get requests we just return queries belonging to that default worksheet.

Changes needed: 
* There are no default worksheets available for user. 
* GET request: If no worksheet provided then return/iterate on all the queries available for user, after applying filter if any (Filters feature). 
* When worksheet deleted it's query history still should be accessible in global query history 

### Important structural changes, relatively easy to implement:
* Query history items will be stored under /qh/{id} => {value} instead of /ws/{worksheet_id}/qh/{id}.
* /ws/{worksheet_id}/qh/{id} will store only a reference (empty value) to a query history item.
* Trivial to add query history unrelated to any worksheet: store directly as /qh/{id} => {value}.
* To associate a query with a worksheet:
  - Store full query: /qh/{id} => {value}
  - Add reference: /ws/{worksheet_id}/qh/{id} => ""
  - Avoid embedding query IDs directly in worksheet data due to possible repetition, as even a single query can be executed numerous amount of times
* Deleting a worksheet: [NOT IMPLEMENTED - Create issue for this]
  - Delete all /ws/{worksheet_id}/qh/* keys
  - Delete worksheet key /ws/{worksheet_id}
* Fetching history:
  - Global: iterate /qh/*
  - Worksheet-specific:
    * Read all history keys under /ws/{worksheet_id}/qh/*
    * Identify the minimum and maximum query IDs from that set
    * Build range /qh/{min}..=/qh/{max} and iterate  

Still an open question what they do in reference implementation when query run from worksheet this worksheet updates. So possibly queries are saved inside of worksheet or just updated access field refreshes entire worksheet item ?

YaroslavLitvinov:
Closes #426 

This PR introduces changes in how query history stored in worksheets: 
Store full query: `/qh/{id}` => {value}
Add reference: `/ws/{worksheet_id}/qh/{id}` => ""

Now if worksheet_id is not defined it will fetch all queries for GET, or add query to global query history if POST.

This PR doesn't delete query references upon worksheets removal, this need to be done in separate PR.


DanCodedThis (approved):
LGTM

PR #451, opened 2025-04-17
slatedb: Simple ScanIterator to remove transport level pagination

### LINKED ISSUE #423: slatedb: Imporved pagination and search for Metastore
## The problem
- The way **_SlateDB_** is setup in _**utils**_ and **_Metastore_** (and that it doesn't have a fully-fledged iterator) is forcing us to use transport-level API pagination (bad) 
- Analogy: it seems like UI is a dripping roof, and the water gets deeper than it should
- All the magic happens inside the scan function, not the **_SlateDB_** iterator
 
## Possible solution (good)
Something like a builder pattern: `state.metastore.list().volumes().get()` - and you can do things like add a `cursor`, `limit` or `search` by function and then with `get` you build the iterator and return the `Vec<T>` you need

### For normal uses:
- `state.metastore.list().volumes().get()`
- `state.metastore.list().databases().get()`
- `state.metastore.list().schemas(ident).get()` ( **_ident_** like in the impl right now)
- `state.metastore.list().tables(ident).get()` ( **_ident_** like in the impl right now)

### More sophisticated example 
`state.metastore.list().tables(ident).cursor(cursor).search(token).limit(10).get()` 

## Methods explanation
We can just leverage the rusts typesystem to not be able to call `fn cursor`, `fn get`, `fn limit` or `fn search` if no key
is chosen via `fn volumes`, `fn databases`, `fn schemas` or `fn tables` so these function can return like a `KeyfullListBuilder` with like a concrete type (we also can have `type DatabaseListBuidler = ListBuidler<IceBucketDatabase>`  maybe)
 
```Rust
KeylessListBuilder {
    key: Option<String>
}
KeyfullListBuilder {
    key: String,
    cursor: Option<String>,
    limit: Option<usize>,
    token: Option<String>
}

//In Metastore
fn list(&self) -> KeylessListBuilder {
    return KeylessListBuiler::new()
}
 
 
 impl KeylessListBuilder {
     fn volumes(&self) -> KeyfullListBuilder //takes a key for volumes
     fn databases(&self) -> KeyfullListBuidler //takes key for databases
     fn schemas(&self, ident: IceBucketSchemaIdent) -> KeyfullListBuilder //takes key for schemas and formats it with the ident
     fn tables(&self, ident: IceBucketTableIdent) -> KeyfullListBuilder //takes key for tables and formats it with the ident
 }
 
impl KeyfullListBuilder { 
//maybe use an option in all `fn` paramas as all `Query` parameters are options
    fn cursor(&self, cursor: String) -> KeyfullListBuilder //adds a cursor
    fn limit(&self, limit: usize) -> KeyfullListBuilder //add a limit
    fn search(&self, token: String) -> KeyfullListBuilder //add search token
    fn get(&self) -> Vec<T> //similar to `fn list_objects` in utils crate lib.rs file 
}
```
 
 ## Alternatively
 We can just return from `list_volumes`, `list_databases`, etc - `ListBuilder::new(key)` with a chosen key like (more simple approach, however if more entities will be added will need to add them inside **_Metastore_**, instead of the **_list_builder.rs_** file). The first approach should be more idiomatic (maybe wrong). However, one might ask does the `ListBuilder` should know about volumes, databases, schemas or tables (not in the context of objects of course)? 
 
 ```Rust
ListBuilder {
    key: String,
    cursor: Option<String>,
    limit: Option<usize>,
    token: Option<String>
}

impl ListBuidler { ... } //all methods here, are the same, only new is different (added to siganture `key: String`)
 ```

DanCodedThis:
Closes #423

- Added an async `ScanIterator` trait (you need to implement sadly `collect` and not `next` which would be more flexible) and you can collect to any type
- Added a `VecScanIterator` which implements this trait which collects to a vec and takes into account: limit, search & cursor
- No `map` or `filter` (for now or ever)
- "Rebased" all api on top of these change
- Now you call `state.metastore.iter_volumes.collect().await`, `state.metastore.iter_databases.collect().await`you can do thing like:
- You can also do this (pagination, search, limit): `state.metastore.iter_tables(ident).cursor(cursor).limit(limit).token(search).collect().await`
- Improvements later:
  - Erroring
  - Addition of other methods such as `map` or`filter` (but they just only save us an extra `collect`, so not much)

P.S. This is a new branch, old designs are available on another branch similarly named, older commits could show you the idea behind `next` recursion which would allow making `map` `filter` & any `iter` methods chainable for any number of times in any sequence (Old branch: https://github.com/Embucket/embucket/tree/DanCodedThis/423-slatedb-scan-iterator-metastore)


rampage644 (approved):
looks like a good way forward

PR #454, opened 2025-04-18
serve static assets for frontend

### LINKED ISSUE #448: serving static frontend from backend
* Serve some compiled static data on path like `/index.html` 
* Run web assets server on 8080 port
* Nothing changed for service on 3000 port

YaroslavLitvinov:
Closes #448

- [x] serve web assets taken from tarball that compiled into a binary
- [x]  tests added

### DIFF
--- a/crates/runtime/src/http/web_assets/handler.rs
+++ b/crates/runtime/src/http/web_assets/handler.rs
@@ -15,22 +15,20 @@
 // specific language governing permissions and limitations
 // under the License.
 
+use super::error::{
+    ArchiveError, BadArchiveSnafu, HandlerError, NonUnicodeEntryPathInArchiveSnafu,
+    ReadEntryDataSnafu, ResponseBodySnafu, Result,
+};
 use axum::{
-    body::Body, extract::Path, http::{header, StatusCode}, response::{IntoResponse, Redirect, Response}
+    body::Body,
+    extract::Path,
+    http::{header, StatusCode},
+    response::{IntoResponse, Redirect, Response},
 };
 use mime_guess;
+use snafu::ResultExt;
 use std::io::Cursor;
 use std::io::Read;
-use snafu::ResultExt;
-use super::error::{
-    Result,
-    ArchiveError,
-    HandlerError,
-    BadArchiveSnafu,
-    ResponseBodySnafu,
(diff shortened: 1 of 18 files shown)


crates/runtime/src/http/web_assets/handler.rs
```diff
@@ -0,0 +1,104 @@
...
+    response::{IntoResponse, Redirect, Response},
+};
+use mime_guess;
+use snafu::ResultExt;
+use std::io::Cursor;
+use std::io::Read;
+
+// Alternative to using tarball is the rust-embed package
+
+const WEB_ASSETS_TARBALL: &[u8] = include_bytes!("../../../../../dist.tar");
```
rampage644: Any chance we could get rid of so many `..`? ;)

YaroslavLitvinov: By moving const to the top dir level or so, I guess.

rampage644: I mean, can we get actual directory path (absolute or relative) without relying on `..`?  It looks like somewhat different to manage and error prone

YaroslavLitvinov: I don't think we can just add dist files into the binary, as normally these artefacts are not a part of source code. They should be built from source code for debug, release builds in target environment (dev, ci/cd). Docker probably is the most  robust way to build it regardless of environment.

So I think following flow should be working: 
* -> Docker builds ui/dist and places it to a mounted volume 
* -> cargo build script creates tar and defines compile time env var to be used like this `include_bytes!(env!("WEB_ASSETS_TARBALL_PATH"))`.
* -> Build `bucketd` finally.

rampage644: Yeah, totally - i was asking if there is anything like `WEB_ASSETS_TARBALL_PATH` instead of `../../../`

YaroslavLitvinov: `../../../` can replaced by something more convenient to read in the source code, by increasing complexity in other parts and hiding things in ./cargo/config and build script

YaroslavLitvinov:
There is one more commit is expected to use `ui/dist` instead of just  `dist`

rampage644 (approved):

PR #467, opened 2025-04-22
ANY_VALUE Aggregation Function

Vedin:
After the latest discussion about repo structure, I just moved PR here from the datafusion fork https://github.com/Embucket/datafusion/pull/19. 
This implementation works both for aggregate and window functions. I didn't include window optimization here due to limitations of how the window function is replaced in the context provider. We can add it later if we need to.


osipovartem (approved):

rampage644 (approved):

ravlio (approved):

ravlio (commented):
do you plan to add DISTINCT?

Vedin:
> do you plan to add DISTINCT?

In fact it works from the box due to snowflake docs
> The DISTINCT keyword can be specified for this function, but it does not have any effect.

PR #471, opened 2025-04-22
Flatten table function

### LINKED ISSUE #476: Flatten Table Function
Implement Snowflake analogue FLATTEN function. Description: https://docs.snowflake.com/en/sql-reference/functions/flatten
Without SQL syntax changes

Flattens (explodes) compound values into multiple rows.

FLATTEN is a table function that takes a JSON STRING column and produces a lateral view (that is, an inline view that contains correlations to other tables that precede it in the [FROM](https://docs.snowflake.com/en/sql-reference/constructs/from) clause).

FLATTEN can be used to convert semi-structured data to a relational representation.

ravlio:
Closes #476


crates/runtime/src/execution/datafusion/functions/table/flatten.rs
```diff
@@ -0,0 +1,852 @@
...
+}
+
+impl TableFunctionImpl for FlattenTableFunc {
+    fn call(&self, args: &[Expr]) -> DFResult<Arc<dyn TableProvider>> {
+        if args.len() != 5 {
+            return plan_err!("flatten() expects 5 args: INPUT, PATH, OUTER, RECURSIVE, MODE");
+        }
+
+        self.row_id.fetch_add(1, Ordering::Acquire);
+        let ScalarValue::Utf8(Some(input_str)) = eval_expr(&args[0])? else {
```
osipovartem, line 256: Should it also work with ColumnarValue::Array column?

osipovartem (approved):

crates/runtime/src/execution/datafusion/functions/table/flatten.rs
ravlio: I'm not sure. For the first glance it is not necessary

crates/.DS_Store
rampage644, line 1: Included by mistake?

rampage644: Oh, it's removed, disregard

crates/runtime/src/execution/datafusion/functions/table/flatten.rs
```diff
@@ -0,0 +1,852 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::execution::datafusion::functions::parse_json::ParseJsonFunc;
+    use arrow::util::pretty::print_batches;
+    use datafusion::prelude::SessionContext;
+    use datafusion_common::assert_batches_eq;
+    use std::sync::Arc;
+
+    // fixme
```
rampage644: I would mark this as expected to fail it it fails but should pass or at least document what is expected behavior and what should be fixed

rampage644 (approved):
lgtm with minor nitpicks

crates/runtime/src/execution/datafusion/functions/table/flatten.rs
ravlio: let me just remove this test. Ideally we should see the failing test in our slt test results. Btw at first glance it is hard to fix, they treat broken json `[1,,3]` as a valid and we can't parse this.

crates/runtime/src/execution/datafusion/functions/table/mod.rs
```diff
@@ -0,0 +1,26 @@
+// Licensed to the Apache Software Foundation (ASF) under one
```
rampage644: Also, let's remove the license from here - I had removed license header from source file and left only `LICENSE` in root dir. Github status check is also removed.

PR #478, opened 2025-04-23
Fix create schema and incorrect empty plan execution during table creation

osipovartem:
- empty input in plan triggering empty snapshot creation
- fixed create schema for iceberg catalog


DanCodedThis (approved):
LGTM

PR #479, opened 2025-04-24
Set correct columns name and default limit

osipovartem:
- set correct column name instead of COALESCE(CAST({column_name} AS STRING), 'Unsupported')
- set default limit since the full dataset is returned (10K-20K rows takes 70MB..)


DanCodedThis (approved):
LGTM

Issue #458, opened 2025-04-21
web assets server: release build failed


YaroslavLitvinov:
# release build failed.

Note: Disregard merged PR, this issue remains relevant until opposite is noted.

Snippet from github workflow :
```
 [builder 7/7] RUN cargo build --release:
     : couldn't read /app/ui/dist.tar: No such file or directory (os error 2)
    1955.6   --> crates/runtime/src/http/web_assets/handler.rs:34:35
```
https://github.com/Embucket/embucket/actions/runs/14579167993/job/40891834811

Details:
There is no evidence that build script ever ran during build session, since tar file can't be located as well as no errors in build log. 
Environment variable WEB_ASSETS_TARBALL_PATH been set in .cargo/config, and not in build script.

Future steps:
Add more error handling to `build.rs` script, to make it easier for debugging.

YaroslavLitvinov:
Docker successfully builds release locally

```
yaroslav@:~/git/embucket$ docker build --platform linux/amd64  . 
[+] Building 818.3s (18/18) FINISHED                                                                                                                                                                       docker:default
 => [internal] load build definition from Dockerfile                                                                                                                                                                 0.0s
 => => transferring dockerfile: 801B                                                                                                                                                                                 0.0s
 => [internal] load metadata for docker.io/library/debian:bookworm-slim                                                                                                                                              0.5s
 => [internal] load metadata for docker.io/library/rust:latest                                                                                                                                                       0.5s
 => [internal] load .dockerignore                                                                                                                                                                                    0.0s
 => => transferring context: 2B                                                                                                                                                                                      0.0s
 => [builder 1/7] FROM docker.io/library/rust:latest@sha256:7b65306dd21304f48c22be08d6a3e41001eef738b3bd3a5da51119c802321883                                                                                        39.6s
 => => resolve docker.io/library/rust:latest@sha256:7b65306dd21304f48c22be08d6a3e41001eef738b3bd3a5da51119c802321883                                                                                                 0.0s
 => => sha256:23b7d26ef1d294256da0d70ce374277b9aab5ca683015073316005cb63d33849 48.49MB / 48.49MB                                                                                                                     3.7s
 => => sha256:07d1b5af933d2dfc3d0dd509d6e20534825e4a537f7b006a6cb5b8e5a1f20905 24.01MB / 24.01MB                                                                                                                     1.7s
[...]

PR #482, opened 2025-04-24
Issues/428 query history features

### LINKED ISSUE #428: api: query history descending sort, keys using inverted query record ids
### SlateDB restrictions
* SlateDB iterator can only iterate in ascending direction
* in SlateDB every new seek() should seek keys in single ascending order, as backward seek not supported too.
  That means descending iterator by inverted indexes can't be used to seek on current query records iterator that goes in ascending direction. So we need to save QueryRecord directly using inverted key to access their values in descending order.

### sort order - Descending
By requirement BE should support only descending sort order by query_start_time field. 
So we will repurpose current keys /qh/{qh_id}, /ws/{worksheet_id}/qh/{qh_id} so it will use inverted query record id for {qh_id} in resulted key, to store QueryRecord value. No duplicated QueryRecord values will be saved by non-inverted key ids for ascending support! 

### Descending order implementation
* Nothing changes except using inverted keys

### LINKED ISSUE #469: api: delete worksheet handler: update logic
Originally it was described in issue  #426, that was preliminary closed. Copying requirements here.

Update Delete worksheet functionality:
* Delete all /ws/{worksheet_id}/qh/* keys
* Delete worksheet key /ws/{worksheet_id}

YaroslavLitvinov:
Added more error handling for #458
Closes #428 
Closes #469

### DIFF
--- a/bin/bucketd/build.rs
+++ b/bin/bucketd/build.rs
@@ -23,7 +23,10 @@ use tar::Builder;
 fn create_web_assets_tarball() -> Result<(), std::io::Error> {
     let source_path = Path::new(env!("WEB_ASSETS_SOURCE_PATH"));
     if !source_path.exists() {
-        eprintln!("Source path does not exist: {}", source_path.display());
+        return Err(std::io::Error::new(
+            std::io::ErrorKind::NotFound,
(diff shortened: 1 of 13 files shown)


DanCodedThis (approved):
LGTM

PR #489, opened 2025-04-25
Check create table input logical plan effectively empty

osipovartem:
- fix **merge into** pattern
- check if logical plan effectively empty (don't create empty parquet files)
Related to #439


DanCodedThis (approved):
LGTM

Issue #425, opened 2025-04-14
api: query / history interfaces implement ExecutionService trait


YaroslavLitvinov:
# implement ExecutionService trait

This issue was repurposed to focus on a singe issue. Please update this issue accordingly and feel free to comment.

History and Query interfaces implementing the same ExecutionService trait

Consider using query interface with enabled query history, proposed by @rampage644 
```
My proposal was to extract exec service into trait/interface and then we could implement HistoryExecService implementing that interface but delegating heavy lifting to real exec service:

pub trait ExecutionService {
    pub async fn query(
        &self,
        query: &str,
        query_context: IceBucketQueryContext,
    ) -> ExecutionResult<(Vec<RecordBatch>, Vec<ColumnInfo>)>;
}



pub struct HisotryExecutionService {
    pub fn new(innner: Arc<dyn ExecutionService>) -> Self {
        Self { innner }
    }

}

impl ExecutionService for HisotryExecutionService {
    pub async fn query(
        &self,
        query: &str,
        query_context: IceBucketQueryContext,
    ) -> ExecutionResult<(Vec<RecordBatch>, Vec<ColumnInfo>)> {
        
        let start = time::Instant::now();
        self.innner.query(query, query_context).await
        let duration = start.elapsed();

        self.save(
            query,
            query_context,
            duration.as_millis(),
        );
    }
}
```

PR #494, opened 2025-04-28
DDL statements in query history (#470)

### LINKED ISSUE #470: api: use query engine for creating databases, schemas instead of using metastore functions
As it was discussed earlier we need to have a complete history of all schema mutations we do via ui handlers. 
This would also fix side effects like absence of standard schemas in created database (link relevant issue). As reported by @nikitastryuk .

That means all the DDL statements should appear in query history. So instead of using metastore directly we should use query engine.

This relevant to create / drop / alter
* databases
* schemas
* tables ?

DanCodedThis:
Closes #470

- Schemas created using the query engine
- Databases cannot be created through the query engine yet
- No `public` or `information` schema creation for databases for now


crates/runtime/src/http/ui/schemas/handlers.rs
```diff
@@ -64,25 +65,61 @@ pub struct ApiDoc;
 )]
 #[tracing::instrument(level = "debug", skip(state), err, ret(level = tracing::Level::TRACE))]
 pub async fn create_schema(
+    DFSessionId(session_id): DFSessionId,
```
osipovartem, line 68: Why do we need such thing? create schema is already a part of query engine and history already creating for it
why we cannot just send "create schema" in sql editor?
and remove create schema endpoint from UI:)

DanCodedThis: It's a button on the UI, but I agree, we can send a request from UI, to the query endpoint no need for this, then all the "heavy" lifting will be done by UI

@nikitastryuk what do you think?

rampage644: @osipovartem We do not want to send any SQL queries from the frontend and effectively keep ui-specific API separate from other API (like `/v1/query` etc). 

This way all the logic to interact with the engine stays on the backend part in a single place.

crates/runtime/src/http/ui/schemas/handlers.rs
```diff
@@ -64,25 +65,61 @@ pub struct ApiDoc;
...
-                data: Schema::from(rw_object),
-            })
-        })
+    let sql_string = format!(
+        "CREATE SCHEMA {}.{}",
+        database_name.clone(),
+        payload.name.clone()
+    );
+    //TODO: figure out how to unify this, with the delete, instead of copying code (possibly a generic query history `fn`)
+    let mut query_record = QueryRecord::query_start(&sql_string, None);
```
rampage644, line 83: Is this here just to make sure we record all sql queries? If so - I agree with you previous comment: this should not even be here. 

Instead this handler should know nothing about query history. Since `state.execution_svc` has type `Arc<ExecutionService>` we can wire it during http server construction to be an instance of `RecordableExecutionService` that wraps `ExecutionService` and performs actual recording.

crates/runtime/src/http/ui/schemas/handlers.rs
osipovartem: Create schema is not UI specific query. Since we have to support it for snowflake dialect we will have two places where we can create a schema. What about other catalogs?  If we going to add external catalogs, we have to write a lot of logic to support schema creation via UI for s3 tables, and other.

osipovartem: I don't mean send query from ui, I mean that user should open sql-editor and write query, as usual, to create table, schema, external snowflake storage, etc.

rampage644: My point was "Send `CREATE SCHEMA` statement from the backend not the frontend".

crates/runtime/src/http/ui/schemas/handlers.rs
DanCodedThis: Yes ideally this should be a trait

crates/runtime/src/http/ui/schemas/handlers.rs
DanCodedThis: > I don't mean send query from ui, I mean that user should open sql-editor and write query, as usual, to create table, schema, external snowflake storage, etc.

As far as I know this endpoint is used in 1-2 places where the `schema` (also `DB`s, but not `table`s) are created through a form (like in Snowflake UI)

DanCodedThis: > Create schema is not UI specific query. Since we have to support it for snowflake dialect we will have two places where we can create a schema. What about other catalogs? If we going to add external catalogs, we have to write a lot of logic to support schema creation via UI for s3 tables, and other.

This is a valid point to velocity

> > I don't mean send query from ui, I mean that user should open sql-editor and write query, as usual, to create table, schema, external snowflake storage, etc.
> 
> As far as I know this endpoint is used in 1-2 places where the `schema` (also `DB`s, but not `table`s) are created through a form (like in Snowflake UI)

We can remove those, but should we though? 

cc @rampage644

crates/runtime/src/http/ui/schemas/handlers.rs
DanCodedThis: > Is this here just to make sure we record all sql queries? If so - I agree with you previous comment: this should not even be here.
> 
> Instead this handler should know nothing about query history. Since `state.execution_svc` has type `Arc<ExecutionService>` we can wire it during http server construction to be an instance of `RecordableExecutionService` that wraps `ExecutionService` and performs actual recording.

Yes, I copied it from our `query` endpoint

YaroslavLitvinov: We were discussing this earlier, how query interface should be looking
https://github.com/Embucket/embucket/issues/425

DanCodedThis: > Yes ideally this should be a trait

Also, I don't like the structure that the `history` is it's own service (kinda of), and even more, that it is its own crate that exists separately (which it shouldn't, since it's a part of execution (at least it should've been relying on it)), but that's more of nitpick, and maybe too big to tackle now

The trait may suffice for now, however not all queries should be recorded

crates/runtime/src/http/ui/schemas/handlers.rs
rampage644: > I don't mean send query from ui, I mean that user should open sql-editor and write query, as usual, to create table, schema, external snowflake storage, etc.

This is possible, the original requirement comes from making a UI as accessible to non-tech savvy users as it is in Snowflake (and thus it has different ways to create a schema, via sql and not). 

This original requirement still holds.

crates/runtime/src/http/ui/schemas/handlers.rs
rampage644: ~~Okay, happy to approve current approach with the promise this will eventually be replaced.~~ I do agree we do want a flexibility on which query to record, just haven't given a proper thought yet what is the correct interface should be (perhaps moving this logic into recorder makes sense, perhaps not).  

+1 on moving `history` module closer to execution.

**Having second thoughts on the PR scope**: i think we can split it into two separate - one that replaces metastore service calls with calls to query engine and another that introduces proper interface (trait) for execution service + refactoring.

Happy to move the discussion to relevant issue.

YaroslavLitvinov: I do support opinion that discussion about history as a module vs crate is irrelevant to this PR. 

Instead, using some history interface is much more relevant. 
And using such interface would replace most of the code in this PR.
I would prioritize implementing HistoryExecutionService trait as described in issue #425, or whatever proposed in this PR. I will repurpose issue #425 so that clearly states the proposal.

DanCodedThis:
More context on what was done:
- `Service` is a trait implemented by `struct History` and `struct Execution`
- To avoid circular dependency it is done inside the `runtime` crate
- `QueryContext` now contains the `worksheet_id`, again UI is a dripping roof, that drips down to internals (not good)
- If you have other solutions to pass down `worksheet_id` feel free to suggest it
- Fixed some tests
- Fixed `make_app`, `run_binary`, etc
- DDL statements for `Schemas` included here (query history)
- Fixed anything else to adhere to new changes

P.S. Sorry for breaking changes to anybody in advance

DanCodedThis:
CC: @osipovartem @rampage644 @YaroslavLitvinov

crates/runtime/src/execution/service.rs
```diff
@@ -23,23 +23,47 @@ use uuid::Uuid;
 
 use super::error::{self as ex_error, ExecutionError, ExecutionResult};
 
-pub struct ExecutionService {
+#[async_trait::async_trait]
+pub trait Service: Send + Sync {
```
rampage644: I would keep `ExecutionService` as a name and instead rename original struct to something else

crates/runtime/src/execution/history.rs
```diff
@@ -0,0 +1,95 @@
...
+use crate::execution::utils::Config;
+use crate::http::ui::queries::models::ResultSet;
+use arrow::csv::reader::Format;
+use arrow_array::RecordBatch;
+use bytes::Bytes;
+use embucket_history::{QueryRecord, QueryRecordActions, WorksheetsStore};
+use embucket_metastore::TableIdent as MetastoreTableIdent;
+use std::sync::Arc;
+
+pub struct History {
```
rampage644: Rename for better expressiveness (RecordingExecutionService, TrackingExecutionService, InstrumentedExecutionService, LoggingExecutionService)

crates/runtime/src/execution/service.rs
rampage644: CoreExecutionService maybe

DanCodedThis: Renamed

crates/runtime/src/execution/history.rs
DanCodedThis: Renamed to RecordingExecutionService

DanCodedThis:
Requested changes, cc: @rampage644

rampage644:
> If you have other solutions to pass down worksheet_id feel free to suggest it

What if we somehow pass either a closure or a builder-something-pattern such that when `.query(sql, execution_context)` it has access to variable from the outer scope (worksheetId in our case)?

DanCodedThis:
> > If you have other solutions to pass down worksheet_id feel free to suggest it
> 
> What if we somehow pass either a closure or a builder-something-pattern such that when `.query(sql, execution_context)` it has access to variable from the outer scope (worksheetId in our case)?

It would either break the signature or it will be similar to `ScanIterator` but way way worse (since it would mean that the **_trait_** `fn query` would need to return some struct, and not record batches)

DanCodedThis:
> > > If you have other solutions to pass down worksheet_id feel free to suggest it
> > 
> > 
> > What if we somehow pass either a closure or a builder-something-pattern such that when `.query(sql, execution_context)` it has access to variable from the outer scope (worksheetId in our case)?
> 
> It would either break the signature or it will be similar to `ScanIterator` but way way worse (since it would mean that the **_trait_** `fn query` would need to return some struct, and not record batches)

Perhaps @osipovartem has any ideas?

DanCodedThis:
> > > If you have other solutions to pass down worksheet_id feel free to suggest it
> > 
> > 
> > What if we somehow pass either a closure or a builder-something-pattern such that when `.query(sql, execution_context)` it has access to variable from the outer scope (worksheetId in our case)?
> 
> It would either break the signature or it will be similar to `ScanIterator` but way way worse (since it would mean that the **_trait_** `fn query` would need to return some struct, and not record batches)

Also, a thing to keep in mind is that stable Rust doesn't allow default values in signatures (but nightly should :) )

DanCodedThis:
Maybe we could do something through the `Query object`, but again we either break the signature or the result of the `fn query`.

We could surley, if the `QueryContext` values are `Option`s, make it option it self, meaning even for normal querying we would need to use like a builder pattern, but all our queries require the inside of the `QueryContext` I think, so maybe a bad lead

DanCodedThis:
> Maybe we could do something through the `Query object`, but again we either break the signature or the result of the `fn query`.
> 
> We could surley, if the `QueryContext` values are `Option`s, make it option it self, meaning even for normal querying we would need to use like a builder pattern, but all our queries require the inside of the `QueryContext` I think, so maybe a bad lead

I would merge for now, if there are no other suggestion/problems. And later down the line, maybe I or anybody else comes up with something better (like `ScanIterator`)

crates/runtime/src/execution/recording_service.rs
```diff
@@ -0,0 +1,95 @@
...
+use crate::execution::utils::Config;
+use crate::http::ui::queries::models::ResultSet;
+use arrow::csv::reader::Format;
+use arrow_array::RecordBatch;
+use bytes::Bytes;
+use embucket_history::{QueryRecord, QueryRecordActions, WorksheetsStore};
+use embucket_metastore::TableIdent as MetastoreTableIdent;
+use std::sync::Arc;
+
+pub struct RecordingExecutionService {
```
rampage644: Would be great to cover this with tests:

1. happy case, query executed with success
2. happy case, query failed
3. post-processing, failure to store the results

crates/runtime/src/execution/recording_service.rs
```diff
@@ -0,0 +1,95 @@
...
+    pub execution: Arc<CoreExecutionService>,
```
rampage644: Can we depend here on `dyn ExecutionService` ?

rampage644:
@DanCodedThis Overall approach looks fine with me. However, this PR is already big enough and is not complete. 

For this PR to land, let's please split it into several smaller ones:
 1. One that groups DDL related changes, metastore -> SQL queries in handlers (#470)
 2. Introduction ExecutionService trait and refactoring `struct ExecutionService` -> `struct CoreExecutionService`
 3. Introduction `RecordingExecutionService` and refactoring handlers to remove unrelated logic from there
 
 Approving this PR only as 1st. Assuming two more are coming as follow ups.

rampage644 (approved):
approve not to block, assuming split this one into 3 different

DanCodedThis:
Moved `ExecutionService` trait to another branch

rampage644 (approved):

Issue #492, opened 2025-04-28
catalog: Prepare unified trait for any catalog


osipovartem:

osipovartem:
This task is done by added DF Catalog, Schema adapters for already existing IcebergCatalog implementations

PR #512, opened 2025-04-29
Create cache layer for catalogs

osipovartem:
Related to #492


DanCodedThis (approved):
LGTM

crates/runtime/src/execution/catalog/catalog_list.rs
```diff
@@ -43,168 +42,163 @@ impl EmbucketCatalogList {
...
+    }
 
-                    let table_provider = match table.format {
-                        embucket_metastore::TableFormat::Parquet => {
-                            let parq_read_options = ParquetReadOptions::default();
-                            let listing_options = parq_read_options.to_listing_options(
-                                ctx.state().config(),
-                                ctx.state().default_table_options(),
-                            );
+    pub async fn s3_tables_catalogs(&self) -> ExecutionResult<Vec<Catalog>> {
```
rampage644: `external_catalogs` seem like a good fit here

crates/runtime/src/execution/catalog/catalogs/embucket/schema.rs
```diff
@@ -36,34 +35,87 @@ impl SchemaProvider for EmbucketSchema {
...
-        let Ok(tabular) = self
-            .iceberg_catalog
-            .clone()
-            .load_tabular(&Identifier::new(&[self.schema.clone()], name))
-            .await
-        else {
-            return Ok(None);
-        };
+        std::thread::spawn(move || {
+            let Ok(rt) = tokio::runtime::Runtime::new() else {
```
rampage644: Can we reuse `try_current()`? https://docs.rs/tokio/latest/tokio/runtime/struct.Handle.html#method.try_current

crates/runtime/src/execution/catalog/catalog_list.rs
osipovartem: In case we have 2-3 additional catalogs it will be mess of code (initial version was external_catalogs :)

crates/runtime/src/execution/catalog/catalog.rs
```diff
@@ -0,0 +1,52 @@
+use crate::execution::catalog::schema::Schema;
+use dashmap::DashMap;
+use datafusion::catalog::{CatalogProvider, SchemaProvider};
+use std::{any::Any, sync::Arc};
+
+#[derive(Clone)]
+pub struct Catalog {
```
rampage644: What's the difference/relationship between `Catalog` and `EmbucketCatalog`?

rampage644: I feel like this Catalog (implementing `CatalogProvider`) is a cached version of wrapped (actual) CatalogProvider - is that correct? If so, would it make more sense to name it explicitly so - like `CachingCatalog` etc?

osipovartem: Catalog is a cache under real catalog_providers implementations, the same as Schema. 
CatalogsProviderList hanlde Catalogs.
 EmbucketCatalog is specific implementation of our own catalog

crates/runtime/src/execution/catalog/catalogs/embucket/schema.rs
```diff
@@ -36,34 +35,87 @@ impl SchemaProvider for EmbucketSchema {
...
+                    Ok(tables) => tables.into_iter().map(|s| s.ident.table.clone()).collect(),
+                    Err(_) => vec![],
+                }
+            })
+        })
+        .join()
+        .unwrap_or_else(|_| vec![])
     }
 
     async fn table(&self, name: &str) -> Result<Option<Arc<dyn TableProvider>>, DataFusionError> {
```
rampage644, line 56: Why do we need runtime and spawn and block inside? Method is `async` so we don't need to block inside

crates/runtime/src/execution/catalog/catalog.rs
rampage644: So it is the cache - let's go with more explicit nameing then (I suggest `CachingCatalog`), it's a bit confusing otherwise. Same for `Schema` --> `CachingSchema`

rampage644 (approved):
lgtm overall

crates/runtime/src/execution/catalog/catalogs/embucket/schema.rs
osipovartem: Service panicked: Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.

PR #543, opened 2025-04-30
Recording Execution Service

### LINKED ISSUE #541: Recording Execution Service
- Implement `ExecutionService` trait for the `RecordingExecutionService` (#425)
- Change the behavior of the `fn query` to use query history by default
- Change `make_app` and `run binary` to use the new `RecordingExecutionService`

DanCodedThis:
Closes #541 & #425

- `RecordingExecutionService` is a struct implementing the `ExecutionService` trait
- To avoid circular dependency it is done inside the `runtime` crate
- `QueryContext` now contains the `worksheet_id`, again UI is a dripping roof, that drips down to internals (not good)
- If you have other solutions to pass down `worksheet_id` feel free to suggest it
- Fixed some tests
- Added tests for `RecordingExecutionService`

### DIFF
--- a/crates/runtime/src/execution/recording_service.rs
+++ b/crates/runtime/src/execution/recording_service.rs
@@ -1,0 +1,96 @@
+use crate::execution::error::ExecutionResult;
+use crate::execution::models::ColumnInfo;
+use crate::execution::query::QueryContext;
+use crate::execution::service::ExecutionService;
+use crate::execution::utils::Config;
+use crate::http::ui::queries::models::ResultSet;
+use arrow::csv::reader::Format;
+use arrow_array::RecordBatch;
+use bytes::Bytes;
+use embucket_history::{QueryRecord, QueryRecordActions, WorksheetsStore};
+use embucket_metastore::TableIdent as MetastoreTableIdent;
+use std::sync::Arc;
+
+pub struct RecordingExecutionService {
+    pub execution: Arc<dyn ExecutionService>,
+    pub store: Arc<dyn WorksheetsStore>,
+}
+
+//TODO: add tests
+impl RecordingExecutionService {
+    pub fn new(execution: Arc<dyn ExecutionService>, store: Arc<dyn WorksheetsStore>) -> Self {
+        Self { execution, store }
+    }
+}
+
+#[async_trait::async_trait]
+impl ExecutionService for RecordingExecutionService {
+    async fn create_session(&self, session_id: String) -> ExecutionResult<()> {
+        self.execution.create_session(session_id).await
+    }
+
+    async fn delete_session(&self, session_id: String) -> ExecutionResult<()> {
+        self.execution.delete_session(session_id).await
+    }
+
+    async fn query(
+        &self,
+        session_id: &str,
+        query: &str,
+        query_context: QueryContext,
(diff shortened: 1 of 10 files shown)


DanCodedThis:
> @YaroslavLitvinov @osipovartem @rampage644 What would you suggest for testing?
> 
> Since we don't test the `CoreExecutionService` internally (only in external tests) and the `Query History / Worksheet store` is tested in its own crate. Also, this could be considered just refactoring. However, there are tests in handlers reflecting this functionality (which are part of the said "external tests").

Would be enough to just add extra http test, maybe? Since it's the main use case.

DanCodedThis:
Disregard the deleted comments about tests.

crates/runtime/src/http/mod.rs
```diff
@@ -47,7 +48,11 @@ pub fn make_app(
     config: &WebConfig,
 ) -> Result<Router, Box<dyn std::error::Error>> {
     let execution_cfg = execution::utils::Config::new(&config.data_format)?;
-    let execution_svc = Arc::new(CoreExecutionService::new(metastore.clone(), execution_cfg));
+    let execution = Arc::new(CoreExecutionService::new(metastore.clone(), execution_cfg));
```
rampage644:
```rust
let execution_svc = Core::new()
let execution_svc = Record::new(execution_svc)
```

crates/runtime/src/execution/recording_service.rs
```diff
@@ -0,0 +1,96 @@
...
+
+//TODO: add tests
+impl RecordingExecutionService {
+    pub fn new(execution: Arc<dyn ExecutionService>, store: Arc<dyn WorksheetsStore>) -> Self {
+        Self { execution, store }
+    }
+}
+
+#[async_trait::async_trait]
+impl ExecutionService for RecordingExecutionService {
```
rampage644, line 27: Possible to have some test coverage for this?

DanCodedThis: > Possible to have some test coverage for this?

@rampage644 Working on it

crates/runtime/src/http/mod.rs
DanCodedThis: > ```rust
> let execution_svc = Core::new()
> let execution_svc = Record::new(execution_svc)
> ```

@rampage644 Do you want to also change the names (possible just `as Core` & `as Record`, etc in `use`) and/or always return `Arc` from `fn new`?

rampage644: no, i just meant redefining `exection_svc` is perfectly valid pattern (no intermediate `let execution`)

DanCodedThis: Done

crates/runtime/src/execution/recording_service.rs
DanCodedThis: Done

rampage644 (approved):

YaroslavLitvinov:
Just define worksheet id in execution service and use then use it from history?

DanCodedThis:
> Just define worksheet id in execution service and use then use it from history?

What did you mean?

YaroslavLitvinov:
Either use i64 in execution interface instead of worksheet id

PR #559, opened 2025-05-01
Update to datafusion 47.0.0. SlateDB to 0.6.1. 

Vedin:


crates/utils/src/lib.rs
```diff
@@ -114,7 +114,7 @@ impl Db {
     pub async fn put<T: serde::Serialize + Sync>(&self, key: &str, value: &T) -> Result<()> {
         let serialized = ser::to_vec(value).context(SerializeValueSnafu)?;
         self.0
-            .put(key.as_bytes(), serialized.as_ref())
+            .put::<&[u8], &[u8]>(key.as_bytes(), serialized.as_ref())
```
DanCodedThis: Nitpick: Actually when I was trying to update SlateDB, I found that we just need to remove `.as_ref()` on the `serialized` since it would do the automatic conversion itself, no need for the `turbofish` operator (`::<>`)

crates/utils/src/lib.rs
```diff
@@ -204,7 +204,7 @@ impl Db {
     ) -> Result<()> {
         let serialized = ser::to_vec(entity).context(SerializeValueSnafu)?;
         self.0
-            .put(entity.key().as_ref(), serialized.as_ref())
+            .put::<&[u8], &[u8]>(entity.key().as_ref(), serialized.as_ref())
```
DanCodedThis: Same here

DanCodedThis (approved):
This is a big pr, obviously some things (probably a lot) will be broken, but the sonner we update the better. In general LGTM

DanCodedThis:
Probably a good idea to ask Copilot to take a look here, also

rampage644 (approved):
I think we need to merge this and have follow-up PRs fixing/correcting problems

crates/utils/src/lib.rs
Vedin: Thanks! Worked for me!

Issue #513, opened 2025-04-29
S3 Tables Support


DanCodedThis:
Show S3 Tables in all handlers that show tables

DanCodedThis:
Partially or fully fixed in: https://github.com/Embucket/embucket/pull/561

camuel:
can we close it? @DanCodedThis

DanCodedThis:
@osipovartem this can be closed?

Issue #560, opened 2025-05-01
catalog: Fetch catalogs tree


osipovartem:

osipovartem:
The issue can be closed since we already fetch catalogs by sql queries implemented 
in 
https://github.com/Embucket/embucket/pull/561
https://github.com/Embucket/embucket/pull/707

PR #561, opened 2025-05-01
Fetch catalogs from session

osipovartem:
- added cache for all catalog layers
- create session: return it instead of empty response
This is the first step to get all nessessary data from session (instead of direct call to metastore)
related to #560


crates/runtime/src/http/ui/navigation_trees/handlers.rs
```diff
@@ -45,69 +45,41 @@ pub struct ApiDoc;
...
             schemas,
         });
     }
 
     Ok(Json(NavigationTreesResponse {
-        items: databases,
+        items,
         current_cursor: parameters.cursor,
-        next_cursor,
+        next_cursor: String::new(),
```
DanCodedThis: Why is there no next_cursor?

DanCodedThis (changes requested):
Pagination is a requirement from UI

DanCodedThis:
Related to #513

crates/runtime/src/execution/service.rs
```diff
@@ -25,14 +26,19 @@ use super::error::{self as ex_error, ExecutionError, ExecutionResult};
...
-    async fn create_session(&self, session_id: String) -> ExecutionResult<()>;
+    async fn create_session(&self, session_id: String) -> ExecutionResult<Arc<UserSession>>;
     async fn delete_session(&self, session_id: String) -> ExecutionResult<()>;
     async fn query(
         &self,
         session_id: &str,
         query: &str,
         query_context: QueryContext,
     ) -> ExecutionResult<(Vec<RecordBatch>, Vec<ColumnInfo>)>;
+    async fn table(
```
rampage644: I am not sure I want to extend this interface with another method

rampage644 (dismissed):
My overall feedback for the change: I am not a fan for extending `ExecutionService` interface and introducing another dependency for http module (and handlers), like `UserSession`. Yes, it already depends on it it thorugh the runtime crate, but I would like to see the movement away not towards more coupled dependency.

With that being said, I envision UI http module (handlers) depend only on `ExecutionService.query()` method and nothing else. Getting data from metastore is suboptimal and should be replaced but the change proposed doesn't materially solve the issue.

crates/runtime/src/http/ui/navigation_trees/handlers.rs
osipovartem: There is no cursor since it is not related to slateDB and our metastore

DanCodedThis (approved):
LGTM, but will this also fix #513 (at least partially)?

osipovartem:
I'll add this logic in a separate PR since it's already too big

Issue #699, opened 2025-05-05
[discussion] Shared catalogs list between sessions


osipovartem:
**🐛 Issue: Redundant and Costly Catalog Scanning on Session Initialization**

For **each new session**, we currently recreate the full catalog list, which involves scanning all internal and external catalogs to retrieve schemas and tables. This process is repeated even though the catalog structure is the same each time. Retrieving metadata from **external catalogs is especially slow**, often taking several seconds.
```rust
impl UserSession {
    pub async fn new(metastore: Arc<dyn Metastore>) -> ExecutionResult<Self> {
        ....
        catalog_list_impl.register_catalogs().await?;
        catalog_list_impl.refresh().await?;
```
This inefficiency is compounded when executing a custom logical plan. In particular, **table_references_for_statement** triggers a deep traversal of all catalogs:

```rust
for catalog in self.session.ctx.state().catalog_list().catalog_names() {
    let provider = self.session.ctx.state().catalog_list().catalog(&catalog).unwrap();
    for schema in provider.schema_names() {
        for table in provider.schema(&schema).unwrap().table_names() {
            let table_source = provider
                .schema(&schema)
                .unwrap()
                .table(&table)
                .await
                .context(super::error::DataFusionSnafu)?
                .ok_or(ExecutionError::TableProviderNotFound {
                    table_name: table.clone(),
                })?;
            let resolved = state.resolve_table_ref(TableReference::full(
                catalog.to_string(),
                schema.to_string(),
                table,
            ));
            tables.insert(resolved, provider_as_source(table_source));
        }
    }
}
```

**✅ Optimization Proposal**

To improve performance and reduce redundant work, we propose the following:

1. **Parallelize Table Fetching**

Fetch table metadata asynchronously using tokio::spawn or equivalent to avoid sequential await bottlenecks.

2. **Persist Catalog Tree in Database**
Maintain the catalog/schema/table structure in a database:
- Load it when a session starts.
- Update the cache and DB only on metadata changes:
  - CREATE TABLE, DROP TABLE, ALTER TABLE
  - Schema creation/deletion

3. **Limit Scope in table_references_for_statement**
Avoid scanning the entire catalog list. Instead:
- Resolve only the tables actually referenced in the logical plan.
- This avoids unnecessary resolution of unrelated tables.

💥 Impact

- Reduces session initialization time.
- Speeds up first query execution, especially in systems with large external catalogs.
- Prevents redundant metadata fetching and improves scalability.

[...]

PR #707, opened 2025-05-06
Add external catalogs to catalogs tree

osipovartem:
**Schema providers cache**
- if the schema is missing in the cache then fetch it from the original catalog and add to cache
- register table in cache and pass to original SchemaProvider implementation if supported
- deregister table from cache and pass to original SchemaProvider implementation if supported
**Tables provider cache**
- add table schema to the cache if possible

**Query logical plan execution**
- **removed logic to get all tables from catalogs since we already added them during session creation**_

**UI handlers**
- Navigation tree fetched from session catalogs (**select from information schema triggers heavy tables scan, but we need only table names, this logic be updated in the next PR**)
- Table schema fetching **not changed** 
- Table data preview
  - removed metastore existence check for tables since we added external catalogs and engine should check this by himself 
  - removed StringArray downcasting since we should use array_value_to_string we wide support of data types

Related to #560 and #699


crates/runtime/src/http/ui/navigation_trees/handlers.rs
```diff
@@ -47,67 +47,51 @@ pub struct ApiDoc;
...
-        })?;
+        .context(error::SessionSnafu)?
+        .fetch_catalogs_tree();
 
-    let next_cursor = rw_databases
-        .iter()
-        .last()
-        .map_or(String::new(), |rw_object| rw_object.ident.clone());
+    let offset = parameters
+        .cursor
```
DanCodedThis: Can you make the cursor into an offset in the `models.rs` and add it in the `utoipa` schema akin to `TablePreviewDataParameters`?

DanCodedThis (approved):
Otherwise LGTM

crates/runtime/src/http/ui/navigation_trees/handlers.rs
osipovartem: Do you mean rename cursor Option String to offset Option u32 in NavigationTreesParameters?

DanCodedThis: > Do you mean rename cursor Option String to offset Option u32 in NavigationTreesParameters?

Yep (`usize` maybe or whatever the type `.take()` wants)

DanCodedThis: > Do you mean rename cursor Option String to offset Option u32 in NavigationTreesParameters?

Also please don't forget the `utoipa` schema description, it also needs to be added there akin to `fn get_table_preview_data`

DanCodedThis: @nikitastryuk the navigation trees now will have offset instead of cursor, keep in mind for the UI

crates/runtime/src/http/ui/navigation_trees/handlers.rs
```diff
@@ -47,67 +47,51 @@ pub struct ApiDoc;
...
     Query(parameters): Query<NavigationTreesParameters>,
     State(state): State<AppState>,
 ) -> NavigationTreesResult<Json<NavigationTreesResponse>> {
-    let rw_databases = state
-        .metastore
-        .iter_databases()
-        .cursor(parameters.cursor.clone())
-        .limit(parameters.limit)
-        .collect()
+    let catalogs_tree = state
```
rampage644: I have left similar feedback in the previous PR: I don't like UI dependency on custom built methods from either execution/session/query.

I envision this is strictly query.execute("SHOW TABLES IN SCHEMA") or query.execute("SELECT * FROM information_schema.tables") and nothing else.

rampage644: Why "select from information schema triggers heavy tables scan, but we need only table names, this logic be updated in the next PR" cannot be part of this PR?

osipovartem: This not possible till this issue is fixed
https://github.com/apache/datafusion/issues/15960
https://github.com/Embucket/datafusion/pull/21

osipovartem: After fix we will call query without other methods

crates/runtime/src/http/ui/navigation_trees/handlers.rs
osipovartem: what about 
```rust
pub struct NavigationTreesResponse {
    pub items: Vec<NavigationTreeDatabase>,
    pub current_cursor: Option<String>,
    pub next_cursor: String,
}

File: crates/runtime/src/http/ui/navigation_trees/handlers.rs (hunk 2 above)
Comment (osipovartem) [reply crates/runtime/src/http/ui/navigation_trees/handlers.rs]: Also this PR will be too huge for review in case of other changes

File: crates/runtime/src/http/ui/navigation_trees/handlers.rs (hunk 1 above)
Comment (DanCodedThis) [reply crates/runtime/src/http/ui/navigation_trees/handlers.rs]: > what about
> 
> ```rust
> pub struct NavigationTreesResponse {
>     pub items: Vec<NavigationTreeDatabase>,
>     pub current_cursor: Option<String>,
>     pub next_cursor: String,
> }
> ```

Since it's an offset I guess we should drop `current_cursor` and `next_cursor` just like the `TablePreviewDataResponse`

File: crates/runtime/src/http/ui/navigation_trees/handlers.rs (hunk 2 above)
Comment (rampage644) [reply crates/runtime/src/http/ui/navigation_trees/handlers.rs]: > This not possible till this issue is fixed [apache/datafusion#15960](https://github.com/apache/datafusion/issues/15960) [Embucket/datafusion#21](https://github.com/Embucket/datafusion/pull/21)

This is helpful, thanks for the context

Comment (osipovartem) [reply crates/runtime/src/http/ui/navigation_trees/handlers.rs]: Replaced with sql query
Once the above issue is fixed, we should update this code.

File: crates/runtime/src/http/ui/tests/navigation_trees.rs [hunk 3]
```diff
@@ -113,7 +113,7 @@ async fn test_ui_databases_navigation() {
     let databases_navigation: NavigationTreesResponse = res.json().await.unwrap();
     assert_eq!(4, databases_navigation.items.len());
     assert_eq!(1, databases_navigation.items.first().unwrap().schemas.len());
-    assert_eq!(0, databases_navigation.items.last().unwrap().schemas.len());
+    assert_eq!(1, databases_navigation.items.last().unwrap().schemas.len());
```
rampage644, line 116: Why is this not working?

osipovartem: We return additional information schema and multiple system tables

PR #708, opened 2025-05-06
`to_boolean` and `try_to_boolean` function

### LINKED ISSUE #602: [BUG] Snowflake Function: to_boolean
**Function:** `to_boolean`
**Category:** Conversion
**Documentation:** https://docs.snowflake.com/en/sql-reference/functions/to_boolean

**Test SQL:**
```sql
SELECT TO_BOOLEAN('False');
```

**Expected Result:**
```
FALSE
```

**Actual Result:**
```
SELECT TO_BOOLEAN('False');
```

**Stderr:**
```
│ 000200: 200: DataFusion error: Error during planning: Invalid function 'to_boolean'.                                                                        │
│ Did you mean 'bool_and'?                                                                                                                                    │
```

**Parent EPIC:** #565

ravlio:
Closes #602

### DIFF
--- a/crates/runtime/src/execution/datafusion/functions/mod.rs
+++ b/crates/runtime/src/execution/datafusion/functions/mod.rs
@@ -1,6 +1,6 @@
 use arrow_array::{
-    ArrayRef, BooleanArray, Decimal128Array, Float32Array, Float64Array, Int16Array,
-    Int32Array, Int64Array, Int8Array, UInt16Array, UInt32Array, UInt64Array, UInt8Array,
+    ArrayRef, BooleanArray, Decimal128Array, Float32Array, Float64Array, Int16Array, Int32Array,
+    Int64Array, Int8Array, UInt16Array, UInt32Array, UInt64Array, UInt8Array,
 };
 use arrow_schema::DataType;
 use datafusion::{common::Result, execution::FunctionRegistry, logical_expr::ScalarUDF};
--- a/crates/runtime/src/execution/datafusion/functions/mod.rs
+++ b/crates/runtime/src/execution/datafusion/functions/mod.rs
@@ -1,5 +1,5 @@
 use arrow_array::{
-    Array, ArrayRef, BooleanArray, Decimal128Array, Float32Array, Float64Array, Int16Array,
+    ArrayRef, BooleanArray, Decimal128Array, Float32Array, Float64Array, Int16Array,
     Int32Array, Int64Array, Int8Array, UInt16Array, UInt32Array, UInt64Array, UInt8Array,
 };
 use arrow_schema::DataType;
--- a/crates/runtime/src/execution/datafusion/functions/mod.rs
+++ b/crates/runtime/src/execution/datafusion/functions/mod.rs
@@ -88,7 +88,6 @@ macro_rules! numeric_to_boolean {
 
 #[allow(clippy::cognitive_complexity, clippy::unwrap_used)]
 pub(crate) fn array_to_boolean(arr: &ArrayRef) -> Result<BooleanArray> {
-    let arr = arr.as_ref();
     Ok(match arr.data_type() {
         DataType::Boolean => {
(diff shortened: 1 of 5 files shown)


crates/runtime/src/execution/datafusion/functions/to_boolean.rs
```diff
@@ -0,0 +1,242 @@
...
+        let arr = match args[0].clone() {
+            ColumnarValue::Array(arr) => arr,
+            ColumnarValue::Scalar(v) => v.to_array()?,
+        };
+
+        let arr = match arr.data_type() {
+            DataType::Utf8 => {
+                let arr = as_string_array(&arr);
+                let mut res = BooleanBuilder::with_capacity(arr.len());
+                let true_ = ["true", "t", "yes", "y", "on", "1"];
```
DanCodedThis: It probably can be `const` and any type of `set`

crates/runtime/src/execution/datafusion/functions/to_boolean.rs
```diff
@@ -0,0 +1,242 @@
...
+    fn signature(&self) -> &Signature {
+        &self.signature
+    }
+
+    fn return_type(&self, _arg_types: &[DataType]) -> DFResult<DataType> {
+        Ok(DataType::Boolean)
+    }
+
+    fn invoke_with_args(&self, args: ScalarFunctionArgs) -> DFResult<ColumnarValue> {
+        assert_eq!(args.args.len(), 1);
```
DanCodedThis: Is this expected to panic?

DanCodedThis (approved):
Otherwise LGTM

crates/runtime/src/execution/datafusion/functions/mod.rs
```diff
@@ -59,3 +68,36 @@ mod macros {
...
+
+pub(crate) fn array_to_boolean(arr: &ArrayRef) -> Result<BooleanArray> {
+    let arr = arr.as_ref();
+    let mut boolean_array = BooleanArray::builder(arr.len());
+    for i in 0..arr.len() {
+        if arr.is_null(i) {
+            boolean_array.append_null();
+        } else {
+            let b = match arr.data_type() {
+                DataType::Boolean => downcast_value!(arr, BooleanArray).value(i),
```
DanCodedThis: Why are we downcasting the whole array multiple times (per element)?

crates/runtime/src/execution/datafusion/functions/to_boolean.rs
ravlio: it is, but signature should not pass arguments of the wrong types, it is additional check. I took this assert from other DF functions. I'll remove it though

crates/runtime/src/execution/datafusion/functions/to_boolean.rs
ravlio: nice catch

crates/runtime/src/execution/datafusion/functions/mod.rs
ravlio: idk to be honest how could I do this :) Thanks for finding the bottleneck, will rewrite

crates/runtime/src/execution/datafusion/functions/to_boolean.rs
DanCodedThis: > it is, but signature should not pass arguments of the wrong types, it is additional check. I took this assert from other DF functions. I'll remove it though

However, my experience with DF signature, is it sometimes doesn't work :) (`convert_timezone`) However it was with similar signature types, but different argument number length. Anyway, here it shouldn't be a problem I guess

ravlio:
@DanCodedThis kindly ask you to re-check. I made a batch processing macro `numeric_to_boolean`

DanCodedThis:
BTW, about `set`, `phf` or `lazy_static` maybe good here for true compile time sets, but probably overkill

DanCodedThis (approved):
LGTM

Issue #711, opened 2025-05-06
api: worksheets deserialize bug


YaroslavLitvinov:
reported by @nikitastryuk 

When some worksheet have queries, then getting worksheets list leads to a runtime error from handler :
GET /ui/worksheets
Error 400:
```
Get worksheets error: Error getting worksheets: Deserialize error: missing field `id` at line 1 column 2
```

PR #712, opened 2025-05-06
Issues/516 basic/bearer auth on backend

### LINKED ISSUE #516: api: UI Basic auth
## Implement basic authentication, with JWT encoded tokens: refresh_token, access_token.

refresh_token is long-lived token: 7 days, securely stored in browser cookies,
access_token is short-lived token: 15 minutes, client responsible keeping it in secret.

### Add authentication handlers:
* POST /auth/login, 
  Request body contains json { "username"="...", "password"="..." }, as application/json 
  Response:
  - Header Set-Cookie refresh_token=..;  HttpOnly; Secure; SameSite=Strict; 
    ~Set Domain attribute supporting cookies across sub-domains (no actual requirements for this)~
  - ~Header Set-Cookie access_token=; Secure; SameSite=Strict; (for smooth authentication support in browser)~
  - response body in camelCase as application/json :
  ```
  {
    "accessToken": "..."
    "tokenType": "Bearer",
    "expiresIn": 900,
  } 
  ```
* POST /auth/refresh,
  Request has no body, only cookie with refresh_token expected
  Response: 
  - Set-Cookie header sets new refresh_token 
  - Body with new access token (the same response as in login)
* POST /auth/logout,
  Request has no body, no matter cookies set or not.
  Response:
  - Sets empty refresh_token cookie,  (access_token can't be revoked and client is responsible for removing access_token)

### Rest of existing handlers, accessing authenticated resource
  Requests should have Authorization header, like this:
  Authorization: Bearer {access_token}

  Response:
  In case if access_token is valid return response as usual
  In case if access_token is invalid: bad token, expired token - return status=401.

### Unauthorized www-authenticate
When 401 unauthorized response returned, it includes
 * www-authenticate header with auth error

Feel free to comment: 

cc @YaroslavLitvinov @rampage644 @nikitastryuk

YaroslavLitvinov:
Closes #516 #711 

* Adds simplified authentication: it only allows to login using credentials provided in cli arguments / env vars.
  /auth/login, /auth/logout, /auth/refresh handlers
  For UI, metastore handlers it xpects Authorization header within every API request in order to be authenticated
* Added auth support covers only: ui, metastore endpoints (actually metastore handlers exposed and not used, but could).
* Fixes critical bug in worksheets described in issue #711 
  - It makes breaking change to queries earlier created  from worksheets forces them to disappear from worksheets.
  - Drop stored data if you already encounter deserialization error.


crates/runtime/src/http/router.rs
```diff
@@ -34,12 +35,16 @@ pub fn create_app(state: AppState) -> Router {
     let ui_router = create_ui_router();
     let dbt_router = create_dbt_router();
     let iceberg_catalog = create_iceberg_router();
+    let auth_router = create_auth_router();
 
     Router::new()
         .merge(dbt_router)
         .merge(metastore_router)
-        .nest("/ui", ui_router)
+        .merge(ui_router)
```
DanCodedThis, line 42: wouldn't this cause that UI would be exposed without nesting in `/ui...`?

crates/runtime/src/http/mod.rs
```diff
@@ -41,12 +42,14 @@ pub mod utils;
 mod tests;
 
 use super::http::config::WebConfig;
+use crate::config::AuthConfig;
 
 #[allow(clippy::needless_pass_by_value)]
 pub fn make_app(
-    metastore: Arc<dyn Metastore>,
-    history_store: Arc<dyn WorksheetsStore>,
```
DanCodedThis, line 48: Why was this changed?

crates/runtime/src/http/auth/router.rs
```diff
@@ -0,0 +1,12 @@
+use crate::http::state::AppState;
+use axum::routing::post;
+use axum::Router;
+
+use super::handlers::{login, logout, refresh_access_token};
+
+pub fn create_router() -> Router<AppState> {
+    Router::new()
```
DanCodedThis, line 8: Nitpick: Shouldn't auth be a part of UI?

crates/runtime/src/http/ui/router.rs
```diff
@@ -140,5 +140,7 @@ pub fn create_router() -> Router<AppState> {
             axum::http::header::AUTHORIZATION,
         ]))
         .layer(axum::middleware::from_fn(add_request_metadata))
-        .layer(DefaultBodyLimit::disable())
+        .layer(DefaultBodyLimit::disable());
+
+    Router::new().nest("/ui", ui_router)
```
DanCodedThis, line 145: I think it's better to nest in main router file, so that it would be seen in one place

crates/runtime/src/http/auth/handlers.rs
```diff
@@ -0,0 +1,225 @@
...
+    T: Serialize,
+{
+    encode(
+        &Header::default(),
+        &claims,
+        &EncodingKey::from_secret(jwt_secret.as_bytes()),
+    )
+}
+
+const fn assert_jwt_secret(jwt_secret: &str) -> AuthResult<()> {
```
DanCodedThis: Nitpick: I would rename this, specifically the assert part, since assert means panic, means stop everything, means service is dead

DanCodedThis (approved):
Other than some comments, LGTM

crates/runtime/src/http/router.rs
YaroslavLitvinov: It doesn't, as it has nested ui router. 
``` Router::new().nest("/ui", ui_router) ```

File: crates/runtime/src/http/mod.rs (hunk 2 above)
Comment (YaroslavLitvinov) [reply crates/runtime/src/http/mod.rs:48]: I consider this change as a code refactoring.

In `make_app` you pass DB itself and not just store interfaces.
* We have 2 separate databases, so 2 arguments.
* App should get DB and not just store trait, traits should appear in more high level.
* Also inside of make_app it's native to have access to DB interface, and not just to store interfaces.
* Also you may need to add more store interfaces, and its better to do in inside of make_app.

File: crates/runtime/src/http/auth/router.rs (hunk 3 above)
Comment (YaroslavLitvinov) [reply crates/runtime/src/http/auth/router.rs:8]: I treat authentication as a service auth. Except of /ui endpoints we have metastore, catalogs and it all should be authenticated before access.

File: crates/runtime/src/http/auth/handlers.rs (hunk 5 above)
Comment (YaroslavLitvinov) [reply crates/runtime/src/http/auth/handlers.rs]: I have no strong opinion on this. But actually that is close to assert - in case if secret is empty that part of code should deny to function at all!

**Comment by YaroslavLitvinov** [comment]
Copilot Review:

1. Configuration Changes (.cargo/config.toml, Cargo.toml):
Changes:

    Added new alias commands in .cargo/config.toml:
        dev and auth_demo aliases for running the application with specific configurations.
    Added jsonwebtoken and updated dependencies in Cargo.toml.

Review:

    Positive: Introducing aliases (e.g., dev and auth_demo) improves developer productivity, making it easier to run the application in specific modes.
    Consideration: Ensure that the added dependencies like jsonwebtoken are necessary and don't introduce unnecessary bloat. Check if the added version (9.3.1) has known vulnerabilities.

2. Enhancements in QueryRecordReference Struct (worksheet_query_ref.rs):
Changes:

    Added Default trait to QueryRecordReference.
    Changed #[serde(skip_serializing)] to #[serde(skip)].
    Updated key formatting logic in get_key to use a new /rws prefix.

Review:

    Positive: The changes align with Rust best practices. The addition of the Default trait increases usability.
    Consideration: Verify that changing the key prefix (/ws → /rws) doesn't break existing functionality or data compatibility.

3. Introduction of AuthConfig (config.rs):
Changes:

    Added a new AuthConfig struct for managing authentication-related configurations (e.g., JWT secret, demo user credentials).
    Provided methods to manage and retrieve configuration values.

Review:

    Positive: Centralizing authentication configuration in a dedicated struct reduces clutter and improves modularity.
    Consideration: Since sensitive data like jwt_secret is handled, ensure it is securely managed (e.g., avoid logging it).

4. Middleware and Routing Updates (e.g., router.rs, mod.rs, state.rs):
Changes:

    Added authentication middleware (require_auth).
    Introduced new routes (e.g., auth_router).
    Updated state management to include AuthConfig.

Review:

    Positive: Adding authentication middleware ensures that routes are protected. Also, separating authentication configuration (AuthConfig) from other state components is a good design choice.
    Consideration: Ensure that the middleware has comprehensive tests to validate its functionality.

5. Test Updates and Additions (tests/auth.rs, queries.rs):
Changes:

    Added tests for JWT token validation (e.g., test_jwt_token_valid, test_jwt_token_expired).
    Updated query test to include a scenario for worksheets.

Review:

    Positive: Including tests for authentication and other features ensures that the changes are robust and backward-compatible.
    Consideration: Some tests have hardcoded values (e.g., JWT secrets). Consider using mocks or environment variables for better flexibility.

6. Error Handling Improvements (utils/src/lib.rs, scan_iterator.rs):
Changes:

    Enhanced error messages in deserialization to include the key and data for better debugging.
    Improved cursor handling in IterableCursor and IterableEntity.

Review:

    Positive: The added context in errors (e.g., key and data) significantly improves the debugging experience.
    Consideration: Ensure these changes don't expose sensitive data in production logs.

7. Dependency Updates (runtime/Cargo.toml):
Changes:

    Added axum-extra for features like cookie and typed-header.

Review:

    Positive: The added library (axum-extra) aligns with the usage of the Axum framework and can simplify HTTP header and cookie management.
    Consideration: Ensure the added features are used effectively and don't unnecessarily increase the binary size.

8. Minor Refactoring:
Examples:

    Renamed methods for clarity (e.g., http_req_with_headers).
    Improved test server setup by allowing demo authentication credentials.

Review:

    Positive: These changes improve code clarity and maintainability.
    Consideration: Ensure backward compatibility if these methods are used in external systems or libraries.

Summary of Review:

    Strengths:
        The changes introduce better modularity (e.g., AuthConfig), improved error handling, and developer-friendly features (e.g., aliases, middleware).
        Tests are updated to reflect the new functionality, ensuring robust coverage.

    Concerns:
        Any changes to key formats (/ws → /rws) or serialization/deserialization logic should be carefully tested to avoid breaking existing systems.
        Sensitive data handling (e.g., jwt_secret) should be reviewed to ensure security best practices are followed.

    Suggestions:
        Add documentation for the new features (e.g., AuthConfig, new routes) to help developers understand and use them effectively.
        Verify performance implications of the new middleware and added dependencies.

File: crates/runtime/src/http/auth/handlers.rs (hunk 5 above)
Comment (DanCodedThis) [reply crates/runtime/src/http/auth/handlers.rs]: > I have no strong opinion on this. But actually that is close to assert - in case if secret is empty that part of code should deny to function at all!

If any other name comes to mind, please rename, if not, this works also

File: crates/runtime/src/http/mod.rs (hunk 2 above)
Comment (DanCodedThis) [reply crates/runtime/src/http/mod.rs:48]: > I consider this change as a code refactoring.
> 
> In `make_app` you pass DB itself and not just store interfaces.
> 
> * We have 2 separate databases, so 2 arguments.
> * App should get DB and not just store trait, traits should appear in more high level.
> * Also inside of make_app it's native to have access to DB interface, and not just to store interfaces.
> * Also you may need to add more store interfaces, and its better to do in inside of make_app.

I asked this, since when I was working on the `ExecutionService` trait and `RecordingExecutionService`, @rampage644 was adamant about using interfaces. But perhaps here it's not as important.

File: crates/runtime/src/http/ui/router.rs (hunk 4 above)
Comment (YaroslavLitvinov) [reply crates/runtime/src/http/ui/router.rs:145]: My intentions were to make it more consistent with metastore, dbt routers. Also I do think we might create an issue for code refactoring that covers specific things like this, and then make it consistent in one or another way.

File: crates/runtime/src/http/auth/router.rs (hunk 3 above)
Comment (DanCodedThis) [reply crates/runtime/src/http/auth/router.rs:8]: > I treat authentication as a service auth. Except of /ui endpoints we have metastore, catalogs and it all should be authenticated before access.

Didn't @rampage644 say that auth is for ui only?

File: crates/runtime/src/http/auth/handlers.rs (hunk 5 above)
Comment (rampage644) [reply crates/runtime/src/http/auth/handlers.rs]: require_jwt_secret_valid, require_jwt_secret_valid, etc

File: crates/runtime/src/http/auth/router.rs (hunk 3 above)
Comment (rampage644) [reply crates/runtime/src/http/auth/router.rs:8]: It doesn't mean the auth is going be the same for every other API we have, `query` API has its own authentication endpoints, so does Iceberg REST API

File: crates/runtime/src/http/router.rs [hunk 6]
```diff
@@ -34,12 +35,16 @@ pub fn create_app(state: AppState) -> Router {
...
         .nest("/catalog", iceberg_catalog)
+        // middleware wraping all routes above ^^
+        .layer(middleware::from_fn_with_state(state.clone(), require_auth))
```
rampage644, line 44: What happens with other connector tools? Will `dbt seed & dbt run` work with this change?

crates/runtime/src/http/mod.rs
YaroslavLitvinov: I'm pretty sure that we don't need to expose high level interfaces too early. Think about using low-level DB functions like  DB - open, close, snapshots functions, etc... 

Do you think those such operations be done outside of make_app ?

crates/runtime/src/http/router.rs
YaroslavLitvinov: Thanks for asking, it wasn't supported yet. I would cover "dbt" endpoint as a separate issue.

crates/runtime/src/http/mod.rs
rampage644: as @DanCodedThis I would prefer trait object dependencies over specific type (`accept generic return specific` approach). 

I would avoid arguing about that specific function as in my mind main.rs/run_binary/make_app split is confusing and is pending some refactoring already.

What is way more important here from my perspective - is the definition of the `State` struct, since this is how we share the state and thus dependencies across bigger application. We still have trait objects there.

crates/runtime/src/http/auth/router.rs
YaroslavLitvinov: @rampage644 
Ok, so since iceberg auth will be added separately, and as this PR doesn't cover dbt, I will update middleware so it will not impact catalog as well as dbt.

Should it authenticate metastore endpoints as well as ui ?

rampage644: My understanding we add auth service and handlers only for ui module

crates/runtime/src/http/ui/router.rs
rampage644: Agree with both - we have refactoring pending + we should merge/nest/connect routers on the top level

crates/runtime/src/http/router.rs
rampage644: Please let's not cover any endpoints except of ui related with this auth change.

rampage644 (approved):
LGTM once made sure this change doesn't affect any endpoints except ui related

PR #713, opened 2025-05-06
Restore runtime insta tests

rampage644:
`runtime` crate insta tests were effectively unused, simplify and restore them so they can be used for holistic query processing testing:
 * Omit comparing AST and logical plan for statements (current API doesn't exposes those)
 * Remove majority of tests (expected failures, duplicative tests)


crates/runtime/src/tests/utils.rs
```diff
@@ -72,28 +72,20 @@ pub mod macros {
...
-                        let plan = plan.map(|plan| plan.to_string().split("\n").map(|s| s.to_string()).collect::<Vec<String>>());
+                        let df =  match res {
+                            Ok(record_batches) => {
+                                Ok(datafusion::arrow::util::pretty::pretty_format_batches(&record_batches).unwrap().to_string())
+                            },
+                            Err(e) => Err(format!("Error: {e}"))
+                        };
                         let df = df.map(|df| df.split("\n").map(|s| s.to_string()).collect::<Vec<String>>());
-                        insta::assert_debug_snapshot!((statement, plan, df));
+                        insta::assert_debug_snapshot!((df));
```
YaroslavLitvinov, line 88: Why we don't use assert_json_snapshot instead of assert_debug_snapshot? Json format vs table text output?

YaroslavLitvinov (approved):

crates/runtime/src/tests/utils.rs
rampage644: I don't know why originally here was used debug instead of json assertion, but if I had to guess that would be "json and yaml assertions require extra dependencies (one has to enable that features in insta), that organically leads to worse compile times, worse performance, etc".

PR #716, opened 2025-05-07
`to_time`, `try_to_time` scalar functions

### LINKED ISSUE #623: [BUG] Missing function: to_time
**Function:** `to_time`
**Category:** Conversion
**Documentation:** https://docs.snowflake.com/en/sql-reference/functions/to_time

**Test SQL:**
```sql
SELECT TO_TIME('14:30:00');
```

**Expected Result:**
```
N/A
```

**Actual Result:**
```
SELECT TO_TIME('14:30:00');
```

**Stderr:**
```
│ 000200: 200: DataFusion error: Error during planning: Invalid function 'to_time'.                                                                           │
│ Did you mean 'to_date'?                                                                                                                                     │
```

**Parent EPIC:** #565

ravlio:
Closes #623

### DIFF
--- a/crates/runtime/src/execution/datafusion/functions/to_time.rs
+++ b/crates/runtime/src/execution/datafusion/functions/to_time.rs
@@ -1,8 +1,18 @@
+use arrow::compute::cast;
+use arrow::compute::kernels::cast_utils::Parser;
+use arrow_array::builder::{Time64NanosecondBuilder, TimestampNanosecondBuilder};
+use arrow_array::types::{Time64NanosecondType, TimestampNanosecondType};
+use arrow_array::{Array, ArrayRef, StringArray, Time64NanosecondArray, TimestampNanosecondArray};
 use arrow_schema::{DataType, TimeUnit};
+use chrono::{NaiveTime, ParseError, Timelike};
 use datafusion::error::Result as DFResult;
 use datafusion::logical_expr::ColumnarValue;
-use datafusion_expr::{ScalarFunctionArgs, ScalarUDFImpl, Signature, TypeSignature};
+use datafusion_common::{exec_err, plan_err, ScalarValue};
+use datafusion_expr::{ScalarFunctionArgs, ScalarUDFImpl, Signature, TypeSignature, Volatility};
 use std::any::Any;
+use std::collections::HashMap;
+use std::num::ParseIntError;
+use std::sync::Arc;
 
 #[derive(Debug)]
 pub struct ToTimeFunc {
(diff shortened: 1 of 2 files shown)


crates/runtime/src/execution/datafusion/functions/to_time.rs
```diff
@@ -0,0 +1,380 @@
...
+    fn aliases(&self) -> &[String] {
+        &self.aliases
+    }
+
+    fn signature(&self) -> &Signature {
+        &self.signature
+    }
+
+    fn return_type(&self, _arg_types: &[DataType]) -> DFResult<DataType> {
+        Ok(DataType::Time64(TimeUnit::Nanosecond))
```
DanCodedThis, line 107: I think `Microseconds` more appropriate since DF likes them, and I think we use them in our codebase also in UDFs particularly, maybe something changed?

DanCodedThis: or Snowflake/dbt not DF

crates/runtime/src/execution/datafusion/functions/to_time.rs
```diff
@@ -0,0 +1,380 @@
...
+            }
+            DataType::Timestamp(
+                TimeUnit::Second
+                | TimeUnit::Millisecond
+                | TimeUnit::Microsecond
+                | TimeUnit::Nanosecond,
+                _,
+            ) => {
+                let arr = cast(arr, &DataType::Timestamp(TimeUnit::Nanosecond, None))?;
+                let arr = arr
```
DanCodedThis, line 249: Why the double cast?

DanCodedThis (approved):
Otherwise LGTM

crates/runtime/src/execution/datafusion/functions/to_time.rs
```diff
@@ -0,0 +1,380 @@
+use arrow::compute::cast;
```
rampage644: From now on, let's please use `datafusion::` dependencies instead of arrow.
`use arrow::compute::cast` -> `use datafusion:arrow:compute::cast` and `use arrow_array` -> `use datafusion::arrow::array`

crates/runtime/src/execution/datafusion/functions/to_time.rs
ravlio: @rampage644 what do you think?

crates/runtime/src/execution/datafusion/functions/to_time.rs
```diff
@@ -0,0 +1,380 @@
...
+// AM , PM - Ante meridiem (AM) / post meridiem (PM). Use this only with HH12 (not with HH24).
+// MI - Two digits for minute (00 through 59).
+// SS - Two digits for second (00 through 59).
+// FF[0-9] - Fractional seconds with precision 0 (seconds) to 9 (nanoseconds), e.g. FF, FF0, FF3, FF9. Specifying FF is equivalent to FF9 (nanoseconds).
+// Example: TO_TIME('11.15.00', 'HH24.MI.SS')
+#[derive(Debug)]
+pub struct ToTimeFunc {
+    signature: Signature,
+    aliases: Vec<String>,
+    try_: bool,
```
rampage644: r#try?

ravlio: or `is_try`. Let's make it `r#try` then.

rampage644: `is_try` also good

crates/runtime/src/execution/datafusion/functions/to_time.rs
rampage644: For the context, we used Microseconds in timestamps since there is no support for nanoseconds accuracy in Iceberg and we have to coerce to microseconds.

I am not sure though if this warrants "always avoid nanoseconds" approach everywhere. One day the Iceberg spec will support nanoseconds and then we'll have to update that. 

@DanCodedThis can you elaborate on "datafusion doesn't like nanoseconds"?

rampage644: I would support as much as possible in UDFs, and if there are problems with timestamp accuracy, we better of deal with this somewhere else, in a single place (code flow).

DanCodedThis: > For the context, we used Microseconds in timestamps since there is no support for nanoseconds accuracy in Iceberg and we have to coerce to microseconds.
> 
> I am not sure though if this warrants "always avoid nanoseconds" approach everywhere. One day the Iceberg spec will support nanoseconds and then we'll have to update that.
> 
> @DanCodedThis can you elaborate on "datafusion doesn't like nanoseconds"?

I think I remember now, you explained exactly what I wanted to convey, it was the Iceberg format. Sorry for the misinfo

rampage644 (approved):

PR #718, opened 2025-05-07
Add information schema provider

### LINKED ISSUE #701: catalog: information schema per catalog
Related to https://github.com/apache/datafusion/issues/15960

osipovartem:
Added our own implementation for information schema since
- snowflake information schema [contains more ](https://docs.snowflake.com/en/sql-reference/info-schema)tables that we have in DF repo
- most of tables have different list of columns
- it will be much easier for us to extend/change information schema tables to support Snowflake
- added 3 additional system tables from DF implementation for registered functions, settings and params
- most of tables should work per catalog, full list can be found here https://docs.snowflake.com/en/sql-reference/info-schema
- **added i[nformation_schema.databases](https://docs.snowflake.com/en/sql-reference/info-schema/databases) to get list of databases (catalogs) to fetch navigation tree** 
Screenshot 2025-05-07 at 19 33 01

Closes #701

### DIFF
--- a/crates/runtime/src/execution/catalog/information_schema/columns.rs
+++ b/crates/runtime/src/execution/catalog/information_schema/columns.rs
@@ -1,0 +1,235 @@
+//! [`InformationSchemaColumns`] that implements the SQL [Information Schema Column] for Snowflake.
+//!
+//! [Information Schema Column]: https://docs.snowflake.com/en/sql-reference/info-schema/columns
+
+use crate::execution::catalog::information_schema::config::InformationSchemaConfig;
+use arrow::{
+    array::StringBuilder,
+    datatypes::{DataType, Field, Schema, SchemaRef},
+    record_batch::RecordBatch,
+};
+use arrow_array::builder::UInt64Builder;
+use arrow_schema::ArrowError;
+use datafusion::execution::TaskContext;
+use datafusion_common::DataFusionError;
+use datafusion_physical_plan::stream::RecordBatchStreamAdapter;
+use datafusion_physical_plan::streaming::PartitionStream;
+use datafusion_physical_plan::SendableRecordBatchStream;
+use std::fmt::Debug;
+use std::sync::Arc;
+use DataType::{
+    Binary, Decimal128, Float16, Float32, Float64, Int16, Int32, Int8, LargeBinary, LargeUtf8,
+    UInt16, UInt32, UInt64, UInt8, Utf8,
+};
+
+#[derive(Debug)]
+pub struct InformationSchemaColumns {
+    schema: SchemaRef,
+    config: InformationSchemaConfig,
+}
+
+impl InformationSchemaColumns {
+    pub fn new(config: InformationSchemaConfig) -> Self {
+        let schema = Arc::new(Schema::new(vec![
+            Field::new("table_catalog", Utf8, false),
+            Field::new("table_schema", Utf8, false),
+            Field::new("table_name", Utf8, false),
+            Field::new("column_name", Utf8, false),
+            Field::new("ordinal_position", UInt64, false),
+            Field::new("column_default", Utf8, true),
+            Field::new("is_nullable", Utf8, false),
+            Field::new("data_type", Utf8, false),
+            Field::new("character_maximum_length", UInt64, true),
+            Field::new("character_octet_length", UInt64, true),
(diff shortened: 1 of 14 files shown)


crates/runtime/src/execution/catalog/information_schema/columns.rs
```diff
@@ -0,0 +1,235 @@
...
+        &mut self,
+        catalog_name: &str,
+        schema_name: &str,
+        table_name: &str,
+        field_position: usize,
+        field: &Field,
+    ) {
+        // Note: append_value is actually infallable.
+        self.catalog_names.append_value(catalog_name);
+        self.schema_names.append_value(schema_name);
```
DanCodedThis, line 133: Are we appending to the same `String`? Is this expected?

crates/runtime/src/execution/catalog/information_schema/schemata.rs
```diff
@@ -0,0 +1,119 @@
...
+    default_character_set_catalog: StringBuilder,
+    default_character_set_schema: StringBuilder,
+    default_character_set_name: StringBuilder,
+    sql_path: StringBuilder,
+}
+
+impl InformationSchemataBuilder {
+    pub(crate) fn add_schemata(
+        &mut self,
+        catalog_name: &str,
```
DanCodedThis, line 68: `impl AsRef<str>` also here?

crates/runtime/src/http/ui/navigation_trees/handlers.rs
```diff
@@ -54,34 +54,50 @@ pub async fn get_navigation_trees(
...
+        for i in 0..db_batch.num_rows() {
+            let catalog = catalog_col.value(i);
 
-            catalogs_tree
-                .entry(catalog)
-                .or_default()
-                .entry(schema)
-                .or_default()
-                .push(name);
+            let query = format!(
```
DanCodedThis, line 75: Why do we need to run double query? (Just curious)

DanCodedThis (approved):
LGTM, outside the appending behavior (highly likely I don't know something)

crates/runtime/src/execution/query.rs
```diff
@@ -1143,6 +1148,35 @@ impl UserQuery {
...
+
+    pub fn schema_for_ref(
+        &self,
+        table_ref: impl Into<TableReference>,
+    ) -> datafusion_common::Result<Arc<dyn SchemaProvider>> {
+        let resolved_ref = self.session.ctx.state().resolve_table_ref(table_ref);
+        if self.session.ctx.state().config().information_schema()
+            && *resolved_ref.schema == *INFORMATION_SCHEMA
+        {
+            return Ok(Arc::new(InformationSchemaProvider::new(
```
rampage644, line 1160: I presume this is how we wire up the information schema into datafusion internals?

crates/runtime/src/execution/catalog/information_schema/columns.rs
osipovartem: it's a StringBuilder, any append operation means append new row to column

crates/runtime/src/http/ui/navigation_trees/handlers.rs
rampage644: yeah - curios to see if these two can combined into a single query with perhaps a join?


Something like
```sql
select * from information_schema.databases join information_schema.tables on d.name = t.database_name
```

crates/runtime/src/execution/catalog/information_schema/schemata.rs
osipovartem: impl AsRef<str> is already &str

rampage644 (approved):
overall lgtm

crates/runtime/src/http/ui/navigation_trees/handlers.rs
osipovartem: Each catalog contains only his schema, tables doesn't know anything about other catalogs.
So we need to get all catalogs (databases) and then for each catalog get the list of schemas and tables

crates/runtime/src/execution/query.rs
osipovartem: yes, DF has the same mechanism, actually same updated func with our own schemaProvider for InformationSchema

PR #724, opened 2025-05-08
Utoipa Spec Fixes for tables & auth

### LINKED ISSUE #725: api: UI No `Utoipa` spec for auth
- Spec: `/auth/login` POST
- Spec: `/auth/refresh` POST
- Spec: `/auth/logout` POST

DanCodedThis:
Closes #725

Also includes a minor fix for tables spec

### DIFF
--- a/crates/runtime/src/http/auth/handlers.rs
+++ b/crates/runtime/src/http/auth/handlers.rs
@@ -1,3 +1,4 @@
+use crate::http::auth::models::RefreshTokenResponse;
 use std::collections::HashMap;
 
 use super::error::CreateJwtSnafu;
@@ -19,6 +20,7 @@ use snafu::ResultExt;
 use time::Duration;
 use tower_sessions::cookie::{Cookie, SameSite};
 use tracing;
+use utoipa::OpenApi;
 
 pub const REFRESH_TOKEN_EXPIRATION_HOURS: u32 = 24 * 7;
 pub const ACCESS_TOKEN_EXPIRATION_SECONDS: u32 = 15 * 60;
@@ -128,6 +130,29 @@ fn set_cookies(headers: &mut HeaderMap, refresh_token: &str) -> AuthResult<()> {
 
     Ok(())
 }
+
+#[derive(OpenApi)]
+#[openapi(
+    paths(
+        login,
+        refresh_access_token,
+        logout,
+    ),
+    components(
+        schemas(
+            LoginPayload,
+            AuthResponse,
+            RefreshTokenResponse,
+            Claims,
+            ErrorResponse,
+        )
+    ),
+    tags(
+        (name = "auth", description = "Authentication endpoints")
+    )
+)]
+pub struct ApiDoc;
+
 #[utoipa::path(
     post,
     path = "/auth/login",
@@ -227,7 +252,6 @@ pub async fn refresh_access_token(
     tags = ["auth"],
     responses(
         (status = 200, description = "Successful Response", body = AuthResponse),
-        (status = 401, description = "Unauthorized", body = ErrorResponse),
         (status = 500, description = "Internal server error", body = ErrorResponse),
     )
 )]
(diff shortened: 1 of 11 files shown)


crates/runtime/src/http/auth/handlers.rs
```diff
@@ -163,6 +175,18 @@ pub async fn login(
...
 }
 
+#[utoipa::path(
+    post,
+    path = "/auth/refresh",
+    operation_id = "refresh",
+    tags = ["auth"],
+    responses(
+        (status = 200, description = "Successful Response", body = AuthResponse),
+        (status = 401, description = "Unauthorized", body = ErrorResponse),
```
YaroslavLitvinov, line 214: OAS of Unauthorized response should include www-authenticate header

YaroslavLitvinov:
Now, every handler now can return Unauthorized status code with www-authenticate header in response, and we need to update spec for every handler too. Better doing it in some programmatic way if possible in opposite to adding utoipa macroses manually into multiple places in code.

DanCodedThis:
> Now, every handler now can return Unauthorized status code with www-authenticate header in response, and we need to update spec for every handler too. Better doing it in some programmatic way if possible in opposite to adding utoipa macroses manually into multiple places in code.

Could've used RustRover's builtin smart find refactor for the directory of ui, but already did it manually

nikitastryuk (approved):

crates/runtime/src/http/auth/handlers.rs
```diff
@@ -128,6 +130,29 @@ fn set_cookies(headers: &mut HeaderMap, refresh_token: &str) -> AuthResult<()> {
...
+        login,
+        refresh_access_token,
+        logout,
+    ),
+    components(
+        schemas(
+            LoginPayload,
+            AuthResponse,
+            RefreshTokenResponse,
+            Claims,
```
YaroslavLitvinov: Claims is not intended to be used in OAS, we don't return it.

crates/runtime/src/http/auth/models.rs
```diff
@@ -35,7 +35,7 @@ pub struct LoginPayload {
     pub password: String,
 }
 
-#[derive(Serialize, Deserialize)]
+#[derive(Serialize, Deserialize, Clone, PartialEq, Eq, ToSchema)]
```
YaroslavLitvinov: remove Clone, PartialEq, Eq, ToSchema from Claims

YaroslavLitvinov (changes requested):

crates/runtime/src/http/auth/handlers.rs
DanCodedThis: > OAS of Unauthorized response should include www-authenticate header

@YaroslavLitvinov like this? (did some changes, plz recheck)

crates/runtime/src/http/auth/models.rs
YaroslavLitvinov: From this:
```
[derive(Serialize, Deserialize, Clone, PartialEq, Eq, ToSchema)]
```

to this::
```
[derive(Serialize, Deserialize)]
```

crates/runtime/src/http/auth/handlers.rs
YaroslavLitvinov: ~Claims still in spec, no it removed.~ No Claims - ok. Probably something didn't refresh

crates/runtime/src/http/auth/handlers.rs
```diff
@@ -196,6 +249,17 @@ pub async fn refresh_access_token(
     }
 }
 
+#[utoipa::path(
+    post,
+    path = "/auth/logout",
+    operation_id = "logout",
+    tags = ["auth"],
+    responses(
+        (status = 200, description = "Successful Response", body = AuthResponse),
```
YaroslavLitvinov: logout doesn't return AuthResponse, just empty

YaroslavLitvinov (changes requested):

YaroslavLitvinov (approved):

Issue #719, opened 2025-05-07
[Discussion] Crates separation proposal


rampage644:
I propose we split the crates with an idea of how it was previously before we merged most of them into single `runtime`.

Problems I would like to address primarily are compilation and thus developer velocity: currently it takes ~30s (at least) on my machine (M1 Max) to iterate on any change within runtime crate. Project codebase grows pretty quickly and compilation times gets worse very quickly.

Another problem to address is clearer dependency split. Example: nothing should depend on UI module and currently (due to Rust allowing cycle dependencies in a crate) there are suboptimal dependencies already. I would like to avoid that and keep dependency graph simple.

With that being said, here is the proposal (pseudo naming is used, feel free to suggest better naming):

- Have single "fat" `main` where all dependencies are constructed, set up and wired up together. This is the only crate/module that will have **all** other dependencies (with possible conditional compilation flags, i.e. "compile with ui disabled", etc)
- `embucket-executor` provides `ExecutionService` trait and `CoreExecutor` implementation, depends on `Metastore`
- `embucket-metastore` provides `Metastore` trait and `SlatedbMetastore implementation, leaf dependency (doesn't depend on other domain services and entities)
- `embucket-http-ui` provides axum `Router`, depends on possibly all/many services, serves UI
- `embucket-http-v1` provides axum `Router`, depends on possibly all/many services, serves snowflake v1 REST API
- `embucket-http-iceberg` provides axum `Router`, depends on possibly all/many services, serves Iceberg REST API
- `embucket-history`, provides `WorksheetsStore` trait and `SlateDBWorksheetsStore` + `RecordingExecutionService` implementations, depends on `dyn ExecutionService`

There are other crates I'm less confident about (mostly cross application dependencies like iceberg spec, utils and such)
- `embucket-datafusion-functions`: does it make sense to also separate from original crate?
- `embucket-datafusion-catalog`: same, not sure 
- `embucket-utils`: historically it provided slatedb related utility functions that were reused across entities storing data in slatedb
- anything else I forgot

rampage644:
@Vedin @ravlio @osipovartem @DanCodedThis @YaroslavLitvinov

Vedin:
I'd say it absolutely make sense to separate functions and catalog. A lot of work happens there so it makes no sense to rebuild all runtime every time.

DanCodedThis:
[...]

PR #744, opened 2025-05-11
Separating crates

rampage644:
This splits single `runtime` crate into several, some ideas and invariants:
 * Prefer flat crate and directory structure 
 * Single "fat" binary crate (handles all the wiring and connecting dependencies)
 * Switched to Rust 2024 edition (seem to be default for new crates for `cargo init`)

Some intermediate results
 * My single use case was "Update SQL query test in executor/tests/query.rs, run test" time improved 3x-4x (30s -> 8s). I also believe it could be even further improved.
 
Upcoming (follow up) PRs
 * Refactor error handling project-wise
 * ~~Restore tests that weren't migrated (Only `test_parallel_queries` left out)~~
 * Refactor Recording service

Open questions
 * Crate for **just** traits (interfaces) such that actual implementation stays separate. This PR effort is still not great from incremental compilation perspective. Example, `api-*-rest` depend on `ExecutorService` now *AND* `CoreExecutor` (real type that implements the trait) and thus every change in CoreExecutor (not the interface!) triggers dependent crates recompilation.
 * tbd
 
Related #719


rampage644:
Rebased on recent main

camuel (approved):

PR #759, opened 2025-05-12
Expired session must be deleted (part of leak #435)

DanCodedThis:
Partial #435

tl;dr: 
Error was that we checked the record expiry date two times: first => now, second < now, effectively filtering out the session, never deleting it, and just removing the record with the session id, but not the session itself.

Much more info is in the #435 last post.


rampage644 (commented):
LGTM, would be great to have this case covered by tests

YaroslavLitvinov:
Issue couldn't be partially closed, mentioning it in the following way `closes #⁹9999` will close an issue 9999

DanCodedThis:
> Issue couldn't be partially closed, mentioning it in the following way `closes #⁹9999` will close an issue 9999

@YaroslavLitvinov removed "closes"

DanCodedThis:
> LGTM, would be great to have this case covered by tests

@rampage644 Working on it, but it proves difficult to not create a cyclic dependency (I think I figured it out, but it is hard to test)

DanCodedThis:
> > LGTM, would be great to have this case covered by tests
> 
> @rampage644 Working on it, but it proves difficult to not create a cyclic dependency (I think I figured it out, but it is hard to test)

I did test it manually with lower expiration time (worked as expected)

rampage644 (approved):

DanCodedThis:
Test error unrelated to PR (`api-ui` web_assests tarball creation), merging

PR #760, opened 2025-05-12
removed wrong jwt check

jatin510:
the initial value jwt secret is empty.
and this check is wrong, as the jwt token is created after verification is done and we decide what dat to store in jwt
.


YaroslavLitvinov:
Having an empty secret is definitely not desirable for non-development environments. So far, an empty secret has only been allowed for development purposes, to avoid breaking existing features and tests.

Instead, we want to encourage users to explicitly specify a secret via the --jwt-secret argument. This way, if the secret is missing or invalid, all authentication handlers will fail due to the ensure_jwt_secret_is_valid check. Removing this validation from the handlers could lead to silently accepting an empty or meaningless secret, which we want to avoid.

Issue #758, opened 2025-05-12
[Discussion] SlateDB tables as memory views for sql queries


osipovartem:
**Description:**

The idea is to support a set of predefined **SlateDB metadata tables** (volumes, databases, schemas, history, tables) as in-memory views, similar to how information_schema.* is implemented.

When a query targets one of these tables, we dynamically create an in-memory view backed by runtime data. This approach allows full SQL access to metadata without having to implement custom TableProvider, ExecutionPlan, or DataSource for each table.

This would significantly simplify integration and provide a flexible way to expose metadata to users through SQL.

rampage644:
Could you elaborate with a code example on what how this is going to be used and where exactly?
Also, having some code sample would help to better understand the proposal

DanCodedThis:
> Could you elaborate with a code example on what how this is going to be used and where exactly? Also, having some code sample would help to better understand the proposal

It's a lot easier to implement then clothing SlateDb in the TableProvider, SchemaProvider, FileSource, etc for the same result (ability to use SQL on query_history, etc). Since it's just a view (we create record batches in place for each select dynamicly)

PR #764, opened 2025-05-12
Implemented sql access to SlateDB tables as views

osipovartem:
Basic implementation for #758

```sql
SELECT * FROM slatedb.public.volumes
SELECT * FROM slatedb.public.databases
SELECT * FROM slatedb.public.volumes WHERE volume_type = 'file'
...
```

### DIFF
--- a/crates/core-executor/src/query.rs
+++ b/crates/core-executor/src/query.rs
@@ -54,6 +54,9 @@ use super::session::UserSession;
 use super::utils::{NormalizedIdent, is_logical_plan_effectively_empty};
 use crate::datafusion::visitors::{functions_rewriter, json_element};
 use df_catalog::catalog::CachingCatalog;
+use df_catalog::catalogs::slatedb::schema::{
+    SLATEDB_CATALOG, SLATEDB_SCHEMA, SlateDBViewSchemaProvider,
+};
 use tracing_attributes::instrument;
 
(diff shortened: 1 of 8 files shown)


DanCodedThis (approved):
LGTM, this a good middle ground between clothing SlateDb into a lot of DF Traits, to keep it much simpler and still powerful to use

rampage644 (commented):
So the only change is final `*View` implements `PartitionStream` instead of `TableProvider`? 

Is partitionstream actually is tableprovider?

osipovartem:
> So the only change is final `*View` implements `PartitionStream` instead of `TableProvider`?
> 
> Is partitionstream actually is tableprovider?

TableProvider has scan method which returns ExecutionPlan, which should be also implemented with additional DataSource implementation. The same for insert_into. The approach above is **read only** as we have all the data in memory (not in real slatedb db files, we read tables only once), the tables should be lightweight, like volumes, databases, schemas, and this approach doesn't allow inserts (volumes and databases should be created via UI for now). So it's just a views, not real tables. Also seems it is a good way to make such tables non-writable because only specific users should have permission to update records via UI.

rampage644:
> TableProvider has scan method which returns ExecutionPlan, which should be also implemented with additional DataSource implementation. The same for insert_into.

This makes sense, appreciate the explanation

DanCodedThis:
@osipovartem merging?

PR #769, opened 2025-05-12
add static-assets crate having its own build script

### LINKED ISSUE #768: Fix web_assets build script
Due to a parallel build having a non-defined build order, we have an issue with a tarball artifacts
when it's trying to compile bytes from file before that file was created by build script.

Error:
```
 error: couldn't read `/home/runner/work/embucket/embucket/ui/dist.tar`: No such file or directory (os error 2)
  --> crates/api-ui/src/web_assets/handler.rs:17:35
   |
17 | const WEB_ASSETS_TARBALL: &[u8] = include_bytes!(env!("WEB_ASSETS_TARBALL_PATH"));
```

Fix it in the following way: 
Create a separate crate with a build script that is guarantied to run before its build process started.
This crate itself will include static bytes.

YaroslavLitvinov:
closes #768


crates/api-ui/Cargo.toml
```diff
@@ -10,6 +10,7 @@ core-metastore = { path = "../core-metastore" }
 core-utils = { path = "../core-utils" }
 core-executor = { path = "../core-executor" }
 core-history = { path = "../core-history" }
+static-assets = { path = "../static-assets" }
```
rampage644: let's go with `api-ui-static` or `api-ui-static-assets` ?

rampage644 (approved):
For consistency would be great to have new crate named somehow similiar to api-ui, otherwise LGTM!

Issue #683, opened 2025-05-01
[BUG] Missing function: startswith


rampage644:
**Function:** `startswith`
**Category:** String & binary
**Status:** Failed (Return Code: 1)
**Documentation:** https://docs.snowflake.com/en/sql-reference/functions/startswith

**Test SQL:**
```sql
SELECT STARTSWITH('Snowflake', 'Snow');
```

**Expected Result:**
```
N/A
```

**Actual Result:**
```
SELECT STARTSWITH('Snowflake', 'Snow');
```

**Stderr:**
```
╭─ Error ──────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ 000200: 200: DataFusion error: Error during planning: Invalid function 'startswith'.                         │
│ Did you mean 'starts_with'?                                                                                  │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

**Parent EPIC:** #576

AmosAidoo:
I took a look at this. I can take it on possibly after a few clarifications.
Firstly as the error message stated, `starts_with` comes built-in with DataFusion. Does it make sense then to only alias the already existing implementation?
Secondly not sure if this is the right place to post this but I noticed the UI doesn't render boolean and null values. Do I create an issue for that separately as well so that I make a different PR for that.

rampage644:
Yes, it makes totally sense to add an alias.

> Secondly not sure if this is the right place to post this but I noticed the UI doesn't render boolean and null values. Do I create an issue for that separately as well so that I make a different PR for that.

Yes, please create a separate issue for this case.

camuel:
thanks @AmosAidoo !!!, please wait with alias, the PR and PR review was fully AI generated as part of experiment, we are thinking how we should introduce best functions impl. into the project, as we plan to be strictly wire compatible with Snowflake (meaning start_width should not work in Embucket and so forth with all other functions, only Snowflake functions must be supported).

AmosAidoo:
@camuel sure makes sense.

PR #770, opened 2025-05-12
Add Snowflake STARTSWITH function implementation

rampage644:
# Add Snowflake STARTSWITH function implementation

## Purpose
This PR adds a compatible implementation of the Snowflake's STARTSWITH function as documented in the [Snowflake documentation](https://docs.snowflake.com/en/sql-reference/functions/startswith).

## Implementation
- Created a new `startswith.rs` file in the `df-builtins` crate
- Implemented the `StartsWithFunc` struct and related functions
- Added the function to the registry in `lib.rs` 
- Function follows Snowflake's specification:
  - Returns TRUE if the first expression starts with the second expression
  - Returns NULL if either input is NULL
  - Returns FALSE otherwise

## Testing
- Added comprehensive test cases:
  - Tests for scalar values
  - Tests for NULL handling
  - Tests for table operations

## Related Documentation
Reference: https://docs.snowflake.com/en/sql-reference/functions/startswith

### DIFF
--- a/crates/df-builtins/src/startswith.rs
+++ b/crates/df-builtins/src/startswith.rs
@@ -1,0 +1,223 @@
+use datafusion::arrow::array::{Array, ArrayRef, BooleanArray, BooleanBuilder, StringArray};
+use datafusion::arrow::datatypes::DataType;
+use datafusion_common::{DataFusionError, Result, ScalarValue, internal_err};
+use datafusion_expr::{ColumnarValue, ScalarUDFImpl, Signature, Volatility};
+use std::any::Any;
+use std::sync::Arc;
+
+/// STARTSWITH function
+/// Syntax: STARTSWITH(expr1, expr2)
+/// Returns: BOOLEAN - TRUE if expr1 starts with expr2, NULL if either input is NULL, FALSE otherwise
+#[derive(Debug)]
+pub struct StartsWithFunc {
+    signature: Signature,
+    aliases: Vec<String>,
+}
+
+impl Default for StartsWithFunc {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl StartsWithFunc {
+    pub fn new() -> Self {
+        Self {
+            signature: Signature::uniform(
+                2,
+                vec![DataType::Utf8, DataType::Utf8View],
+                Volatility::Immutable,
+            ),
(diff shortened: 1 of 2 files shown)


rampage644:
Closes #683

crates/df-builtins/src/startswith.rs
```diff
@@ -0,0 +1,223 @@
...
+        for arg in args {
+            if let ColumnarValue::Array(a) = arg {
+                array_size = a.len();
+                break;
+            }
+        }
+        let is_scalar = array_size == 1;
+
+        // Convert arguments to string arrays
+        let expr1 = match &args[0] {
```
osipovartem, line 81:
```rust
    fn invoke_with_args(&self, args: datafusion_expr::ScalarFunctionArgs) -> Result<ColumnarValue> {
        let args = &args.args;
        let array_size = args
            .iter()
            .find_map(|arg| match arg {
                ColumnarValue::Array(a) => Some(a.len()),
                ColumnarValue::Scalar(_) => None,
            })
            .unwrap_or(1);
        let is_scalar = array_size == 1;

        let result = starts_with_impl(args, array_size)?;
        if is_scalar {
            // If all inputs are scalar, keeps output as scalar
            let result = ScalarValue::try_from_array(&result, 0)?;
            Ok(ColumnarValue::Scalar(result))
        } else {
            Ok(ColumnarValue::Array(Arc::new(result)))
        }
    }
```
we can use this as a template for funcs generator or smth like that
also

```rust
pub fn to_primitive_array<T>(col: &ColumnarValue) -> Result<PrimitiveArray<T>>
where
    T: datafusion::arrow::datatypes::ArrowPrimitiveType,
{
    match col {
        ColumnarValue::Array(array) => Ok(array.as_primitive::<T>().to_owned()),
        ColumnarValue::Scalar(scalar) => {
            let value = scalar.to_array()?;
            Ok(value.as_primitive::<T>().to_owned())
        }
    }
}
```

The usage
```rust
let hours = to_primitive_array::<Int64Type>(&column)?;```
```

Also very useful func take_function_args to parse all columns from args
```rust
let [years, months, days] = take_function_args("date_from_parts", args)?;
```

rampage644:
Closing this experimental PR

Issue #783, opened 2025-05-13
api: UI Navigation trees no schemas


DanCodedThis:
- Schemas are not showed if there is no table attached to it
- Incorrect tests

nikitastryuk:
Not just about schemas actually — newly created databases, schemas, and tables do not appear in the navigation trees.

DanCodedThis:
сс @osipovartem

PR #788, opened 2025-05-13
Fix navigation tree, add created schemas and tables from UI

### LINKED ISSUE #785: api: UI Navigation trees special DF view
- Special DF view for Navigation trees instead of go in for information_schema.databases, information_schema.schemata, information_schema.tables
- Request is becoming slow (important for UI, autocompletion) this should speed this up (since one SQL will be used)
- Maybe not super important because of the cache

osipovartem:
- Added new VIEW to build navigation tree
- Fixed missing schemas/tables when creating from UI
- removed multiple queries (replaced with single query for whole tree)
Closes #785  and #783


DanCodedThis (approved):
LGTM

PR #802, opened 2025-05-14
Add background task to auto-refresh internal catalogs

### LINKED ISSUE #796: api: UI Navigation Trees store View
- Invalidate navigation tree after DB creation
- Other changes, cc @osipovartem

osipovartem:
📌 **Summary**
This pull request introduces a background task that periodically checks the metastore for newly created databases and dynamically adds them as internal catalogs to the current session. This ensures that users who create new databases don’t need to start new session.

✨ **Motivation**
Previously, new databases created via the metastore would not appear in the catalog list until the process was restarted or manually reloaded. This limited usability in multi-tenant or dynamic environments where databases are frequently created on-the-fly.
To address this, we introduce a periodic refresh mechanism that automatically integrates new internal catalogs into the session.

The code:
-  Spawns a background tokio task.
-  Every 10 seconds:
   - Calls internal_catalogs() to get the current list of databases.
   - Inserts any new catalogs into self.catalogs if not already present.
   - Logs any errors that occur during refresh.
   - Ensures thread safety using DashMap for concurrent catalog access.
Closes #796 and #783

### DIFF
--- a/crates/df-catalog/src/catalog_list.rs
+++ b/crates/df-catalog/src/catalog_list.rs
@@ -23,6 +23,8 @@ use object_store::local::LocalFileSystem;
 use snafu::ResultExt;
 use std::any::Any;
 use std::sync::Arc;
+use std::time::Duration;
+use tokio::time::interval;
 use url::Url;
 
 pub const DEFAULT_CATALOG: &str = "embucket";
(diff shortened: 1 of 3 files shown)


DanCodedThis (approved):
LGTM

DanCodedThis:
@osipovartem quick question, if we create volume from UI does it also needs to be refreshed?

osipovartem:
> @osipovartem quick question, if we create volume from UI does it also needs to be refreshed?

No, since we don't show volumes in navigation tree

crates/df-catalog/src/catalog_list.rs
```diff
@@ -197,6 +199,44 @@ impl EmbucketCatalogList {
...
+    /// - Adds any new catalogs not already present in `self.catalogs`.
+    ///
+    /// If any error occurs during fetching or processing, the error is logged via `tracing::warn`
+    /// but does not interrupt the loop.
+    ///
+    /// This ensures that newly created databases are gradually recognized and integrated
+    /// into the query engine without requiring a full session restart.
+    pub fn start_refresh_internal_catalogs_task(self: Arc<Self>) {
+        tokio::spawn(async move {
+            let mut interval = interval(Duration::from_secs(10));
```
rampage644: nitpick: can the be a parameter?

crates/df-catalog/src/catalog_list.rs
```diff
@@ -197,6 +199,44 @@ impl EmbucketCatalogList {
...
+            loop {
+                interval.tick().await;
+                match self.internal_catalogs().await {
```
rampage644, line 223: Why not `refresh`? Why omit external_catalogs?

rampage644 (commented):
I struggle with overall understanding the problem this issue solves from the user perspective. Is it keeping in-process catalog in sync with just internal catalog? What about external? Is is per-session catalog sync with in-process catalog?

Perhaps good start is a diagram on how this supposed to function from several simultaneous sessions in-flight?

crates/df-catalog/src/catalog_list.rs
osipovartem: It's related to internal catalogs only for now (other catalogs are refreshed by calling sql queries)

osipovartem:
> I struggle with overall understanding the problem this issue solves from the user perspective. Is it keeping in-process catalog in sync with just internal catalog? What about external? Is is per-session catalog sync with in-process catalog?
> 
> Perhaps good start is a diagram on how this supposed to function from several simultaneous sessions in-flight?

In case we have common catalog list for all sessions we should probably update sync job for external catalogs
For now it is not possible to create external catalog (database) from UI. So it is related to internal catalogs only, also it's just a simple call to slatedb. For external catalogs it will be different logic and there is no reason to add it right now to not complicate the code

rampage644 (approved):

PR #804, opened 2025-05-14
`is_object` scalar function

### LINKED ISSUE #803: [BUG] Missing function: is_object
https://docs.snowflake.com/en/sql-reference/functions/is_object

ravlio:
closes #803

### DIFF
--- a/crates/df-builtins/src/is_object.rs
+++ b/crates/df-builtins/src/is_object.rs
@@ -1,0 +1,148 @@
+use datafusion::arrow::array::builder::BooleanBuilder;
+use datafusion::arrow::array::{Array, as_string_array};
+use datafusion::arrow::datatypes::DataType;
+use datafusion::error::Result as DFResult;
+use datafusion::logical_expr::{ColumnarValue, Signature, Volatility};
+use datafusion_common::ScalarValue;
+use datafusion_expr::{ScalarFunctionArgs, ScalarUDFImpl};
+use serde_json::Value;
+use std::any::Any;
+use std::sync::Arc;
+
+// is_array SQL function
+// Returns TRUE if the VARIANT argument holds an ARRAY value.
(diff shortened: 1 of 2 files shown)


crates/df-builtins/src/is_object.rs
```diff
@@ -0,0 +1,148 @@
...
+
+// is_array SQL function
+// Returns TRUE if the VARIANT argument holds an ARRAY value.
+// Syntax: IS_ARRAY( <variant_expr> )
+// Arguments:
+// - variant_expr
+// An expression that evaluates to a value of type VARIANT.
+// Example SELECT IS_ARRAY('[1,2,3]') as v;;
+// Note `is_array` returns
+// Returns a BOOLEAN value or NULL.
```
DanCodedThis: nitpick: double comment

crates/df-builtins/src/is_object.rs
```diff
@@ -0,0 +1,148 @@
...
+
+        let mut b = BooleanBuilder::with_capacity(arr.len());
+        let input = as_string_array(&arr);
+        for v in input {
+            if let Some(v) = v {
+                match serde_json::from_str::<Value>(v) {
+                    Ok(v) => {
+                        b.append_value(v.is_object());
+                    }
+                    Err(_) => b.append_value(false),
```
DanCodedThis, line 76: If we can't parse then it's not an object?

DanCodedThis (approved):
LGTM

crates/df-builtins/src/is_object.rs
ravlio: yes, according to doc:

```Returns a BOOLEAN value or NULL.

Returns TRUE if the VARIANT value contains an OBJECT value. Otherwise, returns FALSE.

If the input is NULL, returns NULL without reporting an error.```

DanCodedThis (approved):

PR #806, opened 2025-05-14
Add information schema insta tests

### LINKED ISSUE #761: Add information schema related integration tests
Since we moved the information schema implementation inside the `embucket` it would be great to have comprehensive test coverage for it.

osipovartem:
Added insta tests with snapshots for information schema tables

Routines and parameters tables can return rows in a non-deterministic order,
causing snapshot tests to randomly fail.
To avoid flakiness, we group results by primary identifying columns and count the rows per group.
This makes the output deterministic and ensures stable test snapshots.

Closes #761


rampage644 (commented):
Would be great to amend tests with `SHOW TABLES IN <my-schema>`, `SHOW SCHEMAS IN <my-database>`, `DESCRIBE TABLE` etc. For all query commands interacting with underlying information_schema

osipovartem:
> Would be great to amend tests with `SHOW TABLES IN <my-schema>`, `SHOW SCHEMAS IN <my-database>`, `DESCRIBE TABLE` etc. For all query commands interacting with underlying information_schema

Let's do this on the next PR since we have to add ShowDatabases, ShowTables, ShowColumns, etc. to QueryEngine

PR #810, opened 2025-05-14
`array_flatten` scalar function

### LINKED ISSUE #819: array_flatten

ravlio:
closes #819

### DIFF
--- a/crates/df-builtins/src/array_flatten.rs
+++ b/crates/df-builtins/src/array_flatten.rs
@@ -118,9 +118,6 @@ fn flatten(v: &str) -> DFResult<Option<String>> {
         Value::Array(arr) => {
             let mut result = Vec::new();
             for item in arr {
-                if item.is_null() {
-                    return Ok(None);
-                }
                 match item {
                     Value::Null => return Ok(None),
                     Value::Array(inner) => {
--- a/crates/df-builtins/src/array_flatten.rs
+++ b/crates/df-builtins/src/array_flatten.rs
@@ -16,10 +16,10 @@ use std::sync::Arc;
 // 
 // Arguments:
 // - <array_expr>
-// The ARRAY of ARRAYs to flatten.
-// If any element of array is not an ARRAY, the function reports an error.
+//   The ARRAY of ARRAYs to flatten.
+//   If any element of array is not an ARRAY, the function reports an error.
 //
-// Note: `array_flatten` returns
+// Return:
 // This function produces a single ARRAY by joining together all the ARRAYs within the input array. If the input array is NULL or contains any NULL elements, the function returns NULL.
 #[derive(Debug)]
 pub struct ArrayFlattenFunc {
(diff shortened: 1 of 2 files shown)


crates/df-builtins/src/array_flatten.rs
```diff
@@ -0,0 +1,230 @@
...
+
+        let q = "SELECT ARRAY_FLATTEN('[[1, 2, 3], null, [5, 6]]') as v;";
+        let result = ctx.sql(q).await?.collect().await?;
+
+        assert_batches_eq!(&["+---+", "| v |", "+---+", "|   |", "+---+",], &result);
+
+        let q = "SELECT ARRAY_FLATTEN('[[1, 2, 3], [NULL], [5, 6]]') as v;";
+        let result = ctx.sql(q).await?.collect().await?;
+
+        print_batches(&result)?;
```
rampage644: assert_batches_eq?

crates/df-builtins/src/array_flatten.rs
```diff
@@ -0,0 +1,230 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use datafusion::arrow::util::pretty::print_batches;
+    use datafusion::prelude::SessionContext;
+    use datafusion_common::assert_batches_eq;
+    use datafusion_expr::ScalarUDF;
+
+    #[tokio::test]
+    async fn test_basic() -> DFResult<()> {
```
rampage644: Am I correct test cases cover only `ColumnarValue::Scalar(v)` and not `ColumnarValue::Array(v)`?

ravlio: yes it is. I added an array based test as well. Most of the time I implement universal solution that converts Scalar to Array so it is not necessary to make different tests but here I tried to optimize solution a bit.

crates/df-builtins/src/array_flatten.rs
```diff
@@ -0,0 +1,257 @@
...
+            let mut result = Vec::new();
+            for item in arr {
+                if item.is_null() {
+                    return Ok(None);
+                }
+                match item {
+                    Value::Null => return Ok(None),
+                    Value::Array(inner) => {
+                        for subitem in inner {
+                            result.push(flatten_inner(subitem)?);
```
Vedin, line 125: I don't fully understand why we need flatten_inner at all? So, if we flatten only one layer of the array with this function, shouldn't we just do it once and ignore all inner layers?

Vedin: I think a little bit about current solution. So, you don't need this flatten_inner at all. Only check the subitem for NULL in place and push subitem or 'undefined' in case of NULL inside array.

ravlio: `flatten_inner` converts NULLs to "undefined" strings. It should be done on each level according to Snowflake. How do you plan to do it w/o `flatten_inner` recursive function?

Vedin: Oh, I see. Agree

Vedin (approved):
lgtm

PR #813, opened 2025-05-15
[BUG] 629: implement missing insert function

### LINKED ISSUE #629: [BUG] Missing function: insert
**Function:** `insert`
**Category:** String & binary
**Documentation:** https://docs.snowflake.com/en/sql-reference/functions/insert

**Test SQL:**
```sql
SELECT INSERT('abcdef', 3, 2, 'XX');
```

**Expected Result:**
```
N/A
```

**Actual Result:**
```
SELECT INSERT('abcdef', 3, 2, 'XX');
```

**Stderr:**
```
│ 000200: 200: DataFusion error: Error during planning: Invalid function 'insert'.                                                                            │
│ Did you mean 'instr'?                                                                                                                                       │
```

**Parent EPIC:** #576

AmosAidoo:
## Which issue does this PR close?

- Closes #629 

## Changes

- Added a new scalar udf, `Insert` and implemented logic according to snowflake docs
- Added a new macro, adopted from itertools at [https://docs.rs/itertools/latest/src/itertools/lib.rs.html#321-360](https://docs.rs/itertools/latest/src/itertools/lib.rs.html#321-360) to zip multiple iterators. It was useful for the current task.

## How was this tested?

- Added automated tests to validate functionality
- All existing tests pass

### DIFF
--- a/crates/df-builtins/src/insert.rs
+++ b/crates/df-builtins/src/insert.rs
@@ -1,0 +1,473 @@
+use datafusion::arrow::array::{BinaryBuilder, StringBuilder};
+use datafusion::arrow::datatypes::DataType;
+use datafusion::error::Result as DFResult;
+use datafusion_common::cast::{as_binary_array, as_int64_array, as_string_array};
+use datafusion_common::{exec_err, plan_err};
+use datafusion_expr::{
+    ColumnarValue, ScalarFunctionArgs, ScalarUDFImpl, Signature, TypeSignature, Volatility,
+};
+use std::any::Any;
+use std::sync::Arc;
+
+// insert SQL function
+// Replaces a substring of the specified length, starting at the specified position, with a
+// new string or binary value.
+// This function should not be confused with the INSERT DML command.
+// Syntax: INSERT( <base_expr>, <pos>, <len>, <insert_expr> )
+// Arguments:
+// <base_expr>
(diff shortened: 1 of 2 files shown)


crates/df-builtins/src/insert.rs
```diff
@@ -0,0 +1,333 @@
...
+        Self::new()
+    }
+}
+
+impl Insert {
+    pub fn new() -> Self {
+        Self {
+            signature: Signature::exact(
+                vec![
+                    DataType::Utf8,
```
Vedin: I have some doubts about this. I believe we want to have signature for DataType::Binary here as well. Even datafusion currently doesn't support binary datatype it supports Binary values. So, something like this should work:
`select INSERT(X'1234567890ABCDEF', 4, 2, X'FFFF');`
with the result:
`123456ffffabcdef`

Vedin (changes requested):
Great! I didn't verify macros logic but in general looks good to me. I'd only ask to verify it works with Binary as well.

crates/df-builtins/src/insert.rs
AmosAidoo: Thank you very much for the review. I have handled the binary case and updated the tests

Vedin (approved):

AmosAidoo:
I have resolved the merge conflict

Vedin (approved):

PR #814, opened 2025-05-15
`array_to_string` scalar function

### LINKED ISSUE #820: array_to_string

ravlio:
closes #820

### DIFF
--- a/crates/df-builtins/src/array_to_string.rs
+++ b/crates/df-builtins/src/array_to_string.rs
@@ -127,7 +127,7 @@ fn to_string(v: &Value, sep: &str) -> DFResult<String> {
                         })?;
                         res.push(r);
                     }
-                    Value::Null => res.push("".to_string()),
+                    Value::Null => res.push(String::new()),
                 }
             }
         }
--- a/crates/df-builtins/src/array_to_string.rs
+++ b/crates/df-builtins/src/array_to_string.rs
@@ -127,7 +127,7 @@ fn to_string(v: &Value, sep: &str) -> DFResult<String> {
                         })?;
                         res.push(r);
                     }
-                    Value::Null => res.push("null".to_string()),
+                    Value::Null => res.push("".to_string()),
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


crates/df-builtins/src/array_to_string.rs
```diff
@@ -0,0 +1,204 @@
...
+    fn invoke_with_args(&self, args: ScalarFunctionArgs) -> DFResult<ColumnarValue> {
+        let ScalarFunctionArgs { args, .. } = args;
+
+        match (&args[0], &args[1]) {
+            (ColumnarValue::Array(arr), ColumnarValue::Scalar(ScalarValue::Utf8(Some(sep)))) => {
+                let mut res = StringBuilder::with_capacity(arr.len(), 1024);
+                let arr = as_string_array(arr);
+                for v in arr {
+                    if let Some(v) = v {
+                        let json: Value = serde_json::from_str(v).map_err(|err| {
```
Vedin, line 78: Any reason why use `json` for parse array to string everywhere in this file?

crates/df-builtins/src/array_to_string.rs
```diff
@@ -0,0 +1,204 @@
...
+                        })?;
+                        res.push(r);
+                    }
+                    Value::Object(v) => {
+                        let r = serde_json::to_string(&v).map_err(|err| {
+                            DataFusionError::Execution(format!("failed to serialize JSON: {err:?}"))
+                        })?;
+                        res.push(r);
+                    }
+                    Value::Null => res.push("null".to_string()),
```
Vedin: > A NULL in an array is converted to an empty string in the result.

Vedin (changes requested):

crates/df-builtins/src/array_to_string.rs
ravlio: thanks, fixed

crates/df-builtins/src/array_to_string.rs
ravlio: the expected input is Utf8 json string so we have to deserialize it. We can't use ARRAY, UNION, STRUCT Arrow types because they are not semi-structured so we decided to stick with raw json.

Vedin (approved):

Issue #562, opened 2025-05-01
`Discussion` history_store: Query handler refactor


DanCodedThis:
# Problem

- `QueryRecord` is created in the `fn query` of `RecordingExecutionService` (where now it is added into the `history_store`).
- `QueryRecord` is creating inside the query handler `ui/queries/handlers.rs` just to _**parse**_ the `ResultSet` to pass it down to the result of the handler.
- We need to somehow decouple this, since `fn query` as per the trait should return only `RecordBatch`s.

# Possible solution

If we remove the `ResultSet` from the query handler it should solve 90%-99% of the problem. So this should be a byproduct of solving #445.

### Query handler

```rust
pub async fn query(
    DFSessionId(session_id): DFSessionId,
    State(state): State<AppState>,
    Json(payload): Json<QueryCreatePayload>,
) -> QueriesResult<Json<QueryCreateResponse>> {
    //
    // Note: This handler allowed to return error from a designated place only,
    // after query record successfuly saved result or error.

    let query_context = QueryContext::new(
        payload
            .context
            .as_ref()
            .and_then(|c| c.get("database").cloned()),
        payload
            .context
            .as_ref()
            .and_then(|c| c.get("schema").cloned()),
        payload.worksheet_id,
    );

    //TODO: map to result correctly without using duplicate code
    let mut query_record =
        embucket_history::QueryRecord::query_start(&payload.query, payload.worksheet_id);
    let query_res = state
        .execution_svc
        .query(&session_id, &payload.query, query_context)
        .await;

    match query_res {
        Ok((ref records, ref columns)) => {
            let result_set = ResultSet::query_result_to_result_set(records, columns);
            match result_set {
                Ok(result_set) => {
                    let encoded_res = serde_json::to_string(&result_set);

                    if let Ok(encoded_res) = encoded_res {
                        let result_count = i64::try_from(records.len()).unwrap_or(0);
                        query_record.query_finished(result_count, Some(encoded_res));
                    }
                    // failed to wrap query results
                    else if let Err(err) = encoded_res {
                        query_record.query_finished_with_error(err.to_string());
                    }
                }
                // error getting result_set
                Err(err) => {
                    query_record.query_finished_with_error(err.to_string());
                }
            }
        }
        // query error
        Err(ref err) => {
            // query execution error
            query_record.query_finished_with_error(err.to_string());
        }
    }

    // // add query record
    // if let Err(err) = state.history_store.add_query(&query_record).await {
[...]

PR #815, opened 2025-05-15
Issues/729 execution service query

### LINKED ISSUE #729: ExecutionService::query should return a query id
Following should return query id  along with query result or without query result:

```
async fn query(
        &self,
        session_id: &str,
        query: &str,
        query_context: QueryContext,
    ) -> ExecutionResult<(Vec<RecordBatch>, Vec<ColumnInfo>)> 
```

It makes sense from point of view of RecordingExecutionService to identify saved query,
but it probably useless for just ExecutionService (outside of recording) as it nowhere to use.

YaroslavLitvinov:
Closes #729 #562 
* Add QueryResultData { records, columns_info, query_id };
* Extend WorksheetsStore with get_query function;
* Remove duplicated code from queries handler;


DanCodedThis (approved):
LGTM

Issue #830, opened 2025-05-15
[Discussion] catalog: refactor SlateDb query as DF View


DanCodedThis:
Minor changes to how querying history, meta is made (naming)

cc @osipovartem

DanCodedThis:
@camuel I would like to make the naming of the metastore and history store querying like this:
```
SELECT * FROM store.meta.volumes
SELECT * FROM store.meta.databases
SELECT * FROM store.meta.schemas
SELECT * FROM store.meta.tables

SELECT * FROM store.history.worksheets
SELECT * FROM store.history.queries
```

My question: is there a representation in Snowflake akin to this stores (not `information_schema`), if so how should we name it to adhere to it?

Right now we only have DF Views for the metastore named like this:
```
SELECT * FROM slatedb.public.volumes
SELECT * FROM slatedb.public.databases
SELECT * FROM slatedb.public.schemas
SELECT * FROM slatedb.public.tables
```

cc @osipovartem @YaroslavLitvinov @rampage644

rampage644:
Several unconnected thoughts
 * We might want to let operator specify the name of the database and schema to access internal data
 * Some time in the future slatedb backed tables could be as usual as iceberg tables right now
 * Overall, it might be promising to continue using `information_schema` approach instead of dedicated `slatedb` or `store` reserved databases
 * We might end having dedicated SQL syntax to operate on volumes at least (as there is for databases, schemas, tables)

@camuel Any thoughts on `information_schema`intergace mismatch datafusion <-> snowflake?

rampage644:
There is similar issue with builtin functions: what to do with functions that exist in datafusion, but not in snowflake? From user perspective, expected behavour is to get "unknown function" error rather than seeing the query succesfully executing (one example here is `starts_with` function, in snowflake it is `startswith`).

I think there is value **not** having anything unexpected exposed to the user.

camuel:
[...]

PR #831, opened 2025-05-15
Schemas DF View Store (SlateDb)

### LINKED ISSUE #826: catalog: DF Store View for Metastore schemas
Add the ability to query slatedb schemas as DF View

DanCodedThis:
Closes #826

```
SELECT * FROM slatedb.public.schemas
SELECT * FROM slatedb.public.schemas WHERE database_name = 'mydb1'
...
```

- Next will be starting to use this in `/ui/../schemas` GET
- The change in metastore allows us to get all the schemas (otherwise we would get an empty result), the problem was the extra `/`

### DIFF
--- a/crates/core-metastore/src/metastore.rs
+++ b/crates/core-metastore/src/metastore.rs
@@ -380,6 +380,7 @@ impl Metastore for SlateDBMetastore {
     }
 
     fn iter_schemas(&self, database: &DatabaseIdent) -> VecScanIterator<RwObject<Schema>> {
+        //If database is empty, we are iterating over all schemas
         let key = if database.is_empty() {
             KEY_SCHEMA.to_string()
         } else {
--- a/crates/core-metastore/src/metastore.rs
+++ b/crates/core-metastore/src/metastore.rs
@@ -380,7 +380,11 @@ impl Metastore for SlateDBMetastore {
     }
 
     fn iter_schemas(&self, database: &DatabaseIdent) -> VecScanIterator<RwObject<Schema>> {
-        let key = format!("{KEY_SCHEMA}/{database}");
+        let key = if database.len() != 0 {
+            format!("{KEY_SCHEMA}/{database}")
+        } else {
+            format!("{KEY_SCHEMA}")
+        };
         self.iter_objects(key)
     }
 
--- a/crates/core-metastore/src/metastore.rs
+++ b/crates/core-metastore/src/metastore.rs
@@ -380,10 +380,10 @@ impl Metastore for SlateDBMetastore {
     }
 
     fn iter_schemas(&self, database: &DatabaseIdent) -> VecScanIterator<RwObject<Schema>> {
-        let key = if database.len() != 0 {
-            format!("{KEY_SCHEMA}/{database}")
+        let key = if database.is_empty() {
+            KEY_SCHEMA.to_string()
         } else {
-            format!("{KEY_SCHEMA}")
+            format!("{KEY_SCHEMA}/{database}")
         };
         self.iter_objects(key)
     }
--- a/crates/df-catalog/src/catalogs/slatedb/config.rs
+++ b/crates/df-catalog/src/catalogs/slatedb/config.rs
@@ -56,8 +56,8 @@ impl SlateDBViewConfig {
             .map_err(|e| DataFusionError::Execution(format!("failed to get schemas: {e}")))?;
         for schema in schemas {
             builder.add_schema(
-                schema.ident.schema.as_str(),
-                schema.ident.database.as_str(),
+                &schema.ident.schema,
+                &schema.ident.database,
                 schema.created_at.to_string(),
                 schema.updated_at.to_string(),
             );
--- a/crates/df-catalog/src/catalogs/slatedb/config.rs
+++ b/crates/df-catalog/src/catalogs/slatedb/config.rs
@@ -50,12 +50,12 @@ impl SlateDBViewConfig {
     ) -> datafusion_common::Result<(), DataFusionError> {
         let schemas = self
             .metastore
-            .iter_schemas()
+            .iter_schemas(&"".to_string())
             .collect()
             .await
-            .map_err(|e| DataFusionError::Execution(format!("failed to get databases: {e}")))?;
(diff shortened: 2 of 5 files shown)


DanCodedThis:
@osipovartem maybe a good idea to `schemas` -> `schemata` such as to: 

```
SELECT * FROM slatedb.public.schemata
SELECT * FROM slatedb.public.schemata WHERE database_name = 'mydb1'
...
```

crates/core-metastore/src/metastore.rs
```diff
@@ -380,7 +380,11 @@ impl Metastore for SlateDBMetastore {
     }
 
     fn iter_schemas(&self, database: &DatabaseIdent) -> VecScanIterator<RwObject<Schema>> {
-        let key = format!("{KEY_SCHEMA}/{database}");
+        let key = if database.is_empty() {
```
YaroslavLitvinov, line 384: With just KEY_SCHEMA prefix it will iterate over all the schemas in all databases. Is that what you need - iterate ovel all schemas?

crates/df-catalog/src/catalogs/slatedb/config.rs
```diff
@@ -43,4 +44,24 @@ impl SlateDBViewConfig {
...
+    pub async fn make_schemas(
+        &self,
+        builder: &mut SchemasViewBuilder,
+    ) -> datafusion_common::Result<(), DataFusionError> {
+        let schemas = self
+            .metastore
+            .iter_schemas(&String::new())
+            .collect()
+            .await
+            .map_err(|e| DataFusionError::Execution(format!("failed to get schemas: {e}")))?;
```
YaroslavLitvinov, line 56: Good to have snafu error vs error text in a function

crates/df-catalog/src/catalogs/slatedb/config.rs
```diff
@@ -43,4 +44,24 @@ impl SlateDBViewConfig {
...
+        for schema in schemas {
+            builder.add_schema(
```
YaroslavLitvinov, line 58: Use consistent parameters types, here both types of conversions used: as_str(), to_string(). But add_schema expect all arguments of the same type.

crates/df-catalog/src/catalogs/slatedb/schema.rs
```diff
@@ -14,8 +15,9 @@ pub const SLATEDB_SCHEMA: &str = "public";
 pub const SLATEDB_CATALOG: &str = "slatedb";
```
YaroslavLitvinov, line 15: Why Catalog name is slatedb? It's not that generic, it's just a metastore

crates/core-metastore/src/metastore.rs
DanCodedThis: Yep, it's the only way, because we can run sql such as `SELECT * FROM slatedb.public.schemas` we can't put here the actually db name, outside the `WHERE` clause, because the general query doesn't contain such a parameter.

crates/df-catalog/src/catalogs/slatedb/config.rs
DanCodedThis: Did it exactly like Artem did in https://github.com/Embucket/embucket/pull/764

crates/df-catalog/src/catalogs/slatedb/config.rs
DanCodedThis: `impl AsRef<str>` is `&str` (string slice) which is general over the String type

crates/df-catalog/src/catalogs/slatedb/schema.rs
DanCodedThis: I agree, we have an issue about naming, we should consider how snowflake does it: https://github.com/Embucket/embucket/issues/830

crates/df-catalog/src/catalogs/slatedb/config.rs
DanCodedThis: > Good to have snafu error vs error text in a function

I think this could be added to #830

crates/core-metastore/src/metastore.rs
YaroslavLitvinov: It needs some comment at least if updating iter_schemas interface making database optional is not an option

crates/df-catalog/src/catalogs/slatedb/config.rs
YaroslavLitvinov: function declaration has the same type, but arguments provided differently:
```
    schema.ident.schema.as_str(),
    schema.ident.database.as_str(),
    schema.created_at.to_string(),
    schema.updated_at.to_string(),
```

DanCodedThis: We can add `schema.updated_at.to_string().as_str()` & `schema.updated_at.to_string().as_str()`, but it is done already automatically because of the `impl AsRef`

crates/core-metastore/src/metastore.rs
DanCodedThis: My first idea was to make an option, but it won't be used anywhere outside the DF view, so probably overkill here

DanCodedThis: I'll add a comment

crates/df-catalog/src/catalogs/slatedb/config.rs
DanCodedThis: `fn test(param: impl AsRef<str>)` means the same as `fn test<T: AsRef<str>>(param: T)`, it's essentialy the same as the `dyn` keyword (instead of `impl`), but it's using static dispatch (no virtual table)

YaroslavLitvinov: First two already strings
```
    &schema.ident.schema,
    &schema.ident.database,
    schema.created_at.to_string(),
    schema.updated_at.to_string(),
```

YaroslavLitvinov:
In overall LGTM, comment ned to be added regarding different behavior when database is empty.
`fn iter_schemas(&self, database: &DatabaseIdent) `

YaroslavLitvinov (approved):

YaroslavLitvinov (approved):

PR #833, opened 2025-05-15
Add SHOW * tests

osipovartem:
Added updates for information schema columns
- https://docs.snowflake.com/en/sql-reference/sql/show-databases
- https://docs.snowflake.com/en/sql-reference/sql/show-schemas
- https://docs.snowflake.com/en/sql-reference/sql/show-tables
- https://docs.snowflake.com/en/sql-reference/sql/show-views
- https://docs.snowflake.com/en/sql-reference/sql/show-columns
- https://docs.snowflake.com/en/sql-reference/sql/show-objects
Fixed multiple unsupported statements.
Views returns only views.
Tables returns only tables.
SHOW filtration IN and LIKE, STARTSWITH also supported


Vedin (approved):
lgtm! I really like snapshot tests added! I think it should be common practice for our functionality to add them in case of sql-related changes.

crates/core-executor/src/query.rs
```diff
@@ -268,12 +270,19 @@ impl UserQuery {
...
                 | Statement::Update { .. } => {
                     return Box::pin(self.execute_with_custom_plan(&self.query)).await;
                 }
+                Statement::ShowDatabases { .. }
+                | Statement::ShowSchemas { .. }
+                | Statement::ShowTables { .. }
+                | Statement::ShowColumns { .. }
+                | Statement::ShowViews { .. }
+                | Statement::ShowFunctions { .. }
+                | Statement::ShowObjects { .. } => {
```
Vedin, line 282: I see `ShowObjects` both here and in planner.rs while `ShowSchemas` was moved here. Is it intended? The same question about `ShowVariables` which is presented in `planner.rs` but not here

osipovartem: It will be removed in the next PR
since its already too huge

osipovartem: Also I plan to move ShowVariables to this code

crates/core-executor/src/query.rs
```diff
@@ -957,6 +965,185 @@ impl UserQuery {
...
+            .map(|obj| obj.0.into_iter().map(|ident| ident.to_string()).collect())
+            .unwrap_or_default();
+
+        let (catalog, schema, table_name) = match parts.as_slice() {
+            [one] if table => (self.current_database(), self.current_schema(), one.clone()),
+            [one] => (self.current_database(), one.clone(), String::new()),
+            [s, t] if table => (self.current_database(), s.clone(), t.clone()),
+            [d, s] => (d.clone(), s.clone(), String::new()),
+            [d, s, t] => (d.clone(), s.clone(), t.clone()),
+            _ => (self.current_database(), String::new(), String::new()),
```
DanCodedThis, line 1142: nitpick: hard to understand naming

DanCodedThis (approved):
LGTM

crates/df-catalog/src/information_schema/information_schema.rs
```diff
@@ -60,7 +60,6 @@ impl InformationSchemaProvider {
                     (DF_SETTINGS, InformationSchemaDfSettings::schema()),
                     (ROUTINES, InformationSchemaRoutines::schema()),
                     (PARAMETERS, InformationSchemaParameters::schema()),
-                    (NAVIGATION_TREE, InformationSchemaNavigationTree::schema()),
```
DanCodedThis: We can probably move this to the "slatedb catalog" (different PR perhaps)

osipovartem: This view is not a system view, because of that it was removed from common list, but still available for sql queries

crates/core-executor/src/query.rs
osipovartem: Above we have database, schema, table name 
And then d, s, t as aliases 
I can update it

crates/df-catalog/src/information_schema/information_schema.rs
DanCodedThis: Oh, got it

rampage644:
Does this PR add the snapshots for the failing test here https://github.com/Embucket/embucket/actions/runs/15058349992/job/42328708110 ?

crates/core-executor/src/query.rs
rampage644: re: UserQuery vs ExtendedSqlToRel Statement huge statement handling should be unified in the future, however I still do not have a strong idea how exactly

osipovartem: We can create a separate dir for SHOW statements with specific fn for each Statement

osipovartem:
> Does this PR add the snapshots for the failing test here https://github.com/Embucket/embucket/actions/runs/15058349992/job/42328708110 ?

Yes, it should.

PR #835, opened 2025-05-16
Tables DF View Store (SlateDb)

### LINKED ISSUE #827: catalog: DF Store View for Metastore tables
Add the ability to query slatedb tables as DF View

DanCodedThis:
Closes #827

```
SELECT * FROM slatedb.public.tables
SELECT * FROM slatedb.public.tables WHERE schema_name = 'myschema1'
...
```

- Next will be starting to use this in `/ui/../tables` GET
- The change in metastore allows us to get all the tables (otherwise we would get an empty result), the problem was the two extra `/`

### DIFF
--- a/crates/core-metastore/src/metastore.rs
+++ b/crates/core-metastore/src/metastore.rs
@@ -439,7 +439,12 @@ impl Metastore for SlateDBMetastore {
     }
 
     fn iter_tables(&self, schema: &SchemaIdent) -> VecScanIterator<RwObject<Table>> {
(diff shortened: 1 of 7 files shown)


osipovartem (approved):

DanCodedThis:
@rampage644 @osipovartem @Vedin does anybody know why the df_settings test have fallen here? Is it ok to merge (didn't touch it)?

osipovartem:
> @rampage644 @osipovartem @Vedin does anybody know why the df_settings test have fallen here? Is it ok to merge (didn't touch it)?

Let me fix in my PR

PR #836, opened 2025-05-16
variant support

ravlio:


crates/df-builtins/src/visitors/mod.rs
```diff
@@ -0,0 +1 @@
+pub mod variant;
```
osipovartem: We have visitors in core-executor in datafustion directory. Does it make sense to move visitors from df-builtins to core-executor?
@rampage644

rampage644: @ravlio Mentioned currently there is circular dependency between those. @ravlio could elaborate better on this.

ravlio: moved visitors to the separate crate

rampage644 (approved):

rampage644:
@ravlio Anything else blocking this PR to be merged?

ravlio:
@rampage644 nope, just wanted to gather reviews. Working on resolving conflicts and merging. Please re-approve

crates/visitors/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "visitors"
+version = "0.1.0"
+edition = "2024"
+license-file.workspace = true
+
+[dependencies]
+datafusion = { workspace = true }
```
osipovartem: Do we really need a separate crate for visitors?
nit: the name of the crate doesn't have a context (core/api/df, etc.) 
Why don't you want to put them somewhere in df-builtins?

Vedin (approved):
lgtm ! I didn't check all functions one by one. I think we should merge this and fix all potential problems later.

crates/visitors/Cargo.toml
Vedin: I missed Artem comment but I'd also agree that probably visitor should be placed under some other module if possible

ravlio: ok lets move visitors to functions. I agree that dedicated crate for visitors is too much, but not all the visitors belong to functions, that is the point.

rampage644 (approved):

rampage644 (approved):

jonathanc-n (approved):

Vedin (approved):

Issue #522, opened 2025-04-29
api: UI Use Query engine for all list handlers (Metastore substitution)


DanCodedThis:
- Change the handlers that rely on `Metastore` impl
- Use query engine for aggregation (group by), ordering, sorting, search, etc
- Later, if this approach is doable, use it for query history and worksheets (if needed)

DanCodedThis:
Not needed, until (if) we decide to change the _**Metastore**_ keys to any type of ids, akin to Snowflake `identifier()` "udf"

DanCodedThis:
> Not needed, until (if) we decide to change the _**Metastore**_ keys to any type of ids, akin to Snowflake `identifier()` "udf"

Actually, probably the other way round until we implement search from using DF Store Views for or a full `TableProvider` for `metastore` and `history_store`, this shouldn't be changed

PR #839, opened 2025-05-16
Schemas and Databases search through the query engine

DanCodedThis:
Partial #522

- Changed list handlers for schemas and databases
- Fixed tests
- Added created_at, updated_at to the volume view and database view
- Decided to stop since the pr is already long enough
- And because right now metastore returns more info in list tables and list volumes
- For tables it's understandable that we  need to add `total-files-size` and `total-records` to the view
- For volumes it's not so simple (we don't use the handler yet) we need to understand the scope of what we want to return in the view, right now it seems to return even credentials and a lot of info which, if considering different types won't be possible to put in the view (and we shouldn't)


crates/api-ui/src/databases/handlers.rs
```diff
@@ -194,20 +191,18 @@ pub async fn update_database(
...
-        })
+        .map(|o| Json(DatabaseUpdateResponse { data: o.into() }))
 }
 
 #[utoipa::path(
     get,
     operation_id = "getDatabases",
     params(
-        ("cursor" = Option<String>, Query, description = "Databases cursor"),
+        ("offset" = Option<usize>, Query, description = "Databases offset"),
```
YaroslavLitvinov, line 202: UI is ready to switch from cursor to offset ?

crates/api-ui/src/databases/handlers.rs
```diff
@@ -223,30 +218,52 @@ pub async fn update_database(
...
     State(state): State<AppState>,
 ) -> DatabasesResult<Json<DatabasesResponse>> {
-    state
-        .metastore
-        .iter_databases()
-        .cursor(parameters.cursor.clone())
-        .limit(parameters.limit)
-        .token(parameters.search)
-        .collect()
+    let context = QueryContext::new(None, None, None);
```
YaroslavLitvinov: QueryContext::default() ?

crates/api-ui/src/databases/handlers.rs
```diff
@@ -223,30 +218,52 @@ pub async fn update_database(
     )
 )]
 #[tracing::instrument(level = "debug", skip(state), err, ret(level = tracing::Level::TRACE))]
+#[allow(clippy::unwrap_used)]
 pub async fn list_databases(
+    DFSessionId(session_id): DFSessionId,
     Query(parameters): Query<DatabasesParameters>,
```
YaroslavLitvinov: DatabasesParameters is the same as SchemaParameters, that in fact sql parameters. The need just one struct for sql query parameters, since we switch to generic use.

crates/api-ui/src/databases/handlers.rs
```diff
@@ -223,30 +218,52 @@ pub async fn update_database(
...
 ) -> DatabasesResult<Json<DatabasesResponse>> {
-    state
-        .metastore
-        .iter_databases()
-        .cursor(parameters.cursor.clone())
-        .limit(parameters.limit)
-        .token(parameters.search)
-        .collect()
+    let context = QueryContext::new(None, None, None);
+    let sql_string = "SELECT * FROM slatedb.public.databases".to_string();
```
YaroslavLitvinov, line 229: We are going to change a piece of code by something that is not very consice.

crates/api-ui/src/databases/models.rs
```diff
@@ -7,13 +8,29 @@ use utoipa::{IntoParams, ToSchema};
...
+    pub updated_at: String,
 }
 
 impl From<MetastoreDatabase> for Database {
     fn from(db: MetastoreDatabase) -> Self {
         Self {
             name: db.ident,
             volume: db.volume,
+            //TODO: fix this, we must use a different payload or change the test suite for dbs
+            created_at: "ERROR".to_string(),
```
YaroslavLitvinov, line 20: This ERROR thing can lead to parsing errors, I mean ser/de

crates/api-ui/src/schemas/handlers.rs
```diff
@@ -89,9 +90,19 @@ pub async fn create_schema(
         .query(&session_id, sql_string.as_str(), context)
         .await
         .map_err(|e| SchemasAPIError::Create { source: e })?;
-    Ok(Json(SchemaCreateResponse {
-        data: Schema::new(payload.name, database_name),
-    }))
+    let schema_ident = MetastoreSchemaIdent::new(database_name.clone(), payload.name.clone());
```
YaroslavLitvinov, line 93: This looks strange: schema is created using SQL, and fetched using metastore. 

If that is done just for created_at, updated_at, that means those fields not available for quering yet?

crates/api-ui/src/databases/handlers.rs
```diff
@@ -223,30 +218,52 @@ pub async fn update_database(
...
+    let sql_string = parameters.search.map_or_else(|| sql_string.clone(), |search|
+        format!("{sql_string} WHERE (database_name LIKE '%{search}%' OR volume_name LIKE '%{search}%' OR created_at LIKE '%{search}%' OR updated_at LIKE '%{search}%')")
```
rampage644: Why do we search in created_at and updated_at fields?

YaroslavLitvinov (approved):

YaroslavLitvinov (approved):
I would wait before merging this before getting additional approval, some things looks controversial and better to be fixed

YaroslavLitvinov (approved):
I would wait before merging this before getting additional approval, some things looks controversial and better to be fixed

crates/api-ui/src/databases/handlers.rs
DanCodedThis: @rampage644 just in case, we can remove this though, probably nobody would search by a timestamp

crates/api-ui/src/databases/handlers.rs
DanCodedThis: @nikitastryuk said we are not using this search yet, so it shouldn't affect anything for now

crates/api-ui/src/databases/handlers.rs
DanCodedThis: Agreed, thought of that too, decided not to add yet (pr to big)

crates/api-ui/src/databases/handlers.rs
DanCodedThis: @YaroslavLitvinov didn't understand you

crates/api-ui/src/databases/models.rs
DanCodedThis: @YaroslavLitvinov Yep, it's part of a bigger problem of that for volumes and databases both payload and response `data`: contains the same struct.

crates/api-ui/src/schemas/handlers.rs
DanCodedThis: @YaroslavLitvinov when we create a schema we actually create it in the DF context, which doesn't contain this info, it is then refreshed and added to the metastore, after which it could be retrieved both in the view and here

crates/api-ui/src/databases/models.rs
DanCodedThis: I tried changing it here, but since the PR is already big enough and that it's probably outside of this scope, I retracted the change (it's messy, needs more thought put into it)

crates/api-ui/src/schemas/handlers.rs
DanCodedThis: Before this change, we had a silent bug of showing different created_at and updated_at

crates/api-ui/src/databases/handlers.rs
osipovartem: Let's remove all unrelated search filters

osipovartem: Also add lower for search string (to be not case-sensetive)

crates/api-ui/src/databases/handlers.rs
```diff
@@ -194,20 +191,18 @@ pub async fn update_database(
...
     get,
     operation_id = "getDatabases",
     params(
-        ("cursor" = Option<String>, Query, description = "Databases cursor"),
+        ("offset" = Option<usize>, Query, description = "Databases offset"),
         ("limit" = Option<usize>, Query, description = "Databases limit"),
-        ("search" = Option<String>, Query, description = "Databases search (start with)"),
+        ("search" = Option<String>, Query, description = "Databases search"),
+        ("order_by" = Option<String>, Query, description = "Order by: database_name (default), volume_name, created_at, updated_at"),
+        ("order_direction" = Option<String>, Query, description = "Order direction: ASC, DESC (default)"),
```
osipovartem: Add to enum

crates/api-ui/src/databases/handlers.rs
DanCodedThis: @osipovartem Is `ILIKE` better here (case insensitive) or `lower` is preferred?

crates/api-ui/src/databases/handlers.rs
```diff
@@ -223,30 +218,52 @@ pub async fn update_database(
...
-            Json(DatabasesResponse {
-                items: o.into_iter().map(|x| x.data.into()).collect(),
-                current_cursor: parameters.cursor,
-                next_cursor,
-            })
-        })
+        .map_err(|e| DatabasesAPIError::List { source: e })?;
+    let mut items = Vec::new();
+    for record in records {
+        let database_names = record.column_by_name("database_name").unwrap().as_ref();
```
osipovartem: handle unwrap 
you can use downcast_string_column from navigation_tree/handlers

```rust
name: array_value_to_string(database_names, i).unwrap() 
```
-> 
```rust
name:  databases.value(i).to_string());
```

crates/api-ui/src/schemas/handlers.rs
```diff
@@ -228,9 +239,11 @@ pub async fn update_schema(
...
     params(
         ("databaseName" = String, description = "Database Name"),
-        ("cursor" = Option<String>, Query, description = "Schemas cursor"),
-        ("limit" = Option<usize>, Query, description = "Schemas limit"),
-        ("search" = Option<String>, Query, description = "Schemas search (start with)"),
+        ("offset" = Option<usize>, Query, description = "Schemas offset"),
+        ("limit" = Option<u16>, Query, description = "Schemas limit"),
+        ("search" = Option<String>, Query, description = "Schemas search"),
+        ("order_by" = Option<String>, Query, description = "Order by: schema_name (default), database_name, created_at, updated_at"),
+        ("order_direction" = Option<String>, Query, description = "Order direction: ASC, DESC (default)"),
```
osipovartem: Same, move to enum

crates/df-catalog/src/catalogs/slatedb/volumes.rs
```diff
@@ -24,6 +24,8 @@ impl VolumesView {
         let schema = Arc::new(Schema::new(vec![
             Field::new("volume_name", DataType::Utf8, false),
             Field::new("volume_type", DataType::Utf8, false),
+            Field::new("created_at", DataType::Utf8, false),
+            Field::new("updated_at", DataType::Utf8, false),
```
osipovartem, line 28: can be nullable

crates/api-ui/src/databases/handlers.rs
DanCodedThis: @osipovartem This is what is sent in the query path ex. `/ui/../databases?orderBy=database_name&orderDirection=DESC`, I'm not sure if it will work changing it to enum

crates/api-ui/src/databases/handlers.rs
osipovartem: Up to you

crates/api-ui/src/databases/handlers.rs
osipovartem: why not? maybe we need to add serialiser/deserialiser or request validation

crates/api-ui/src/schemas/handlers.rs
```diff
@@ -89,9 +89,19 @@ pub async fn create_schema(
         .query(&session_id, sql_string.as_str(), context)
         .await
         .map_err(|e| SchemasAPIError::Create { source: e })?;
-    Ok(Json(SchemaCreateResponse {
-        data: Schema::new(payload.name, database_name),
-    }))
+    let schema_ident = MetastoreSchemaIdent::new(database_name.clone(), payload.name.clone());
+    match state.metastore.get_schema(&schema_ident).await {
```
osipovartem, line 94: metastore don't save any external table
if 
```rust 
let _ = state
.execution_svc
.query(&session_id, sql_string.as_str(), context)
.await
.map_err(|e| SchemasAPIError::Create { source: e })?;
```
was successfull we should form response (not call metastore)

crates/df-catalog/src/catalogs/slatedb/databases.rs
```diff
@@ -62,13 +66,23 @@ pub struct DatabasesViewBuilder {
...
+    updated_at_timestamps: StringBuilder,
 }
 
 impl DatabasesViewBuilder {
-    pub fn add_database(&mut self, database_name: impl AsRef<str>, volume_name: impl AsRef<str>) {
+    pub fn add_database(
+        &mut self,
+        database_name: impl AsRef<str>,
+        volume_name: impl AsRef<str>,
+        created_at: impl AsRef<str>,
```
osipovartem, line 78:
```rust
created_at: impl AsRef<str>
 ->
created_at: Option<String>
```
Since it is nullable

crates/api-ui/src/schemas/handlers.rs
DanCodedThis: @osipovartem we return created_at and updated_at there, that's why I decided to call the metastore, before we returned UTC::now which was incorrect, since the time it was added was different

osipovartem (approved):

PR #845, opened 2025-05-19
Implement TRUNCATE TABLE

### LINKED ISSUE #843: [Discussion] Implement `TRUNCATE TABLE` statement
❗ Problem
Currently, the SQL command TRUNCATE TABLE is not supported in the system. This presents a usability gap, especially for users who need to quickly remove all data from a table without dropping and recreating it manually.
In standard SQL systems, TRUNCATE TABLE is a DDL operation that:
- Deletes all rows from a table very efficiently,
- Often bypasses row-by-row deletion,
- Preserves the table structure, schema, and metadata,
- May or may not be transactional depending on the storage engine.

📦 Target: Support for Iceberg Tables
In the context of Apache Iceberg, TRUNCATE TABLE does not have a built-in shortcut yet, but conceptually:
- Iceberg is designed to handle large-scale table modifications via snapshot-based, immutable data files.
- A TRUNCATE TABLE operation should ideally:
  - Remove all data files associated with the current snapshot,
  - Keep the table schema and metadata intact,
  - Create a new snapshot that reflects an empty table state.

✅ Proposed Implementation (Workaround)
```sql
CREATE OR REPLACE TABLE {object_name} AS (SELECT * FROM {object_name} WHERE FALSE)
```
✅ How it works:
- SELECT * FROM {table} WHERE FALSE returns no rows.
- CREATE OR REPLACE TABLE ... AS (...) replaces the existing table with a new, empty one.
- This effectively:
  - Deletes all data files,
  - Creates a new Iceberg snapshot with an empty state,
  - Keeps schema and properties.

🟡 Caveats:
- This removes all snapshots and history of the table (unless retained elsewhere).
- It does not preserve table-level metadata like custom table properties, unless you explicitly carry them over.
- This is semantically heavier than a real TRUNCATE in Iceberg’s design, which might ideally be a snapshot rewrite or metadata-only operation.

osipovartem:
- added truncate table statement as **create or replace** to fetch all necessary table metadata (or replace means drop existing table)
- fixed SHOW OBJECTS upper case for snowflake (internally we use only normalised names)
- fixed navigation tree VIEW type (during statement parsing we add the table itself to cache, so there was a duplicate) 

This fix allows to run snowplow **dbt seed** multiple times since there we truncate table if it is already exist
Closes #843

### DIFF
--- a/crates/core-executor/src/datafusion/planner.rs
+++ b/crates/core-executor/src/datafusion/planner.rs
@@ -100,8 +100,6 @@ where
                 let column_defaults = self
                     .inner
                     .build_column_defaults(&columns, planner_context)?;
-                // println!("column_defaults: {:?}", column_defaults);
-                // println!("statement 11: {:?}", statement);
                 let has_columns = !columns.is_empty();
                 let schema = self.build_schema(columns.clone())?.to_dfschema_ref()?;
                 if has_columns {
--- a/crates/core-executor/src/query.rs
+++ b/crates/core-executor/src/query.rs
@@ -41,7 +41,7 @@ use sqlparser::ast::helpers::attached_token::AttachedToken;
 use sqlparser::ast::{
     BinaryOperator, GroupByExpr, MergeAction, MergeClauseKind, MergeInsertKind, ObjectNamePart,
     ObjectType, PivotValueSource, Select, SelectItem, ShowObjects, ShowStatementFilter,
-    ShowStatementIn, Use, Value,
+    ShowStatementIn, TruncateTableTarget, Use, Value,
 };
 use std::collections::HashMap;
 use std::collections::hash_map::Entry;
@@ -287,6 +287,11 @@ impl UserQuery {
                 | Statement::ShowVariables { .. }
                 | Statement::ShowVariable { .. } => {
                     return Box::pin(self.show_query(*s)).await;
+                }
+                Statement::Truncate { table_names, .. } => {
+                    let result = Box::pin(self.truncate_table(table_names)).await;
+                    self.refresh_catalog().await?;
+                    return result;
                 }
                 Statement::Query(mut subquery) => {
                     self.update_qualify_in_query(subquery.as_mut());
@@ -1155,6 +1160,28 @@ impl UserQuery {
             }
         };
         Box::pin(self.execute_with_custom_plan(&query)).await
+    }
+
+    pub async fn truncate_table(
+        &self,
+        table_names: Vec<TruncateTableTarget>,
+    ) -> ExecutionResult<QueryResult> {
+        let Some(first_table) = table_names.into_iter().next() else {
+            return Err(ExecutionError::DataFusion {
+                source: DataFusionError::NotImplemented(
+                    "No table names provided for TRUNCATE TABLE".to_string(),
+                ),
+            });
+        };
+
+        let object_name = self.resolve_table_object_name(first_table.name.0)?;
+        let mut query = self.session.query(
+            format!(
+                "CREATE OR REPLACE TABLE {object_name} as (SELECT * FROM {object_name} WHERE FALSE)",
+            ),
+            QueryContext::default(),
+        );
+        query.execute().await
     }
 
     #[must_use]
--- a/crates/core-executor/src/tests/snapshots/query_truncate_table.snap
+++ b/crates/core-executor/src/tests/snapshots/query_truncate_table.snap
@@ -1,0 +1,14 @@
+---
+source: crates/core-executor/src/tests/query.rs
+description: "\"TRUNCATE TABLE employee_table\""
+snapshot_kind: text
+---
+Ok(
+    [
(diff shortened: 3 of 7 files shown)


crates/core-executor/src/query.rs
```diff
@@ -1122,6 +1127,28 @@ impl UserQuery {
...
+    ) -> ExecutionResult<Vec<RecordBatch>> {
+        let Some(first_table) = table_names.into_iter().next() else {
+            return Err(ExecutionError::DataFusion {
+                source: DataFusionError::NotImplemented(
+                    "No table names provided for TRUNCATE TABLE".to_string(),
+                ),
+            });
+        };
+
+        let object_name = self.resolve_table_object_name(first_table.name.0)?;
```
Vedin, line 1177: I think we should somehow check that such a table exists before trying to perform TRUNCATE with `CREATE OR REPLACE TABLE`. As I understood, it fails anyway because `(SELECT * FROM {object_name} WHERE FALSE)` is empty. I'd check it ourselves to give more descriptive behaviour. Also, add test for this.

crates/core-executor/src/tests/snapshots/query_truncate_table.snap
```diff
@@ -0,0 +1,14 @@
+---
+source: crates/core-executor/src/tests/query.rs
+description: "\"TRUNCATE TABLE employee_table\""
+snapshot_kind: text
+---
+Ok(
```
Vedin, line 6: Not a part of this PR but do you know about any plans dealing with correct statuses? For example: print `Statement executed successfully.` here instead of `count 0` and so on.

Vedin (changes requested):

crates/core-executor/src/tests/snapshots/query_truncate_table.snap
osipovartem: We have status in our snaps, Ok or Error

crates/core-executor/src/query.rs
osipovartem: Select will fail with missing table during planning. Will add tests
```
Error: DataFusion error: Error during planning: table 'embucket.public.missing_table' not found
```
for non-existing table `TRUNCATE TABLE missing_table`

crates/core-executor/src/datafusion/planner.rs
```diff
@@ -114,8 +114,6 @@ where
                 let column_defaults = self
                     .inner
                     .build_column_defaults(&columns, planner_context)?;
-                // println!("column_defaults: {:?}", column_defaults);
```
rampage644, line 103: nitpick: safe to delete I think

osipovartem: Delete what 😅

crates/core-executor/src/query.rs
```diff
@@ -1060,10 +1065,10 @@ impl UserQuery {
                 let sql = format!(
                     "SELECT
                         NULL as created_on,
-                        table_name as name,
+                        upper(table_name) as name,
```
rampage644: This seems like an unrelated change

crates/core-executor/src/query.rs
```diff
@@ -1122,6 +1127,28 @@ impl UserQuery {
...
+        let Some(first_table) = table_names.into_iter().next() else {
+            return Err(ExecutionError::DataFusion {
+                source: DataFusionError::NotImplemented(
+                    "No table names provided for TRUNCATE TABLE".to_string(),
+                ),
+            });
+        };
+
+        let object_name = self.resolve_table_object_name(first_table.name.0)?;
+        let mut query = self.session.query(
```
rampage644, line 1178: In general, I think we should eventually stop re-executing queries like this. 

It would be great to understand what is missing currently to implement TRUNCATE differently (without rewritting)? Is it logical node missing? Planning fails? If so, how exactly it fails, etc.

crates/core-executor/src/query.rs
osipovartem: This change is necessary to fix ShowObjects, related to dbt seed

rampage644:
It seems this PR solves single high-level problems, but involves several seemingly unrelated changes. I would like to see it split into several so those easier to review: i think I want more discussion about how to implement TRUNCATE but other changes seem totally legit and "lgtm".

Having everything in the same PR makes it harder to review

crates/core-executor/src/datafusion/planner.rs
rampage644: print statements

rampage644: Oh, they were... disregard

rampage644 (changes requested):
As mentioned in the comments, let's split changes that could be split and have bigger discussions about TRUNCATE implementation

rampage644 (approved):

rampage644:
@Vedin Anything to further remove?

PR #852, opened 2025-05-19
Tables search through the query engine

DanCodedThis:
Partial #522

- Making incremental prs per handler (easier to review and trying to minimize breaking changes to Yaroslav's api-structs pr)
- Extended the `slatedb.public.tables` view
- Changed the `/ui../tables` GET (list) handler to use query engine for advanced search
- Fixed & added tests

### DIFF
--- a/crates/api-ui/src/tables/handlers.rs
+++ b/crates/api-ui/src/tables/handlers.rs
@@ -7,8 +7,9 @@ use crate::tables::error::{
 use crate::tables::models::{
     Table, TableColumn, TableColumnsResponse, TablePreviewDataColumn, TablePreviewDataParameters,
     TablePreviewDataResponse, TablePreviewDataRow, TableStatistics, TableStatisticsResponse,
-    TableUploadPayload, TableUploadResponse, TablesParameters, TablesResponse, UploadParameters,
+    TableUploadPayload, TableUploadResponse, TablesResponse, UploadParameters,
 };
+use crate::{SearchParameters, downcast_int64_column, downcast_string_column};
 use api_sessions::DFSessionId;
 use axum::extract::Query;
 use axum::{
@@ -16,9 +17,8 @@ use axum::{
     extract::{Multipart, Path, State},
 };
 use core_executor::{models::QueryResultData, query::QueryContext};
+use core_metastore::TableIdent as MetastoreTableIdent;
 use core_metastore::error::MetastoreError;
-use core_metastore::{SchemaIdent as MetastoreSchemaIdent, TableIdent as MetastoreTableIdent};
-use core_utils::scan_iterator::ScanIterator;
 use datafusion::arrow::csv::reader::Format;
 use datafusion::arrow::util::display::array_value_to_string;
 use snafu::ResultExt;
@@ -48,7 +48,6 @@ use utoipa::OpenApi;
             TableUploadPayload,
             TableUploadResponse,
             TablesResponse,
-            TablesParameters,
             ErrorResponse,
         )
     ),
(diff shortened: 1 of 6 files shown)


DanCodedThis:
@rampage644 GitHub workflows Tests Check finished in 4 minutes, first time ever (before it was always 18 minutes+). Is this the caching?

crates/api-ui/src/tables/handlers.rs
```diff
@@ -378,62 +377,80 @@ pub async fn upload_file(
...
-        .token(parameters.search)
-        .collect()
+    let context = QueryContext::new(Some(database_name.clone()), None, None);
+    let sql_string = format!(
+        "SELECT * FROM slatedb.public.tables WHERE schema_name = '{}' AND database_name = '{}'",
+        schema_name.clone(),
+        database_name.clone()
+    );
+    let sql_string = parameters.search.map_or_else(|| sql_string.clone(), |search|
+        format!("{sql_string} AND (table_name ILIKE '%{search}%' OR schema_name ILIKE '%{search}%' OR database_name ILIKE '%{search}%' OR volume_name ILIKE '%{search}%' OR table_type ILIKE '%{search}%' OR table_format ILIKE '%{search}%' OR owner ILIKE '%{search}%')")
```
osipovartem: schema_name ILIKE '%{search}%' OR database_name ILIKE '%{search}
You already have initial filter in sql_string for such fields

crates/api-ui/src/tables/handlers.rs
```diff
@@ -378,62 +377,80 @@ pub async fn upload_file(
...
+    let context = QueryContext::new(Some(database_name.clone()), None, None);
+    let sql_string = format!(
+        "SELECT * FROM slatedb.public.tables WHERE schema_name = '{}' AND database_name = '{}'",
+        schema_name.clone(),
+        database_name.clone()
+    );
+    let sql_string = parameters.search.map_or_else(|| sql_string.clone(), |search|
+        format!("{sql_string} AND (table_name ILIKE '%{search}%' OR volume_name ILIKE '%{search}%' OR table_type ILIKE '%{search}%' OR table_format ILIKE '%{search}%' OR owner ILIKE '%{search}%')")
+    );
+    let sql_string = parameters.order_by.map_or_else(
```
osipovartem, line 394: I suggest to create some unified func to apply filters, ordering, offsets, limit and use it for all api queriers. 
Can be done in a separate PR.
Seems this code duplicated already multiple times

osipovartem (approved):

crates/api-ui/src/tables/handlers.rs
osipovartem: Like 
```rust
pub fn build_select_query(
    base_query: &str,
    search_columns: &[&str],
    parameters: &SearchParameters,
) -> String {
    let mut sql = base_query.to_string();

    // WHERE search
    if let Some(search) = &parameters.search {
        let search_filter = search_columns
            .iter()
            .map(|col| format!("{col} ILIKE '%{search}%'"))
            .collect::<Vec<_>>()
            .join(" OR ");
        sql = format!("{sql} AND ({search_filter})");
    }

    // ORDER BY
    sql = match &parameters.order_by {
        Some(order_by) => format!("{sql} ORDER BY {order_by}"),
        None => format!("{sql} ORDER BY {}", search_columns[0]), // default: first column
    };

    // ORDER direction
    sql = match &parameters.order_direction {
        Some(direction) => format!("{sql} {direction}"),
        None => format!("{sql} DESC"),
    };

    // OFFSET
    if let Some(offset) = &parameters.offset {
        sql = format!("{sql} OFFSET {offset}");
    }

    // LIMIT
    if let Some(limit) = &parameters.limit {
        sql = format!("{sql} LIMIT {limit}");
    }

    sql
}
let base_sql = format!(
    "SELECT * FROM slatedb.public.tables WHERE schema_name = '{}' AND database_name = '{}'",
    schema_name, database_name
);

let search_columns = &[
    "table_name",
    "volume_name",
    "table_type",
    "table_format",
    "owner",
];

let sql_string = build_select_query(&base_sql, search_columns, &parameters);
```

DanCodedThis: it's not really duplicated (there are subtle changes), but maybe we can make a general func for it

PR #855, opened 2025-05-20
Fix navigation tree view duplicates

osipovartem:
- Remove duplicates from navigation tree since selected query was already added to caching schema
- fixed view type


DanCodedThis (approved):
LGTM

Issue #851, opened 2025-05-19
api: POST /ui/databases, POST /ui/databases/${dbName}/schemas not working


nikitastryuk:
DB creation fails with 422:
```
Failed to deserialize the JSON body into the target type: missing field `created_at` at line 1 column 29
```

Schema creation fails with 500: 
```
Create schema error: DataFusion error: SQL error: ParserError(\"Expected: identifier, found: 3.1 at Line: 1, Column: 15\")
```

jatin510:
```{"name":"test","volume":"test", "created_at": "", "updated_at": ""}```

providing empty values in body payload, will work.
In frontend, we can provide empty string for these two fields: `created_at` and `updated_at`

DanCodedThis:
on the recent main, schema from UI is created successfully, database needs a new model similar to volumes in the #859 pr

DanCodedThis:
@nikitastryuk cc

Issue #850, opened 2025-05-19
api: UI Volumes return unintended info


DanCodedThis:
- Volume returns not only the type
- Volume returns credentials for S3 and S3Tables, endpoints, etc
- Volume returns extra metadata (File type for ex.)

Issue #848, opened 2025-05-19
api: UI Data modeling (refactor)


DanCodedThis:
- Response and payload models are the same
- For any post (create) request we can just send 200 OK and not resend info that was in the payload already

DanCodedThis:
cc @nikitastryuk

PR #859, opened 2025-05-20
Volumes search through query engine & removed unintended info

### LINKED ISSUE #522: api: UI Use Query engine for all list handlers (Metastore substitution)

DanCodedThis:
Closes #522 & #850

- `/ui/volumes` GET LIST now search though the query engine DF View
- Added `SimpleVolume`, our `Volume` struct which is used for both payload and response, when called GET or GET LIST was leaking credentials and other info - no more (`SimpleVolume` doesn't contain such info), related to #848 
- If you have a better name for `SimpleVolume` -  please (I will use such naming for related #848 & #851)
- Added `fn apply_other_parameters` to generalize search code (make it more concise)
- Added volumes tests for list handler and delete handler
- Rebased onto main (to add `fn apply_other_parameters` to the Tables search)

### DIFF
--- a/crates/api-ui/src/tables/handlers.rs
+++ b/crates/api-ui/src/tables/handlers.rs
@@ -9,7 +9,9 @@ use crate::tables::models::{
     TablePreviewDataResponse, TablePreviewDataRow, TableStatistics, TableStatisticsResponse,
     TableUploadPayload, TableUploadResponse, TablesResponse, UploadParameters,
 };
-use crate::{SearchParameters, downcast_int64_column, downcast_string_column};
+use crate::{
+    SearchParameters, apply_other_parameters, downcast_int64_column, downcast_string_column,
+};
 use api_sessions::DFSessionId;
 use axum::extract::Query;
 use axum::{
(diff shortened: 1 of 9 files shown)


crates/api-ui/src/volumes/models.rs
```diff
@@ -142,21 +141,31 @@ pub struct VolumeUpdateResponse {
...
+    pub data: SimpleVolume,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
 #[serde(rename_all = "camelCase")]
-pub struct VolumesResponse {
-    pub items: Vec<Volume>,
-    pub current_cursor: Option<String>,
-    pub next_cursor: String,
+pub struct SimpleVolume {
```
rampage644, line 149: ~~How is `SimpleVolume` different from `Volume`?~~ Nevermind, read the PR description

crates/api-ui/src/tables/handlers.rs
```diff
@@ -388,25 +390,10 @@ pub async fn get_tables(
...
-    );
-    let sql_string = parameters.offset.map_or_else(
-        || sql_string.clone(),
-        |offset| format!("{sql_string} OFFSET {offset}"),
-    );
-    let sql_string = parameters.limit.map_or_else(
-        || sql_string.clone(),
-        |limit| format!("{sql_string} LIMIT {limit}"),
-    );
+    let sql_string = apply_other_parameters(&sql_string, parameters, "table_name");
```
rampage644: This begins to look like builder pattern, how about applying it fully? Something like
```rust
let query = QueryBuilder::new()
    .with_database(database_name)
    .with_schema(schema_name)
    .with_table(table_name)
    .with_parameters(parameters)
    .build();
```

DanCodedThis: I was in the process of trying to make a builder. But since the db and schema are always `slatedb.public` and the search parameters need to have `search_columns` and `order_by_default` like this:
```
fn with_parameters(parameters: SearchParameters, search_columns: Vec<String>, order_by_default: String) -> SelectQueryBuilder
```
I don't think it's a good use of the builder pattern here. However, it gave me and idea to make search a part of applying the parameters.

osipovartem (approved):

Issue #714, opened 2025-05-07
PIVOT command


Vedin:
https://docs.snowflake.com/en/sql-reference/constructs/pivot

PR #868, opened 2025-05-21
PIVOT Command

Vedin:
I added subquery execution case here to complete solution with cases not covered in the DF code. 
https://github.com/Embucket/embucket/issues/714


crates/core-executor/src/tests/query.rs
```diff
@@ -639,13 +638,14 @@ macro_rules! test_query {
                     prepend_module_to_snapshot => false
                 }, {
                     let df = match res {
-                        Ok(mut record_batches) => {
+                        Ok(record_batches) => {
+                            let mut batches = record_batches;
```
rampage644, line 642: :+1:

crates/core-executor/src/query.rs
```diff
@@ -1676,169 +1840,6 @@ impl UserQuery {
         }
     }
 
-    // TODO: Modify this function to modify the statement in-place to
-    // avoid extra allocations
-    #[allow(clippy::too_many_lines)]
-    pub fn update_statement_references(
```
rampage644, line 1682: :+1:

Vedin: This is dead code.

crates/core-executor/src/query.rs
```diff
@@ -1604,6 +1606,168 @@ impl UserQuery {
         }
     }
 
+    // This function is used to update the table references in the query
+    // It executes the subquery and convert the result to a list of string expressions
+    // It then replaces the pivot value source with the list of string expressions
+    fn update_subquery_to_list_pivot<'a>(
```
rampage644: By any chance this can be simplified? What if we do this (or parts of it) during planning (i guess analyzer/optimizer rules won't suffice here?)? Will it be easier to manipulate logical operators/exprs vs. ast structs? 

Visitor pattern?

Just refactor some codepaths into functions?

rampage644: i.e. https://chatgpt.com/share/682e1bf1-feec-8011-af99-d90c2b3f2da8

Vedin: Unfortunately, it's impossible to do during planning. If it were possible, then I'd have done it in the datafusion PR itself. The idea is to imitate `execute immediate` here. So, I need to execute the subquery, and after that, change the original one. I updated the structure of this and made some methods that can be useful in future to replicate query update and traversal.

Vedin: > i.e. https://chatgpt.com/share/682e1bf1-feec-8011-af99-d90c2b3f2da8

i missed this comment but basically do mostly the same

crates/core-executor/src/tests/query.rs
Vedin: lol, fixed

rampage644: I meant I liked local scoped mutability update

crates/core-executor/src/query.rs
rampage644: Yeah - I understood the part to execute part of the query immediately, my comment was more about structuring the function a bit differently so it is easier to read.

rampage644 (approved):

ravlio (approved):

PR #872, opened 2025-05-21
feat: enhance CI workflow to include SLT tests and update README with…

### LINKED ISSUE #869: Update CI workflow to include SLT tests and update README

YevheniiNiestierov:
Closes #869


.github/workflows/tests.yml
```diff
@@ -56,5 +56,69 @@ jobs:
         uses: dtolnay/rust-toolchain@stable
         with:
           toolchain: ${{ matrix.toolchain }}
-      - name: cargo test
-        run: cargo test --workspace --all-features --all-targets
+
+#      - name: cargo test
```
rampage644: Why this is commented out?

.github/workflows/tests.yml
```diff
@@ -56,5 +56,69 @@ jobs:
...
+#        run: cargo test --workspace --all-features --all-targets
+
+  slt-tests:
+    needs: required
+    runs-on:
+      group: large-runners
+    name: SQL Language Tests
+    env:
+      SNOWFLAKE_ACCOUNT: test
```
rampage644: Does it means every CI/CD pipeline run connects to Snowflake?

.github/workflows/tests.yml
```diff
@@ -56,5 +56,69 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: true
+
+      - name: Build embucketd binary
+        run: cargo build
+
+      - name: Start Embucket Server
+        run: |
```
rampage644: Just curios what this action does?

rampage644 (changes requested):
My general feeling SLT tests should be decoupled from unit tests, artifact generation should also be decoupled

.github/workflows/tests.yml
YevheniiNiestierov: Tests were temporarily commented out to reduce execution time for troubleshooting purposes

.github/workflows/tests.yml
YevheniiNiestierov: This is related to how the SLT Runner library works, creds should always be there

YevheniiNiestierov: Will add comment to make this clear

.github/workflows/tests.yml
YevheniiNiestierov: Please ignore this — was just testing some environment behavior

andheroe (approved):

andheroe:
> My general feeling SLT tests should be decoupled from unit tests, artifact generation should also be decoupled

I've decoupled SLT from unit tests. Also tried decoupling artifact building - but it slows down CI/CD significantly. The process of uploading the build artifact and then downloading it to be reused takes around 7 minutes in total. The artifact is more than 3 GB in size.

rampage644:
@andheroe is the binary that big? Will `--release` help? `strip` afterwards?

andheroe:
@rampage644 It is that big. Will `--release` work for unit tests? What is `strip`?

andheroe:
`cargo build --release puts the resulting binary in target/release instead of target/debug . Compiling in debug mode is the default for development-- compilation time is shorter since the compiler doesn't do optimizations, but the code will run slower. Release mode takes longer to compile, but the code will run faster.`
why do you think `--release` will improve the time?

andheroe:
Currently, 2 jobs run in parallel. Unit tests (12 minutes) and SLT (11 minutes). In both jobs, Embucket is built independently.
In order to optimise this, we would need to build the binary once. It should be the binary with unit tests (as far as I understand, --release builds without tests). Then this binary should be uploaded and then downloaded and used in two independent jobs. Given all the information, I don't think this would be faster than the current setup.

andheroe:
@rampage644 what do you think?

rampage644 (commented):
## We want to ensure that the main branch remains stable and clean. This means no direct commits into main

Solution that calculates the result and updates it somewhere vs. than directly manipulating README file seems more simple and secure.

I investigated a bit how other projects handle such thing and it seems there are couple of strategies (mostly reliant on https://shields.io):

 * Generate a payload and store it in a github gist 
 * Generate a payload and store it in a special branch of the repository 
 * Send the payload directly to some service that will store it for us (mimic typical code coverage badges, need deeper investigation here).

The README is updated only **once** with something like `Compatibility`.

I've seen different "Dynamic Badges Action" in different projects with most popular seem to be `schneegans/dynamic-badges-action` and `RubbaBoy/BYOB`.

## Separate Github Actions workflow

Configure the workflow to trigger exclusively upon merges into main or even less frequently, such as a nightly schedule, depending on the duration and resource-intensity of our end-to-end tests. Avoid triggering it on every push or PR to prevent excessive usage and delays in our CI pipeline. 

This potentally frees us from the need to cache and share infrastructure with "per-PR-commit" workflow.

andheroe:
@rampage644 thanks for your feedback. It was addressed.

rampage644 (approved):

PR #873, opened 2025-05-21
UI Data modeling change (naming & payload != response)

### LINKED ISSUE #851: api: POST /ui/databases, POST /ui/databases/${dbName}/schemas not working

DanCodedThis:
Closes #851 & #848

- Implemented `From<>` trait for both `Metastore` structs and `RwObject<>`
- There are now `Timestamped` variants of the previous structs (`SimpleVolume` renamed) for schemas, tables & databases
- As was discussed, we decided to change only the get and list handlers to return such structs
- Create & Update (which is not used right now) is unchanged

### DIFF
--- a/crates/api-ui/src/auth/handlers.rs
+++ b/crates/api-ui/src/auth/handlers.rs
@@ -175,6 +175,7 @@ pub struct ApiDoc;
 )]
 #[tracing::instrument(level = Level::ERROR, skip_all, err)]
 pub async fn login(
+    //TODO: add DFSessionId (to start the session on login)
     State(state): State<AppState>,
     Json(LoginPayload { username, password }): Json<LoginPayload>,
 ) -> AuthResult<impl IntoResponse> {
--- a/crates/api-ui/src/databases/handlers.rs
+++ b/crates/api-ui/src/databases/handlers.rs
@@ -5,7 +5,7 @@ use crate::{
     databases::error::{DatabasesAPIError, DatabasesResult},
     databases::models::{
         Database, DatabaseCreatePayload, DatabaseCreateResponse, DatabaseResponse,
-        DatabaseUpdatePayload, DatabaseUpdateResponse, DatabasesResponse,
+        DatabaseUpdatePayload, DatabaseUpdateResponse, DatabasesResponse, TimestampedDatabase,
     },
     downcast_string_column,
     error::ErrorResponse,
@@ -38,6 +38,7 @@ use validator::Validate;
             DatabaseResponse,
             DatabasesResponse,
             Database,
+            TimestampedDatabase,
             ErrorResponse,
         )
     ),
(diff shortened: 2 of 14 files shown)


YaroslavLitvinov (approved):
I would expect that:
* all schemas responses include TimestampedSchema
* all database responses include TimestampedDatabase

DanCodedThis:
> I would expect that:
> 
> * all schemas responses include TimestampedSchema
> * all database responses include TimestampedDatabase

@YaroslavLitvinov So create and update handler responses also, but only for databases and schemas (volumes - untouched, tables - nothing to touch)?

YaroslavLitvinov:
> > I would expect that:
> > 
> > * all schemas responses include TimestampedSchema
> > * all database responses include TimestampedDatabase
> 
> @YaroslavLitvinov So create and update handler responses also, but only for databases and schemas (volumes - untouched, tables - nothing to touch)?

I mean applying the principle - Responses always return timestamps, when request payload uses generic Database/Schema struct w/o timestamps. Regarding Tables - sure if you think such principle is applied to them.

rampage644 (commented):
I was curios if this specific struct can be expicilty named like with `payload`, `list`, `get` words in it? It looks - I might be very wrong - this structure is only needed for a specific get/list responses, and only as part of it. (i.e.  for`DatabaseGetListResponse`).

Maybe some expresiveness won't hurt? `ListResponseDatabase`, etc. Also not great looking, but what if?
Some AI suggested namings:
```
VolumeListItem - Clearly indicates it's an item in a list response
VolumeResponsePayload - Explicitly states it's a payload for responses
VolumeApiResponse - Directly references its use in API responses
VolumeGetResponseItem - Specifies it's for GET response items
VolumeListResponseEntry - Very explicit about its purpose in list responses
```

DanCodedThis:
> I was curios if this specific struct can be expicilty named like with `payload`, `list`, `get` words in it? It looks - I might be very wrong - this structure is only needed for a specific get/list responses, and only as part of it. (i.e. for`DatabaseGetListResponse`).
> 
> Maybe some expresiveness won't hurt? `ListResponseDatabase`, etc. Also not great looking, but what if? Some AI suggested namings:
> 
> ```
> VolumeListItem - Clearly indicates it's an item in a list response
> VolumeResponsePayload - Explicitly states it's a payload for responses
> VolumeApiResponse - Directly references its use in API responses
> VolumeGetResponseItem - Specifies it's for GET response items
> VolumeListResponseEntry - Very explicit about its purpose in list responses
> ```

I think this won't be necessary, since Yaroslav pointed out that it's a good idea that we return the `Timestamped` variant always from all request to further payload != response. Now all responses will be `Timestamped` (still probably need a better name, but for no it's ok I think)

osipovartem (approved):

nikitastryuk (approved):

nikitastryuk (approved):

PR #878, opened 2025-05-22
updated the metastore trait

### LINKED ISSUE #867: Seperate name for the volume/database then Volume/Database struct
For the trait: 

```rust 
#[async_trait]
pub trait Metastore: std::fmt::Debug + Send + Sync {
}
```

Functions like: have a name identifier and a seperate `ident` in Volume/Database

```rust
    async fn create_volume(
        &self,
        name: &VolumeIdent,
        volume: Volume,
    ) -> MetastoreResult<RwObject<Volume>>;

    async fn update_volume(
        &self,
        name: &VolumeIdent,
        volume: Volume,
    ) -> MetastoreResult<RwObject<Volume>>;

async fn create_database(
        &self,
        name: &DatabaseIdent,
        database: Database,
    ) -> MetastoreResult<RwObject<Database>>;

  async fn update_database(
        &self,
        name: &DatabaseIdent,
        database: Database,
    ) -> MetastoreResult<RwObject<Database>>;
```

Is there any particular reason, why the `name` can be different from the Database/Volume `ident` field ?
- can we merge them, into one ?
- can  `create databse/volume`,  get the name from the Database/Volume `ident` key ?

jatin510:
Issue: https://github.com/Embucket/embucket/issues/867

This will help us in accidentally having different name for volume/database ident and name.


rampage644 (approved):

osipovartem (approved):

rampage644:
I also dig into the failing test (one that pass if panics with correct error message) and it break one of the UI endpoint:
```
thread 'tests::databases::test_ui_databases_metastore_update_bug' panicked at crates/api-ui/src/tests/databases.rs:76:5:
assertion `left == right` failed
  left: 200
 right: 500
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
note: panic did not contain expected string
      panic message: `"assertion `left == right` failed\n  left: 200\n right: 500"`,
 expected substring: `"Failed to get error response: reqwest::Error { kind: Decode, source: Error(\"missing field `message`\", line: 1, column: 122) }"`
 ```

The failure happens on `assert_eq!(http::StatusCode::OK, res.status());` line that was passing previously.

PR #841, opened 2025-05-19
BLOCKED Issues/808 seed demo data on server startup

YaroslavLitvinov:
### When it's done it will close #808, #844

Note: _Work on PR stopped, since it touches architectural aspects that need to be discussed before resuming work._

* api-seed crate: Seeds fake entities into embucket if enabled with cmd line arg `--seed-data-variant`.  Planned to create a separate executable for doing the same.
* api-structs crate: Structs traits decoupling and refactoring, split on derive macros level. 
* api-client-rest crate (Rust SDK): 
  - High level operations on DB, create volumes, create schemas, (no tables and table data yet)
  - *requests* generic nested module supports embucket authentication, run queries with session support, generic rest calls
* Tests OK

PR #881, opened 2025-05-22
add seed executable

### LINKED ISSUE #808: api: Ability to seed demo data on server startup
Implement automatic creation of a default in-memory volume with database and schema on ~server startup~ (maybe some tables as well). 
Update: **Using a separate executable.** as per concerns raised at previous PR review.

This will:
- Speed up development / QA.
- Improve onboarding and demo UX.

I suggest to control enabling/disabling seed with an environment variable.

Another issue about seeding:
https://github.com/Embucket/embucket/issues/475

YaroslavLitvinov:
Closes #808 

This PR was added instead of blocked and now closed PR #841 
It is based on features added in PR #841, addressing concerns raised on PR review:
* Crates added in old PR are eliminated, meaning proposed approach linked to a `api-structs` crate left in old PR and didn't move here 
* Seeding feature moved from embucket executable and now only available via a separate `embucket-seed` executable.
* No `api-structs`, `api-client-rest`, `api-seed` available anymore, instead got a single crate `embucket-seed`.
* No seeding tables data, as it wasn't a requirement

### Dependencies
Below illustrated dependencies for both old PR - v1 , new PR - v2
which means v2 now has dependencies on monolithic crates 

* v1 PR #841
[dependencies]
api-client-rest = { path = "../api-client-rest" }
api-structs = { path = "../api-structs", features = ["serde", "partial", "eq"] } 
...

* v2 
[dependencies]
core-metastore = { path = "../core-metastore" }
api-ui = { path = "../api-ui", features = ["client"] }
...


crates/embucket-seed/src/tests.rs
```diff
@@ -0,0 +1,141 @@
...
+
+    // Save output of ^^ this to typical_seed.yaml when changing code ^^
+
+    let yaml_serialized =
+        serde_yaml::to_string(&seed_root).expect("Failed to serialize seed template");
+
+    eprintln!(
+        "programmatically created typical seed template: \n{}",
+        yaml_serialized
+    );
```
DanCodedThis: nitpick: why the `eprintln` (error println) and not the normal `println`?

DanCodedThis (approved):
LGTM

crates/embucket-seed/src/tests.rs
YaroslavLitvinov: It's just for tests, not makes any difference here, stderr vs stdout

crates/embucket-seed/src/external_models.rs
```diff
@@ -0,0 +1,15 @@
+#[cfg(test)]
```
rampage644, line 3: Am I correct this applies only to the next statement - just first type declaration?

YaroslavLitvinov: yep, FileVolume is used only in tests, optionally could be marked as dead code or to removed from tests

rampage644 (approved):
I thought I already approved this

PR #884, opened 2025-05-22
Update clippy workflow. Current clippy fixes

Vedin:
Original PR: I noticed we don't fail on Clippy warnings and even errors. Just verify this for now with this PR. 
Updated PR: 
1. I updated clippy workflow as well as check everything with rust and cargo 1.87.0. 
2. Fixed all clippy errors that exist because of the non-working check. 
3. First of all some changes have fixed the Clippy result-large-err warnings, ensuring that large error types are boxed to avoid stack allocations. It should slightly reduce build time as well. 
4. Other clippy errors were some minor fixes of not having `must_use` where needed and using it where not needed. 
5. Removed dead code from query.rs.


crates/api-ui/src/databases/error.rs
```diff
@@ -24,6 +24,18 @@ pub enum DatabasesAPIError {
     List { source: ExecutionError },
 }
 
+impl From<Box<MetastoreError>> for DatabasesAPIError {
+    fn from(boxed_error: Box<MetastoreError>) -> Self {
+        let error = *boxed_error;
+        match error {
```
YaroslavLitvinov: If I correctly understand, instead of this we could just use snafu selector in corresponding handlers: 
`.context(GetSnafu)` , `.context(CreateSnafu)`

rampage644 (approved):
Overall lgtm, though 

> First of all some changes have fixed the Clippy result-large-err warnings, ensuring that large error types are boxed to avoid stack allocations. It should slightly reduce build time as well.

That changes look a bit scary, I still hope we will improve error handling still, so okay to move forward

YaroslavLitvinov (approved):
In overall it's a huge work, and at least api-ui part of code needs a refactoring. But here I wanted mention introduced From conversions into error classes that could be replaced by snafu selectors.

YaroslavLitvinov (approved):

PR #889, opened 2025-05-22
remove build.rs from api-ui-static-assets

### LINKED ISSUE #887: Replace /ui/dist folder by /ui/dit.tar
### Problem
Currently, the Embucket server serves its Web UI using committed artifacts from the /ui/dist directory. To serve the UI, the server's build script creates a .tar archive from the files in /ui/dist. This intermediate step of generating a tarball from committed artifacts could be replaced by directly committing a pre-built /ui/dist.tar.

Following a discussion with @nikitastryuk , it was proposed to commit /ui/dist.tar instead of the raw /ui/dist files. This approach would reduce noise in pull requests, as only a single binary artifact would appear, instead of the numerous files currently present.

### Changes on the UI Side
The generated frontend artifacts are now placed into /ui/dist.tar and committed. As a result, the /ui/dist directory can be removed from the repository.

### Changes on the Server Side
The build script responsible for creating the tar archive is removed. Instead, the pre-existing /ui/dist.tar (already committed) will be embedded into the binary in the same way as before.

YaroslavLitvinov:
Closes #887


rampage644 (approved):

YaroslavLitvinov:
@dependabot rebase

PR #894, opened 2025-05-22
feat: add parallel execution option for SQL logic tests, update readme

### LINKED ISSUE #893: Add parallelism option to SLT runner

YevheniiNiestierov:
Closes #893


andheroe (changes requested):
So only cactegories run in parallel, right? Why don't just have a pool of workers and send  all jobs to parallel workers? If there are significantly more tests in one category versus another (we have such cases) - then the parallelization could be very suboptimal.

rampage644 (approved):

andheroe (approved):

PR #898, opened 2025-05-23
Add QueryResult struct with Schema

### LINKED ISSUE #897: Add QueryResult struct with Schema
- Add new QueryResult as return entity for execution service trait
```rust
pub struct QueryResult {
    pub records: Vec<RecordBatch>,
    /// The schema associated with the result.
    /// This is required to construct a valid response even when `records` are empty
    pub schema: Arc<ArrowSchema>,
    pub query_id: i64,
}
```
- Move convert batches logic to snowflake rest api (this is only related to DBT results)
- ExecutionService:
  - remove useless fallback to check empty columns info since we always have schema now in QueryResult (this logic treggers a bug when we tried to get schema from **RAW sql query**, not modified by our logic)
  - Remove **DBT only related config** from execution service (move to snowflake rest api)
 
Related to https://github.com/Embucket/embucket/issues/729

osipovartem:
- Added new QueryResult as return entity for execution service trait
```rust
pub struct QueryResult {
    pub records: Vec<RecordBatch>,
    /// The schema associated with the result.
    /// This is required to construct a valid response even when `records` are empty
    pub schema: Arc<ArrowSchema>,
    pub query_id: i64,
}
```
- Moved convert batches logic to snowflake rest api (this is only related to DBT results)
- ExecutionService:
  - removed useless fallback to check empty columns info since we always have schema now in QueryResult (this logic treggers a bug when we tried to get schema from **RAW sql query**, not modified by our logic)
  - Removed **DBT only related config** from execution service (moved to snowflake rest api)
 
❗❗❗ **This PR not introducing anything new, just moved unrelated logic to correct crates.** ❗❗❗

Related to https://github.com/Embucket/embucket/issues/729
Closes #897


crates/core-history/src/recording_service.rs
```diff
@@ -206,14 +209,12 @@ mod tests {
...
-                dbt_serialization_format: DataSerializationFormat::Json,
-            },
-        ));
-        let execution_svc =
-            RecordingExecutionService::new(execution_svc.clone(), history_store.clone());
+        let execution_svc = Arc::new(CoreExecutionService::new(metastore.clone()));
+        let execution_svc = RecordingExecutionService::new(
+            execution_svc.clone(),
+            history_store.clone(),
+            DataSerializationFormat::Json,
```
DanCodedThis, line 216: Is there a better way to add this? Perhaps into the `CoreExecutionService` or is it specific to the history?

DanCodedThis (approved):
Overall LGTM

crates/core-history/src/recording_service.rs
osipovartem: We also need to remove this from history, data format is required for dbt snowflake resp API, but we also use it to save results in format in Recording service. We don't need it in ExecutionService for sure

crates/core-executor/src/models.rs
```diff
@@ -1,16 +1,40 @@
...
+    }
+    #[must_use]
+    pub const fn with_query_id(mut self, new_id: i64) -> Self {
+        self.query_id = new_id;
+        self
+    }
+
+    #[must_use]
+    pub fn column_info(&self) -> Vec<ColumnInfo> {
+        ColumnInfo::from_batch(self.schema.clone())
```
YaroslavLitvinov, line 34: Could it work without .clone() if ColumnInfo::from_batch would expect a reference?

crates/core-executor/src/utils.rs
```diff
@@ -128,13 +116,13 @@ pub fn first_non_empty_type(union_array: &UnionArray) -> Option<(DataType, Array
 }
 
 pub fn convert_record_batches(
```
YaroslavLitvinov, line 118: Just curious if converted Vec<RecordBatch> was returned as WithConversion<Vec<RecordBatch>> by convert_record_batches?

YaroslavLitvinov (approved):

crates/core-executor/src/models.rs
osipovartem: self.schema.clone() means Arc::clone(self.schema)

crates/core-executor/src/utils.rs
osipovartem: yes, but just as inner records field

PR #904, opened 2025-05-24
Optimize read lock scope in query function

jatin510:
Scopes the read lock on df_sessions to only the session lookup operation, releasing it immediately after getting the user session. This reduces lock contention and improves concurrency by minimizing lock hold time.

As the subsequent steps (query execution and record batch processing) can take significant time to complete, releasing the lock early allows other operations to access the sessions map while these long-running operations are in progress.


osipovartem (approved):

DanCodedThis (approved):
LGTM

YaroslavLitvinov (approved):
Looks good

PR #909, opened 2025-05-26
`get_path` scalar function

ravlio:


crates/df-builtins/src/get_path.rs
```diff
@@ -0,0 +1,188 @@
...
+        };
+        let ColumnarValue::Scalar(ScalarValue::Utf8(Some(path))) = args[1].clone() else {
+            return exec_err!(
+                "get_path function requires the second argument to be a scalar string"
+            );
+        };
+
+        let input = as_string_array(&arr);
+        let mut res = StringBuilder::new();
+        for v in input {
```
rampage644, line 64: nitpick (streamline to eliminate error cases early):
```rust
let input = as_string_array(&arr);
let mut res = StringBuilder::new();
for v in input {
    // Early exit for None
    let Some(v) = v else {
        res.append_null();
        continue;
    };

    // Early exit for JSON parse error
    let Ok(json_value) = serde_json::from_str::<Value>(v) else {
        res.append_null();
        continue;
    };

    // Early exit for invalid path/tokenize or missing value
    let Some(tokenized) = json::tokenize_path(&path)
        .and_then(|v| json::get_json_value(&json_value, &v))
    else {
        res.append_null();
        continue;
    };

    // Happy path: serialize and append
    res.append_value(serde_json::to_string_pretty(&tokenized).map_err(|e| {
        DataFusionError::Internal(format!("Failed to serialize JSON value: {e}"))
    })?);
}
```

rampage644 (approved):

crates/df-builtins/src/get_path.rs
ravlio: Agree, thanks. let Some(...) else is a such powerful thing

osipovartem (approved):

PR #911, opened 2025-05-26
Insert into table issue

### LINKED ISSUE #905: Insert into table error
Image

Keeping only necessary data in curl: 
```
curl 'http://localhost:3000/ui/queries' \
  --data-raw $'{"query":"insert into test (name) values (\'jatin\')","worksheetId":1748112553825,"context":{"database":"test_database","schema":"public"}}'
```
Errors : 
```
{
    "message": "Query execution error: DataFusion error: Error during planning: table 'embucket.public.test' not found",
    "status_code": 422
}
```

jatin510:
Issue: https://github.com/Embucket/embucket/issues/905


rampage644:
@jatin510 Could you please elaborate on what was fixed and what was root cause for the problem? I am worried the issue might be more foundational.

jatin510:
I updated the `Statement` which was generated by `sql_to_statement` from df crate.

The statement generated only have info about `table_name`.
Then I converted `TableReference` from `Bare` to `Full`, which gives the full path `catalog.schema.table`

and assigned `new table_ident` generated by function `resolve_table_object_name` to the `Statement`.

rampage644:
Thank you for the explanation! 

In the issue linked `INSERT INTO` statement without fully qualified name (catalog.schema.table) identifier present (just a table name) fails with "Table not found" error. The error though is "Table embucket.public.test not found". It seems context database and schema wasn't properly selected.

My question though is "why other queries work"?  Do they actually work? Having some tests for this issue - that reproduce original bug - would also help here.

jatin510:
Actually, other queries, 
like, creating table, drop table, etc. calls a function `resolve_table_object_name` explicitly to resolve the full qualified name for the table.

I can see, even the `select * from table` failing with same error `embucket.public.test_db` not found.

I will setup some tests, to highlight the issue related to `full qualified name` resolution . @rampage644

osipovartem:
This commit removed update_statement_references funcionality which previously checked for fully qualified names in tableRef
https://github.com/Embucket/embucket/commit/6823571b826ddfc986c5a182e6759f7d67c74375#diff-3085a4b172528640e1310fcb901a30dd382db1fcc3fd3a2148bb3d7ac05ce691

Initially it was removed here from custom_logical_plan
https://github.com/Embucket/embucket/commit/8363e0b586fe18222890b63dd0b45b0df2ee7cd4#diff-47593f63b91f01d347b1c023fc2b7b82f8f4703ba72e9efd266460f32c2b0203

```rust
let mut statement = state
    .sql_to_statement(query, dialect)
statement = self.update_statement_references(statement)?;
```

jatin510:
@osipovartem  @rampage644 
we need the functionality of :
`update_statement_references` function

As I see, currently `Insert` statements are also failing.
We need to make sure our tests are able to detect these issues.

rampage644:
@osipovartem Thanks for the input.

@jatin510 This is exactly what I've meant - the problem is not with specific INSERT INTO statement but rather one that affects others too. 

From my perspective the way forward is to make sure the problem could be detected with tests (`core-executor/tests/query.rs` or perhaps another one to test for proper context) + restore functionality that was removed by mistake in previous PRs.

jatin510:
I would love to work on setting up test for the `core-executor` crate.  @rampage644 
Shall I create an issue for it and work on it ?

rampage644:
@jatin510 I think it's sufficient to update #905 to include all the context, then test suite can be extended and the problem fixed either in this PR or another (i have no preference here).

jatin510:
Solved in PR: https://github.com/Embucket/embucket/pull/996

PR #944, opened 2025-05-28
open api spec changes & api-ui error refactoring

YaroslavLitvinov:
* Optimize open api spec
* Refactor snafu error in ui handlers


crates/api-ui/src/tests/dashboard.rs
```diff
@@ -25,57 +23,39 @@ async fn test_ui_dashboard() {
...
     assert_eq!(http::StatusCode::OK, res.status());
     let dashboard: DashboardResponse = res.json().await.unwrap();
-    assert_eq!(0, dashboard.data.total_databases);
-    assert_eq!(0, dashboard.data.total_schemas);
-    assert_eq!(0, dashboard.data.total_tables);
-    assert_eq!(0, dashboard.data.total_queries);
+    assert_eq!(0, dashboard.0.total_databases);
+    assert_eq!(0, dashboard.0.total_schemas);
+    assert_eq!(0, dashboard.0.total_tables);
+    assert_eq!(0, dashboard.0.total_queries);
```
DanCodedThis: This (and all others) can be simplified to:

```rust
    let DashboardResponse(dashboard) = res.json().await.unwrap();
    assert_eq!(0, dashboard.total_databases);
    assert_eq!(0, dashboard.total_schemas);
    assert_eq!(0, dashboard.total_tables);
    assert_eq!(0, dashboard.total_queries);
```

crates/api-ui/src/volumes/handlers.rs
```diff
@@ -2,11 +2,10 @@ use crate::state::AppState;
 use crate::{
     OrderDirection, SearchParameters, apply_parameters, downcast_string_column,
     error::ErrorResponse,
-    volumes::error::{VolumesAPIError, VolumesResult},
+    volumes::error::{CreateSnafu, DeleteSnafu, GetSnafu, ListSnafu, VolumesResult},
```
DanCodedThis, line 5: This can be done in a separate PR, but we can "bring a common denominator" to the `Result` type. Currently, we have too many, and they only exist to wrap an enum of errors that can be generalized. I wrote more on this in the Sergei's error proposal.

DanCodedThis (approved):
A little bit cleanlier to work with errors, a welcome change, LGTM

crates/api-ui/src/dashboard/handlers.rs
```diff
@@ -51,9 +50,9 @@ pub async fn get_dashboard(
         .iter_databases()
         .collect()
         .await
-        .map_err(|e| DashboardAPIError::Metastore {
-            source: MetastoreError::UtilSlateDB { source: e },
-        })?;
+        .context(UtilSlateDBSnafu)
```
rampage644, line 53: I've already shared my initial feeling towards `map_err`, so second that. However, I still feel double `.context` can (should?) be replaced with better context selector like `GetDashboardListDatabasesSnafu` or `GetDashboardSnafu { message: "Failed to list databases" }`

Two following contexts are to be extended in the same way: using concrete error enum variant or common enum variant with some context about what it was trying to do.

crates/api-ui/src/databases/handlers.rs
```diff
@@ -116,15 +123,22 @@ pub async fn get_database(
...
+                Box::new(MetastoreError::DatabaseNotFound {
+                    db: database_name.clone(),
+                })
+            })
+        })
+        .context(GetSnafu)?
+        .map(Database::from)
+        .map(DatabaseResponse)
+        .map(Json)
+        .context(GetSnafu)
```
rampage644, line 142: Can you explain why there is second `.context` here?

crates/api-ui/src/databases/handlers.rs
```diff
@@ -116,15 +123,22 @@ pub async fn get_database(
...
-        ),
-        Err(e) => Err(e).context(GetSnafu),
-    }
+    state
+        .metastore
+        .get_database(&database_name)
+        .await
+        .map(|opt_rw_obj| {
+            opt_rw_obj.ok_or_else(|| {
+                Box::new(MetastoreError::DatabaseNotFound {
```
rampage644, line 133:
```rust
DatabaseNotFoundSnafu {
  db: database_name.clone(),
}
.build()/fail()
```

rampage644: Both option work until error enum variants are extended to include extra info, that snafu deals with automatically/implicitly (source. backtrace, location, other fields marked with `#[snafu(implicit)]`

crates/api-ui/src/databases/handlers.rs
```diff
@@ -186,17 +200,24 @@ pub async fn update_database(
...
-        source: MetastoreError::Validation { source: e },
-    })?;
+    let database = MetastoreDatabase {
+        ident: database.name,
+        volume: database.volume,
+        properties: None,
+    };
+    database
+        .validate()
+        .map_err(|e| Box::new(MetastoreError::Validation { source: e }))
```
rampage644: possible to replace with `.context()`?

rampage644: Is it because of boxing?

crates/api-ui/src/databases/handlers.rs
YaroslavLitvinov: `.map(|r| r.ok_or_else)` is returning `Result<> `
which need to be handled to avoid nesting  `Result<Result<>>`  that is why we call `.context(GetSnafu)?` to get a fully compatible result from that nested result, so it can be returned by a handler function

crates/api-ui/src/schemas/error.rs
```diff
@@ -15,26 +15,15 @@ pub enum SchemasAPIError {
     #[snafu(display("Create schema error: {source}"))]
     Create { source: ExecutionError },
     #[snafu(display("Get schema error: {source}"))]
-    Get { source: MetastoreError },
+    Get { source: Box<MetastoreError> },
```
rampage644, line 18: Is it due to clippy?

crates/api-ui/src/dashboard/handlers.rs
YaroslavLitvinov: Exactly that piece of code you marked, is actually handling a Metastore error. As empty option is not a part of business logic of Dashboard, it's just a thing from the Metastore errors "design".  If metastore would return some error instead of empty option we would use that error without introducing anything more.

Dashboard is not best example as it simple enough and don't have additional context, since it were not added, 
but I would say `DatabasesAPIError` is a thing which introduced specifically for context.

rampage644 (approved):
lgtm, I reviewed mostly the `api-ui error` part. Next time, I would appreciate this is actually split into two PRs

crates/api-ui/src/dashboard/handlers.rs
rampage644: > If metastore would return some error instead of empty option we would use that error without introducing anything more.

I was doing some error refactoring and came to the same conclusion - that it would be convenient and handy to handle "not found" case with an `Result` rather than `Option`: it streamlines the case with just reporing/propogation while keeping "let's deal with not found" case almost the same.

crates/api-ui/src/databases/handlers.rs
YaroslavLitvinov: Yes sure, it's need  to prepared before using a data types conversion

YaroslavLitvinov:
```
    database
        .validate()
        .context(ValidationSnafu)
        .map_err(Into::into)
        .context(UpdateSnafu)?;
```

crates/api-ui/src/schemas/error.rs
YaroslavLitvinov: No, I've just wanted to remove Box conversion from several recurring places, Adding Box only here is inconsistent, but since we have no rules regarding using Box everywhere i decided that is also ok.

DanCodedThis (approved):

Issue #939, opened 2025-05-28
[DISCUSSION] Introducing a History Trait as Part of UserSession / UserQuery


osipovartem:
✨ **Motivation**

The proposal is to integrate a history: Arc<dyn HistoryStore> (WorksheetStore) field into the UserSession structure to enable first-class support for per-session query history. This will mirror behavior found in platforms like Snowflake, where functions like [LAST_QUERY_ID()](https://docs.snowflake.com/en/sql-reference/functions/last_query_id) return query metadata for recently executed statements.
```rust
pub struct UserSession {
    pub metastore: Arc<dyn Metastore>,
    pub ctx: SessionContext,
    pub ident_normalizer: IdentNormalizer,
    pub executor: DedicatedExecutor,
    pub history: Arc<dyn HistoryStore>, // ← new field
}
```
The HistoryStore will be a trait responsible for recording and accessing query execution history within a session.

---

**🔍 Problem with Current Architecture**

Currently, history recording is handled in core-history via a custom implementation of the ExecutionService:
```rust
impl ExecutionService for RecordingExecutionService {
    async fn query(...) { ... }
}
```

This approach leads to duplication and architectural coupling that can be avoided. It also limits query history visibility strictly to the external recording service layer, rather than integrating it into the session-level context where it naturally belongs.
___

**✅ Proposed Solution**

Remove the existing RecordingExecutionService impl with a cleaner and unified CoreExecutionService implementation that records query history via the new History trait:

```rust
pub struct CoreExecutionService {
    metastore: Arc<dyn Metastore>,
    df_sessions: Arc<RwLock<HashMap<String, Arc<UserSession>>>>,
}

#[async_trait::async_trait]
impl ExecutionService for CoreExecutionService {
    async fn query(...) {
        // execute query
        // record query in session.history
    }
}
```
By making history part of the UserSession, we gain several benefits:
	•	History recording becomes mandatory and consistent across all queries.
	•	Query history (including last_query_id() support) is available at runtime inside SQL query processing.
	•	We avoid unnecessary indirection and implementation overhead in core-history.

___
**🧩 Benefits**
	•	✅ Cleaner architecture with centralized session state.
	•	✅ Full support for SQL functions like LAST_QUERY_ID().
	•	✅ Removes boilerplate from RecordingExecutionService.
	•	✅ Enables more extensible and testable history handling.

___
**🔄 Next Steps**
[...]

Issue #866, opened 2025-05-21
[WIP] RFC: Error handling in `embucket`


rampage644:
**Disclaimer: I am still working on it, publishing to save progress**

## Summary

This RFC proposes a standardized approach to error handling across the `embucket` codebase to improve code readability, maintainability, and developer experience.

### Use the `snafu` crate for error handling

While `embucket` already uses `snafu` in many places, we should standardize its usage across the entire codebase:

```rust
use snafu::prelude::*;

#[derive(Debug, Snafu)]
#[snafu(visibility(pub(crate)))]
pub enum MyError {
    #[snafu(display("Failed to open file: {path}"))]
    FileOpen { 
        source: std::io::Error,
        path: String,
        backtrace: Backtrace,
    },
    
    #[snafu(display("Invalid configuration: {message}"))]
    InvalidConfig { 
        message: String,
        backtrace: Backtrace,
    },
}
```

### Prefer flat error types hierarchy <--- THIS IS VERY DEBATABLE!!!

Rather than deeply nested error types, prefer a flatter hierarchy:

**Instead of this**:
```rust
enum ApiError {
    Database(DatabaseError),
    Auth(AuthError),
}

enum DatabaseError {
    Connection(ConnectionError),
    Query(QueryError),
}

enum ConnectionError {
    Timeout(TimeoutError),
    Credentials(CredentialsError),
}
```

**Prefer this**:
```rust
enum ApiError {
    DatabaseConnection { source: ConnectionError },
    DatabaseQuery { source: QueryError },
    DatabaseTimeout { source: TimeoutError },
    DatabaseCredentials { source: CredentialsError },
    AuthFailure { source: AuthError },
}
```

Benefits of flatter hierarchies:
- Easier to understand error types without diving into multiple files
- Clearer error paths
- Simpler pattern matching
- More explicit error handling

### Always provide a backtrace/context

All errors should include sufficient context for debugging:

```rust
#[derive(Debug, Snafu)]
pub enum DatabaseError {
    #[snafu(display("Failed to connect to database: {source}"))]
    ConnectionFailed {
        source: sqlx::Error,
        backtrace: Backtrace,
    },
    
    #[snafu(display("Query failed: {query}"))]
    QueryFailed {
        source: sqlx::Error,
        query: String,
        backtrace: Backtrace,
    },
}

// Usage example
fn execute_query(query: &str) -> Result<(), DatabaseError> {
[...]

PR #951, opened 2025-05-29
Refactor query recording

osipovartem:
Related to 

📌 Summary
This PR refactors how query history is recorded by removing the RecordingService from core-history and integrating the HistoryStore directly into core-executor.
🔍 Key Changes

- ✅ Removed the RecordingService and its usage from core-history.
- ✅ HistoryStore is now passed directly into UserSession via core-executor.
- ✅ Eliminated the reverse dependency from core-history to core-executor.
- ✅ Query recording is now handled internally within UserQuery, improving separation of concerns and testability.
- ✅ Removed replacement of query ID during history recording.
- ✅ Existing tests were moved to new place

🎯 Motivation
This change aligns with the design discussed in #939 . The goals were to:
- Decouple history tracking from query execution.
- Avoid the circular dependency between core-history and core-executor.
- Enable easier testing by injecting HistoryStore directly where it’s needed.

🧪 Testing
- Existing tests have been updated to work with the new architecture.
- Verified that queries are still properly recorded in history.

📘 Notes
This paves the way for further improvements to history persistence, including mockable stores for testing or pluggable history backends.


crates/core-executor/src/models.rs
```diff
@@ -16,11 +47,11 @@ pub struct QueryResult {
 
 impl QueryResult {
     #[must_use]
-    pub fn new(records: Vec<RecordBatch>, schema: Arc<ArrowSchema>) -> Self {
+    pub const fn new(records: Vec<RecordBatch>, schema: Arc<ArrowSchema>, query_id: i64) -> Self {
```
DanCodedThis, line 50: Why did it become const?

crates/core-executor/src/query.rs
```diff
@@ -181,6 +163,33 @@ impl UserQuery {
...
+                        Ok(encoded_res) => {
+                            let result_count = i64::try_from(res.records.len()).unwrap_or(0);
+                            query_record.query_finished(result_count, Some(encoded_res));
+                        }
+                        Err(err) => query_record.query_finished_with_error(err.to_string()),
+                    },
+                    Err(err) => query_record.query_finished_with_error(err.to_string()),
+                }
+            }
+            Err(err) => query_record.query_finished_with_error(err.to_string()),
```
DanCodedThis: I think I finally understood this. I think this can be rewritten to:

```rust
match serde_json::to_string(&query_result_to_result_set(query_execution_result, self.session.config.dbt_serialization_format)))
    Ok(Ok(Ok(encoded_res)))) => {
        let result_count = i64::try_from(res.records.len()).unwrap_or(0);
        query_record.query_finished(result_count, Some(encoded_res));
    }
    other => {
        query_record.query_finished_with_error(other.to_string())
    } 
```

To avoid multiple matches, it should still allow us to propogate the error to the query_record.

crates/core-executor/src/service.rs
```diff
@@ -97,9 +116,20 @@ impl ExecutionService for CoreExecutionService {
...
+        // This ensures consistent tracking and logging of the query across all layers.
+        let mut query_obj =
+            user_session.query(query, query_context.with_query_id(query_record.query_id()));
+
+        let query_result = query_obj.execute().await;
+        // Record the query in the session’s history, including result count or error message.
+        // This ensures all queries are traceable and auditable within a session, which enables
+        // features like `last_query_id()` and enhances debugging and observability.
+        query_obj
+            .record_query(&mut query_record, &query_result)
```
DanCodedThis: Just curious, why not make this `fn` part of the session? Seems confusing to me that the `query_obj` records its own result. Except if we need the `history_store` for the query_obj also elsewhere (not only for this `fn`)?

DanCodedThis (approved):
Overall LGTM (a very welcome change)

crates/core-executor/src/models.rs
osipovartem: clippy warning

crates/core-executor/src/service.rs
osipovartem: It should be moved, forgot to push last commit :D

crates/api-snowflake-rest/src/error.rs
```diff
@@ -130,6 +130,7 @@ fn convert_into_response(error: &ExecutionError) -> axum::response::Response {
         | ExecutionError::UploadFailed { .. } => http::StatusCode::BAD_REQUEST,
         ExecutionError::ObjectAlreadyExists { .. } => http::StatusCode::CONFLICT,
         ExecutionError::Arrow { .. }
+        | ExecutionError::SerdeParse { .. }
```
YaroslavLitvinov, line 133: It's not related particularly to this PR, but it seems there should be some common errors reused by api-snowflake-rest and api-ui

crates/core-executor/src/query.rs
osipovartem: With nested Results we still have to extract it every time

DanCodedThis: :(

crates/core-executor/Cargo.toml
```diff
@@ -7,6 +7,7 @@ license-file.workspace = true
 [dependencies]
 core-utils = { path = "../core-utils" }
```
YaroslavLitvinov, line 8: it depends on core_utils in tests, and also on ScanIterator.

 @DanCodedThis could ScanIterator be moved out from utils, so we core-executor wouldn't depend on core-utils?

DanCodedThis (approved):
LGTM

YaroslavLitvinov (approved):

crates/api-snowflake-rest/src/error.rs
DanCodedThis: @YaroslavLitvinov #866

crates/core-executor/Cargo.toml
DanCodedThis: @YaroslavLitvinov I think it can be deleted, it was created for pagination on UI for Metastore. Perhaps will need it later ot will be able to extend it, if so it can probably be moved to Metastore (not sure though).

crates/core-executor/src/models.rs
```diff
@@ -158,10 +189,32 @@ impl ColumnInfo {
...
+
+impl Row {
+    #[must_use]
+    pub const fn new(values: Vec<Value>) -> Self {
+        Self(values)
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct ResultSet {
```
rampage644, line 209: Isn't it something used only in the UI to return results to the user? I realize this is also used to persist results in the database as well.

Hypothetically:

 * had we drop the requirement to persist query results (results only! i.e. record batches), then it's clear this serialization belongs to api-ui only
 * even with database layer serialization, it looks this is either core-history or api-ui

crates/core-executor/src/service.rs
```diff
@@ -97,9 +142,18 @@ impl ExecutionService for CoreExecutionService {
                 })?
                 .clone()
         };
+        let mut query_record = QueryRecord::query_start(query, query_context.worksheet_id);
```
rampage644, line 145: Honestly, I was hoping it would be possible to somehow move out all serialization stuff out of this crate and isolate it somewhere else. 

```rust
let builder = self.history_store.query_record_builder().with_worksheet_id(worksheet_id).with_query(query);
let query = user_session.query(query, query_context.with_query_id(query_record.query_id()));
let results = query.execute().await;
let builder = builder.with_result(results); // Here to decouple core-history from core-executor completely can avoid hard dependency on core-executor models
// i.e.
// let builder = builder.with_schema(results.schema());
// let builder = builder.with_columns(results.columns());
// let builder = builder.with_metadata(results.metadata());
self.history_store.save_query_record(builder.build());
results
```

crates/core-executor/src/service.rs
```diff
@@ -40,14 +44,48 @@ pub trait ExecutionService: Send + Sync {
...
+    }
+
+    pub async fn record_query(
+        &self,
+        query_record: &mut QueryRecord,
+        query_execution_result: &ExecutionResult<QueryResult>,
+    ) {
+        match query_execution_result {
+            Ok(res) => {
+                match query_result_to_result_set(res, self.config.dbt_serialization_format) {
```
rampage644, line 73: In addition to other comments, I feel now it should be core-history responsibility to persist query results, serializing it the way it wants to.

rampage644 (approved):
Overall, it is a move towards removal of RecordingService. However, I would love to see further redesign. 

I have a strong opinion core-executor should not depend on serde as serialization is an api and storage layers concern (serialize to send through the network and serialize to persist in the database).

I find it is not easy to untangle current dependencies but ideally it could be something like this:

 * ExecutionService returns its own struct/models (with no dependency on `serde`)
 * HistoryStore operates on its own structs (ideally with no serde dependency, but I realize this might be hard)
 * in `query()` we call history store to maybe even construct query record and later persist it
 * It is history store responsibily to convert to whatever format it wants data to be stored
 * API handlers call history store to get query records again with query records models and can convert it to perhaps different representation. Or we refactor it later such that it calls query engine with `select * from embucket.history limit 100 order by start_time desc` instead of going for `HistoryStore`

crates/core-executor/src/models.rs
osipovartem: We should move dbt format config to history store. I don't remember if we store record batches in slatedb.

PR #952, opened 2025-05-29
add embucket query planner

JanKaul:
This PR introduces a new "EmbucketQueryPlanner" which can be used in the future with custom embucket LogicalPlans and PhysicalPlans.

It currently doesn't provide much functionality but provides the basis for further developments.

### DIFF
--- a/crates/core-executor/src/datafusion/extension_planner.rs
+++ b/crates/core-executor/src/datafusion/extension_planner.rs
@@ -1,0 +1,21 @@
+use async_trait::async_trait;
+use datafusion::physical_planner::ExtensionPlanner;
+
+#[derive(Debug, Default)]
+pub struct EmbucketExtensionPlanner {}
+
+#[async_trait]
+impl ExtensionPlanner for EmbucketExtensionPlanner {
+    async fn plan_extension(
+        &self,
+        _planner: &dyn datafusion::physical_planner::PhysicalPlanner,
+        _node: &dyn datafusion_expr::UserDefinedLogicalNode,
+        _logical_inputs: &[&datafusion_expr::LogicalPlan],
+        _physical_inputs: &[std::sync::Arc<dyn datafusion_physical_plan::ExecutionPlan>],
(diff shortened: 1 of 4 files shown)


crates/core-executor/src/session.rs
```diff
@@ -71,7 +71,7 @@ impl UserSession {
             .with_default_features()
             .with_runtime_env(Arc::new(runtime_config))
             .with_catalog_list(catalog_list_impl.clone())
-            .with_query_planner(Arc::new(IcebergQueryPlanner::new()))
```
DanCodedThis, line 83: Is it ok to just remove it, since our own planner will not do anything at the moment?

osipovartem (approved):

DanCodedThis (approved):
Otherwise LGTM

crates/core-executor/src/session.rs
DanCodedThis: P.S. I know I wasn't requested, just curious

JanKaul: I thought it would be a good idea to get this in as it demonstrates the structure of the whole setup. We will start adding extension nodes in the coming days.

crates/core-executor/src/datafusion/extension_planner.rs
```diff
@@ -0,0 +1,21 @@
+use async_trait::async_trait;
+use datafusion::physical_planner::ExtensionPlanner;
+
+#[derive(Debug, Default)]
+pub struct EmbucketExtensionPlanner {}
```
rampage644: I would avoid `embucket` in naming, seems duplicative, what about just `PhysicalPlanner`, `CustomExtensionPlanner`, `CustomPhysicalPlanner`, etc (still haven't made my mind about just planner vs. planner extension)

rampage644 (approved):
naming nitpicks

rampage644 (approved):

jonathanc-n (approved):
This looks really useful, thanks @JanKaul!

PR #953, opened 2025-05-29
`get` scalar function

ravlio:


crates/df-builtins/src/get.rs
```diff
@@ -0,0 +1,264 @@
...
+                "+-------+",
+                "| value |",
+                "+-------+",
+                "|       |",
+                "+-------+",
+            ],
+            &result
+        );
+
+        let sql = "SELECT get('[1, {\"a\":4}]',null) AS value;";
```
rampage644, line 197: Does it also cover a case for `"SELECT get('[1, {\"a\":4}]', "random-key") AS value;";`? Should it return null in this case as well?

rampage644 (approved):

crates/df-builtins/src/get.rs
ravlio: it will return null as intended

PR #955, opened 2025-05-30
add test for merge into

JanKaul:
Add test to verify if MERGE INTO implementation is running as expected.


JanKaul:
@jonathanc-n @osipovartem Do you guys have an idea why the SQL parsing fails? Shouldn't sqlparser-rs be able to parse the MERGE INTO statement.

osipovartem:
> @jonathanc-n @osipovartem Do you guys have an idea why the SQL parsing fails? Shouldn't sqlparser-rs be able to parse the MERGE INTO statement.

Try to use fully qualified name for inserts. We have an existing not fixed bug https://github.com/Embucket/embucket/issues/905
Also try to set MERGE _INTO statement as a main query to check the error (return initial SELECT back after that)
```rust
test_query!(
    merge_into_insert_and_update,
        "MERGE INTO merge_target USING merge_source ON merge_target.ID = merge_source.ID WHEN MATCHED THEN UPDATE SET description = merge_source.description WHEN NOT MATCHED THEN INSERT (ID, description) VALUES (merge_source.ID, merge_source.description)",
    setup_queries = [
        "CREATE TABLE embucket.public.merge_target (ID INTEGER, description VARCHAR)",
        "CREATE TABLE embucket.public.merge_source (ID INTEGER, description VARCHAR)",
        "INSERT INTO embucket.public.merge_target VALUES (1, 'existing row'), (2, 'existing row')",
        "INSERT INTO embucket.public.merge_source VALUES (2, 'updated row'), (3, 'new row')",]
);
```
This worked for me 
```sql
 MERGE INTO merge_target mt USING merge_source ms ON mt.ID = ms.ID
   WHEN MATCHED THEN UPDATE SET description = ms.description
   WHEN NOT MATCHED THEN INSERT (ID, description) VALUES (ms.ID, ms.description)
```

JanKaul:
Thanks a lot. So one current limitation seems to be that the MERGE INTO statement needs to use aliases.

osipovartem:
This is limitation of SQL parser, need to check why it cannot parse query without aliases. Also we support only insert operation, and just skip updates/deletes

JanKaul:
I intentionally made the test to fail. It should be the test to check if our proper merge support is actually working. That's why I created it's own branch for it now and didn't target main.

rampage644:
@JanKaul Safe to close this as outdated?

JanKaul:
Closed in favor of #1022

Issue #964, opened 2025-05-30
Update `metastore` trait to rely on `NotFound` errors instead of `Option`


rampage644:
Part of #866.

At this moment, metastore trait indicates entity not found with `Option` (i.e.     `async fn get_volume(&self, name: &VolumeIdent) -> MetastoreResult<Option<RwObject<Volume>>>;`).

However, there are already error enum variants with `TableNotFound` and others that are being used by callers.

I.e. 
```rust
#[tracing::instrument(level = "debug", skip(state), err, ret(level = tracing::Level::TRACE))]
pub async fn get_volume(
    State(state): State<AppState>,
    Path(volume_name): Path<String>,
) -> MetastoreAPIResult<Json<RwObject<Volume>>> {
    match state.metastore.get_volume(&volume_name).await {
        Ok(Some(volume)) => Ok(Json(hide_sensitive(volume))),
        Ok(None) => Err(MetastoreError::VolumeNotFound {
            volume: volume_name.clone(),
        }
        .into()),
        Err(e) => Err(MetastoreAPIError::from(e)),
    }
}
```

The goal is to move such errors origination into `metastore` (as it is done in `HistoryStore`) to simplify error handling across codebase.

PR #965, opened 2025-05-30
WIP Streamline error handling 1st attempt

rampage644:
First attempt to address #866. **Collecting early feedback.**

Introduce common error trait (stacked one, inspired by greptimedb).
Fixing meanwhile #964 - ideally this is isolated in a dedicated PR.

This is my current approach - by no means it is finished - but at least all crates affected compile (there are still 3 api crates to fix).

I would like to collect early feedback, especially on `pub enum IcebergAPIError {`. There was previously only `Metastore` option, I am not sure this is good approach in snafu (and stacked errors) philosophy.

### DIFF
--- a/crates/common-error/src/lib.rs
+++ b/crates/common-error/src/lib.rs
@@ -1,0 +1,62 @@
+// Copyright 2023 Greptime Team
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::sync::Arc;
+
+pub trait StackError: std::error::Error {
+    fn debug_fmt(&self, layer: usize, buf: &mut Vec<String>);
+
+    fn next(&self) -> Option<&dyn StackError>;
+
+    fn last(&self) -> &dyn StackError
+    where
+        Self: Sized,
+    {
+        let Some(mut result) = self.next() else {
+            return self;
+        };
+        while let Some(err) = result.next() {
+            result = err;
+        }
+        result
+    }
+
+    /// Indicates whether this error is "transparent", that it delegates its "display" and "source"
+    /// to the underlying error. Could be useful when you are just wrapping some external error,
+    /// **AND** can not or would not provide meaningful contextual info. For example, the
+    /// `DataFusionError`.
+    fn transparent(&self) -> bool {
+        false
+    }
+}
+
+impl<T: ?Sized + StackError> StackError for Arc<T> {
+    fn debug_fmt(&self, layer: usize, buf: &mut Vec<String>) {
+        self.as_ref().debug_fmt(layer, buf)
+    }
+
+    fn next(&self) -> Option<&dyn StackError> {
+        self.as_ref().next()
+    }
+}
+
+impl<T: StackError> StackError for Box<T> {
+    fn debug_fmt(&self, layer: usize, buf: &mut Vec<String>) {
+        self.as_ref().debug_fmt(layer, buf)
+    }
+
+    fn next(&self) -> Option<&dyn StackError> {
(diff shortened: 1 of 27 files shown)


crates/common-error/src/lib.rs
```diff
@@ -0,0 +1,62 @@
+// Copyright 2023 Greptime Team
```
osipovartem, line 1: do we need this?

osipovartem (approved):

crates/core-executor/src/query.rs
```diff
@@ -363,11 +361,17 @@ impl UserQuery {
...
-                catalog: name.to_string(),
-            },
-        )
+        self.session
+            .ctx
+            .state()
+            .catalog_list()
+            .catalog(name)
+            .ok_or_else(|| {
+                ex_error::CatalogNotFoundSnafu {
```
YaroslavLitvinov, line 370: This is common to this PR. What is the reason for creating error using generated CatalogNotFoundSnafu { catalog }.build() vs direct error creation CatalogNotFound { catalog } ? Is this way providing additional context or so ?

crates/core-executor/src/query.rs
```diff
@@ -1184,11 +1200,11 @@ impl UserQuery {
                 apply_show_filters(sql, &filters)
             }
             _ => {
-                return Err(ExecutionError::DataFusion {
-                    source: DataFusionError::NotImplemented(format!(
-                        "unsupported SHOW statement: {statement}"
-                    )),
-                });
+                return ex_error::NotYetImplementedSnafu {
```
YaroslavLitvinov, line 1203: It seems non implemented things categorization would be better if we would introduce additional snafu context here as well. Optionally with a not_implemented namespace like this not_implemented::OnlyShowStatementSupportedSnafu { statement }

YaroslavLitvinov (approved):

crates/core-metastore/src/metastore.rs
```diff
@@ -263,30 +249,23 @@ impl Metastore for SlateDBMetastore {
...
 
-    async fn get_volume(&self, name: &VolumeIdent) -> MetastoreResult<Option<RwObject<Volume>>> {
+    async fn get_volume(&self, name: &VolumeIdent) -> MetastoreResult<RwObject<Volume>> {
         let key = format!("{KEY_VOLUME}/{name}");
         self.db
             .get(&key)
             .await
-            .context(metastore_error::UtilSlateDBSnafu)
-            .map_err(Box::new)
+            .context(metastore_error::UtilSlateDBSnafu)?
```
YaroslavLitvinov, line 262: Support of empty results from SlateDb ruining advantage of using fancy .context approach. Would we consider refacatoring Db utils as well ?

crates/core-executor/src/query.rs
rampage644: Yes, this is to collect all required context - backtraces, location, any `implicit` ly generated data

crates/core-metastore/src/metastore.rs
rampage644: Can you elaborate? Not sure I got the problem

YaroslavLitvinov: Sorry, I picked a wrong line.
I mean we still should handle results in a custom way using `.ok_or_else`, instead of using .context without extra manipulations on all traits that is under our control.

```
.ok_or_else(|| {
    metastore_error::VolumeNotFoundSnafu {
        volume: name.clone(),
    }
    .build()
})
```

rampage644: Oh, did you mean following core-utils `Db` interface ?
```rust
pub async fn get<T: for<'de> serde::de::Deserialize<'de>>(
        &self,
        key: &str,
    ) -> Result<Option<T>> {
```

YaroslavLitvinov: This Option actually allows us having None instead of object, but I am not sure if it is supposed to be used.

YaroslavLitvinov:
It would be quite natural having base Error as a separate crate, then we all could use it.

rampage644:
@YaroslavLitvinov Do you think we should close this one in favor of #1050 ?

YaroslavLitvinov:
> @YaroslavLitvinov Do you think we should close this one in favor of #1050 ?

I think we can close, but so far need to keep related `errors` branch.

crates/common-error/src/lib.rs
rampage644: Yes, this has to be kept due to the Apache license

PR #967, opened 2025-06-01
POC: Support Merge `LogicalPlan` + `ExecutionPlan`

jonathanc-n:
Implements `LogicalPlan` + `ExecutionPlan` for Merge Into statement.


crates/core-executor/src/query.rs
```diff
@@ -892,6 +898,185 @@ impl UserQuery {
         self.execute_with_custom_plan(&insert_query).await
     }
 
+    #[instrument(level = "trace", skip(self), err, ret)]
+    pub async fn merge_query_(&self, statement: Statement) -> ExecutionResult<Vec<RecordBatch>> {
```
jonathanc-n, line 926: This part is currently a bit rough, I will be fixing up this part by referencing the correct tables.

crates/core-executor/src/query.rs
```diff
@@ -892,6 +898,185 @@ impl UserQuery {
...
+                }
+            }
+        }
+
+        let delete_pred = delete_condition
+            .into_iter()
+            .reduce(|a, b| a.or(b))
+            .unwrap_or_else(|| lit(false));
+
+        let merge_exec: MergeExec = MergeExec {
```
jonathanc-n, line 1087: What is being passed in here is currently the old idea I had in mind for the `MergeExec` please ignore this

crates/core-executor/src/datafusion/physical_plan/merge.rs
```diff
@@ -0,0 +1,281 @@
...
+use datafusion_expr::Expr;
+use datafusion_expr::JoinType;
+use datafusion_expr::LogicalPlan;
+use datafusion_physical_plan::RecordBatchStream;
+use datafusion_physical_plan::joins::PartitionMode;
+use datafusion_physical_plan::{ExecutionPlan, PlanProperties, SendableRecordBatchStream};
+use futures::Stream;
+use futures::StreamExt;
+
+pub struct MergeExec {
```
jonathanc-n, line 29: This is the final self iteration of the `MergeExec` logic right now, however it only supports Insert and updates as of right now.

jonathanc-n (commented):
This is a rough poc for now, in the next two days i will be making heavy revisions. Right now, I want this to be more of a discussion for the MergeExec logic and if it seems correct. @JanKaul @rampage644 @jatin510 @osipovartem Would appreciate some feedback here.

JanKaul:
Hey jonathan, thanks for the great work!

In the current state it's a bit difficult to entirely understand the implementation. Would it be possible to split the PR into smaller chunks?

I think the `query_merge` function with the generation of the case expressions looks really great. Could we tackle that in a first PR and then move to the rest. For now it could just execute a ProjectionExec with the case expressions applied.

And then we can tackle the MergeExec in a follow up PR. I think there are still a couple of open questions.

crates/core-executor/src/query.rs
JanKaul: To reduce the size of this PR I would execute a `ProjectionExec` here that applies the case Expressions. It takes the `join_plan` as input. We can then later use the projection_plan as input the the MergeExec implementation.

crates/core-executor/src/datafusion/physical_plan/merge.rs
```diff
@@ -0,0 +1,281 @@
...
+            self.output_schema.clone(),
+        );
+
+        Ok(Box::pin(merge_stream))
+    }
+}
+
+/// A helper stream that wraps a `RecordBatchStream`` of joined batches and
+/// appends a new int32 column called “op” = 0/1.
+struct MergeStream {
```
JanKaul, line 207: Wouldn't it be possible to achieve the same with a CASE expression that returns 0,1,2 for the appropriate cases? Then we don't have to implement a custom RecordbatchStream.

JanKaul:
I would have created a custom Logical & Physical Plan only for the final step of the MERGE INTO operation. Something along the lines of: https://github.com/JanKaul/embucket/compare/main...JanKaul:embucket:merge-extension. The idea would be to adopt the `merge_query` function from this PR and execute the "MergeIntoSink" as a final step.

PR #969, opened 2025-06-02
`try_parse_json` scalar function

### LINKED ISSUE #977: Missing function: try_parse_json

ravlio:
closes #977

### DIFF
--- a/crates/df-builtins/src/try_parse_json.rs
+++ b/crates/df-builtins/src/try_parse_json.rs
@@ -1,0 +1,52 @@
+use datafusion::arrow::datatypes::DataType;
+use datafusion::error::Result as DFResult;
+use datafusion::logical_expr::{ColumnarValue, Signature, TypeSignature, Volatility};
+use datafusion_expr::{ScalarFunctionArgs, ScalarUDFImpl};
+use std::any::Any;
+
+#[derive(Debug)]
+pub struct TryParseJsonFunc {
+    signature: Signature,
+}
+
+impl Default for TryParseJsonFunc {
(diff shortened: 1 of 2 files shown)


crates/df-builtins/src/try_parse_json.rs
```diff
@@ -0,0 +1,52 @@
...
+
+    fn signature(&self) -> &Signature {
+        &self.signature
+    }
+
+    fn return_type(&self, _arg_types: &[DataType]) -> DFResult<DataType> {
+        Ok(DataType::Utf8)
+    }
+
+    fn invoke_with_args(&self, args: ScalarFunctionArgs) -> DFResult<ColumnarValue> {
```
jonathanc-n, line 46: I think we should include the parameters in this pull request ([snowflake spec](https://docs.snowflake.com/en/sql-reference/functions/try_parse_json))

jonathanc-n (commented):
Tests should be added as well for testing the different parameters once implemented.

crates/df-builtins/src/try_parse_json.rs
ravlio: in the next pr

Vedin (approved):

crates/df-builtins/src/try_parse_json.rs
ravlio: Checked. We don't have any SLT tests or workloads with parameters. At the same time, implementing duplicate object key checking would require significant effort — it would need a recursive deserializer, and the parsing process would be slow. @jonathanc-n

PR #983, opened 2025-06-02
Parse current ip address for current_ip_address() func

### LINKED ISSUE #970: [BUG] Missing current_ip_address
**Function:** `current_ip_address `
**Category:** Context
**Documentation:** https://docs.snowflake.com/en/sql-reference/functions/current_ip_address

**Test SQL:**
```sql
SELECT CURRENT_IP_ADDRESS();
```

**Expected Result:**
```
N/A
```

**Actual Result:**
```
SELECT CURRENT_IP_ADDRESS();
```

**Stderr:**
```
╭─ Error ──────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ 000200: 200: DataFusion error: Error during planning: Invalid function 'current_ip_address'.                     │
│ Did you mean 'current_database'?                                                                                 │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

**Parent EPIC:** #571

osipovartem:
Closes #970

Implementation for current_ip_address func
https://docs.snowflake.com/en/sql-reference/functions/current_ip_address

### DIFF
--- a/crates/api-snowflake-rest/src/handlers.rs
+++ b/crates/api-snowflake-rest/src/handlers.rs
@@ -7,7 +7,7 @@ use crate::state::AppState;
 use api_sessions::DFSessionId;
 use axum::Json;
 use axum::body::Bytes;
-use axum::extract::{Query, State};
+use axum::extract::{ConnectInfo, Query, State};
 use axum::http::HeaderMap;
 use base64;
 use base64::engine::general_purpose::STANDARD as engine_base64;
@@ -23,6 +23,7 @@ use flate2::read::GzDecoder;
 use regex::Regex;
 use snafu::ResultExt;
 use std::io::Read;
+use std::net::SocketAddr;
 use tracing::debug;
 use uuid::Uuid;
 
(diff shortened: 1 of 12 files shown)


crates/api-snowflake-rest/src/handlers.rs
```diff
@@ -93,6 +94,7 @@ fn records_to_json_string(recs: &[RecordBatch]) -> Result<String, DbtError> {
 
 #[tracing::instrument(level = "debug", skip(state, body), err, ret(level = tracing::Level::TRACE))]
 pub async fn query(
+    ConnectInfo(addr): ConnectInfo<SocketAddr>,
```
rampage644, line 97: What if we move it even out of handlers and instead get it during app initialization?

osipovartem: This is related to client IP, not server, so each request should be handled separately 
Also ConnectInfo is just like headers info, specific for each request.

rampage644 (approved):

PR #986, opened 2025-06-02
Fix/statement update logic

jatin510:
Issue: https://github.com/Embucket/embucket/issues/905

reverted the changes which was removed in :

https://github.com/Embucket/embucket/pull/911#issuecomment-2918550733

### DIFF
--- a/crates/core-executor/src/query.rs
+++ b/crates/core-executor/src/query.rs
@@ -23,7 +23,7 @@ use datafusion::catalog::{CatalogProvider, SchemaProvider};
 use datafusion::datasource::default_table_source::provider_as_source;
 use datafusion::execution::session_state::SessionContextProvider;
 use datafusion::execution::session_state::SessionState;
-use datafusion::logical_expr::{LogicalPlan, TableSource};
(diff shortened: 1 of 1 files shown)


crates/core-executor/src/query.rs
```diff
@@ -1401,6 +1404,309 @@ impl UserQuery {
         }
     }
 
+    // TODO: Modify this function to modify the statement in-place to
+    // avoid extra allocations
+    #[allow(clippy::too_many_lines)]
+    pub fn update_statement_references(
```
osipovartem, line 1410: what do you think about using relations visitor mut here?
we call this code to get all references which is based on RelationVisitor
```rust
  let references = state
      .resolve_table_references(statement)
```
we can implement simple VisitorMut which can replace (fix) all relateions in statement

osipovartem: Check this PR please 
https://github.com/Embucket/embucket/pull/996

jatin510:
Solved in PR: https://github.com/Embucket/embucket/pull/996

PR #1000, opened 2025-06-03
GET `/ui/query/{queryRecordId}`

### LINKED ISSUE #972: api: Add GET /ui/queries/${queryId} endpoint
Returns QueryRecord

DanCodedThis:
Closes #972

- Added GET `/ui/query/{queryRecordId}`
- Added test

### DIFF
--- a/crates/api-ui/src/queries/error.rs
+++ b/crates/api-ui/src/queries/error.rs
@@ -19,7 +19,7 @@ pub enum QueryError {
         source: core_executor::error::ExecutionError,
     },
 
-    #[snafu(transparent)]
+    #[snafu(display("History store: {source}"))]
     Store { source: HistoryStoreError },
 
     #[snafu(display("Failed to parse row JSON: {source}"))]
@@ -41,7 +41,7 @@ pub enum QueriesAPIError {
     Query { source: QueryError },
     #[snafu(display("Error getting queries: {source}"))]
     Queries { source: QueryError },
-    #[snafu(display("Error getting queries: {source}"))]
+    #[snafu(display("Error getting query record: {source}"))]
     GetQueryRecord { source: QueryError },
 }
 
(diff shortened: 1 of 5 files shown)


crates/api-ui/src/queries/error.rs
```diff
@@ -41,6 +41,8 @@ pub enum QueriesAPIError {
     Query { source: QueryError },
     #[snafu(display("Error getting queries: {source}"))]
     Queries { source: QueryError },
+    #[snafu(display("Error getting queries: {source}"))]
```
YaroslavLitvinov: unrelated display string

crates/api-ui/src/queries/error.rs
```diff
@@ -66,6 +68,15 @@ impl IntoStatusCode for QueriesAPIError {
                 QueryError::ResultParse { .. } => StatusCode::UNPROCESSABLE_ENTITY,
                 _ => StatusCode::INTERNAL_SERVER_ERROR,
             },
+            Self::GetQueryRecord { source } => match &source {
```
YaroslavLitvinov: Is GetQueryRecord that much differs from GetQuery ? Maybe it makes sence to combine them like
`Self::GetQueryRecord { source } | Self::Queries { source } => match &source`

crates/api-ui/src/queries/handlers.rs
```diff
@@ -114,6 +118,41 @@ pub async fn query(
...
+        (status = 400, description = "Bad query record id", body = ErrorResponse),
+        (status = 500, description = "Internal server error", body = ErrorResponse)
+    )
+)]
+#[tracing::instrument(level = "debug", skip(state), err, ret(level = tracing::Level::TRACE))]
+pub async fn get_query(
+    State(state): State<AppState>,
+    Path(query_record_id): Path<QueryRecordId>,
+) -> QueriesResult<Json<QueryGetResponse>> {
+    match state.history_store.get_query(query_record_id).await {
```
YaroslavLitvinov: Could we use .context and .map here? In a way it was recently refactored.

crates/api-ui/src/router.rs
```diff
@@ -118,6 +118,7 @@ pub fn create_router() -> Router<AppState> {
                 .patch(update_worksheet),
         )
         .route("/queries", post(query).get(queries))
+        .route("/query/{queryRecordId}", get(get_query))
```
YaroslavLitvinov: I'm not sure how it should be, maybe /queries/get-query-id ?

YaroslavLitvinov (approved):

crates/api-ui/src/router.rs
DanCodedThis: I think it could be just `/queries/{queryRecordId}` just like in dbs, schemas, etc

crates/api-ui/src/queries/handlers.rs
DanCodedThis: Possibly, yes

nikitastryuk (approved):

PR #1009, opened 2025-06-04
`to_variant` scalar function

ravlio:


Vedin (changes requested):
let's add some snapshot tests to new folder src/tests/conversions for this PR. Like test for null and 2-3 tests for different datatypes is fine.

Vedin (approved):

Vedin (approved):

PR #1019, opened 2025-06-05
Add Visitor to automatically alias duplicate select expressions

### LINKED ISSUE #1018: [BUG]: Projections require unique expression names
Here is the same issue

```sql
CREATE TABLE test (col1 DATE,  col2 DATE);
INSERT INTO test
  SELECT TO_DATE('2013-05-08T23:39:20.123'), TO_DATE('2013-05-08T23:39:20.123')
```
```rust
Projections require unique expression names but the expression "to_date(Utf8("2013-05-08T23:39:20.123"))" at position 0 and "to_date(Utf8("2013-05-08T23:39:20.123"))" at position 1 have the same name. Consider aliasing ("AS") one of them.
```

https://github.com/apache/datafusion/issues/6543

osipovartem:
This PR introduces a new VisitorMut implementation called AddAliasesToSelectExpressions, which traverses SQL statements and automatically adds unique aliases to duplicate expressions in SELECT clauses.

**Motivation**

This change addresses an issue where SQL queries containing repeated expressions without aliases fail during logical plan construction due to non-unique projection names. An example of such a query is:
```sql
CREATE TABLE test (col1 DATE, col2 DATE);
INSERT INTO test
  SELECT TO_DATE('2013-05-08T23:39:20.123'), TO_DATE('2013-05-08T23:39:20.123');
```
Without aliasing, this produces a planning error in DataFusion:
```rust
Projections require unique expression names but the expression "to_date(Utf8("2013-05-08T23:39:20.123"))" at position 0 and "to_date(Utf8("2013-05-08T23:39:20.123"))" at position 1 have the same name. Consider aliasing ("AS") one of them.
```
https://github.com/apache/datafusion/issues/6543

**Solution**
To ensure compatibility and prevent planner errors, we introduced a visitor that:
- Traverses all SELECT statements within a SQL query.
- Detects repeated expressions (based on stringified form).
- Automatically adds aliases like expr_0, expr_1, etc., to duplicate expressions that do not already have aliases.
This logic is encapsulated in a reusable visitor and applied before logical planning, ensuring that expression names are always unique within a single SELECT clause.

**Testing**
Unit tests have been added to verify:
- Aliases are added to duplicate expressions.
- Expressions with existing aliases are respected.
- Mixed aliased and unaliased expressions are handled correctly.

Closes #1018


DanCodedThis (approved):
LGTM

Vedin (approved):
lgtm! I know just think about do we need to do some postprocessing when return result in such cases? As I understand resulting column names still be different from Snowflake behaviour.

osipovartem:
> lgtm! I know just think about do we need to do some postprocessing when return result in such cases? As I understand resulting column names still be different from Snowflake behaviour.

Do you think we need to save initial hasmap of aliases? and apply them in the end?

PR #1022, opened 2025-06-05
POC: Merge extension

JanKaul:
This PR provides a rough structure of how the writing phase of the MERGE INTO statement looks like. There are still some parts missing. But I think the general structure with a LogicalPlan "MergeIntoSink" and a PhysicalPlan "MergeIntoSinkExec" should work.

The idea is to adopt the `merge_query` function to prepare a logical plan that emits a RecordBatchStream of the desired form. As a last step of the `merge_query` function this logical plan is then passed as an input to the "MergeIntoSink" LogicalPlan. The Extension plan will then during execution convert the "MergeIntoSink" into the "MergeIntoSinkExec" which will perform the desired operations.


crates/core-executor/src/datafusion/logical_plan/merge.rs
```diff
@@ -0,0 +1,110 @@
...
+// The MergeIntoSink performs the final writing step of the "MERGE INTO" statement. It assumes
+// that the typical Join of the target and source tables is already performed and that the Recordbatches
+// have a particular form. As such the Recordbatches must contain a "__target_exists" and
+// "__source_exists" column that indicate whether the given row exists in the target and source
+// tables respectivaly. Additionally, it requires a "__data_file_path" column to keep track of which files to overwrite.
+// It then writes the resulting data to parquet files and updates the target
+// table accordingly.
+pub struct MergeIntoSink {
+    pub input: Arc<LogicalPlan>,
+    pub target: DataFusionTable,
```
JanKaul, line 20: @jonathanc-n The target could also be an `Arc<dyn TableProvoder>`

JanKaul: You can also pass the target as LogicalPlan

crates/core-executor/src/datafusion/logical_plan/merge.rs
```diff
@@ -0,0 +1,110 @@
...
+    ) -> datafusion_common::Result<()> {
+        Ok(())
+    }
+
+    fn expressions(&self) -> Vec<Expr> {
+        vec![]
+    }
+
+    fn fmt_for_explain(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(f, "MergeIntoSink")
```
jonathanc-n, line 73: Maybe we can specify the catalog, schema, and table here

crates/core-executor/src/datafusion/logical_plan/merge.rs
```diff
@@ -0,0 +1,110 @@
...
+        Ok(Arc::new(Self {
+            input: Arc::new(inputs.into_iter().next().unwrap()),
+            target: self.target.clone(),
+            schema: self.schema.clone(),
+        }))
+    }
+
+    fn dyn_hash(&self, state: &mut dyn Hasher) {
+        "MergeIntoSink".dyn_hash(state);
+        self.input.dyn_hash(state);
```
jonathanc-n, line 100: Should hash target as well, if two sinks point at different tables can lead to errors

jonathanc-n (commented):
Currently just looking at it, will look at it again when it is more complete

crates/core-executor/src/datafusion/logical_plan/merge.rs
```diff
@@ -0,0 +1,114 @@
...
+    pub target: DataFusionTable,
+    pub schema: DFSchemaRef,
+}
+
+impl MergeIntoCOWSink {
+    pub fn new(
+        input: Arc<LogicalPlan>,
+        target: DataFusionTable,
+    ) -> datafusion_common::Result<Self> {
+        let field = Field::new("number of rows updated", DataType::Int64, false);
```
rampage644, line 29: "number of rows updated" --> "count"/"number" etc? What datafusion schema for returning affected rows?

PR #1023, opened 2025-06-05
`last_day` scalar function

### LINKED ISSUE #627: [BUG] Missing function: last_day
**Function:** `last_day`
**Category:** Date & time
**Documentation:** https://docs.snowflake.com/en/sql-reference/functions/last_day

**Test SQL:**
```sql
SELECT LAST_DAY('2025-02-15'::date, 'month');
```

**Expected Result:**
```
N/A
```

**Actual Result:**
```
SELECT LAST_DAY('2025-02-15'::date, 'month');
```

**Stderr:**
```
│ 000200: 200: DataFusion error: Error during planning: Invalid function 'last_day'.                                                                          │
│ Did you mean 'list_has'?                                                                                                                                    │
```

**Parent EPIC:** #515

ravlio:
closes #627

### DIFF
--- a/crates/embucket-functions/src/datetime/last_day.rs
+++ b/crates/embucket-functions/src/datetime/last_day.rs
@@ -1,0 +1,188 @@
+use chrono::{DateTime, Datelike, NaiveDate, NaiveDateTime, Utc};
+use datafusion::arrow::array::{Array, Date64Builder};
+use datafusion::arrow::datatypes::{DataType, TimeUnit};
+use datafusion::error::Result as DFResult;
+use datafusion::logical_expr::TypeSignature::{Coercible, Exact};
+use datafusion::logical_expr::{Coercion, ColumnarValue, TypeSignatureClass};
+use datafusion_common::types::logical_string;
+use datafusion_common::{ScalarValue, exec_err};
+use datafusion_expr::{ScalarFunctionArgs, ScalarUDFImpl, Signature, Volatility};
+use std::any::Any;
+use std::sync::Arc;
+
+#[derive(Debug)]
+pub struct LastDayFunc {
+    signature: Signature,
+}
+
+impl Default for LastDayFunc {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl LastDayFunc {
+    #[must_use]
+    pub fn new() -> Self {
+        Self {
+            signature: Signature::one_of(
+                vec![
+                    Coercible(vec![
+                        Coercion::new_exact(TypeSignatureClass::Timestamp),
+                        Coercion::new_exact(TypeSignatureClass::Native(logical_string())),
+                    ]),
+                    Coercible(vec![Coercion::new_exact(TypeSignatureClass::Timestamp)]),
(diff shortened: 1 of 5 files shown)


crates/embucket-functions/src/datetime/last_day.rs
```diff
@@ -0,0 +1,203 @@
...
+        };
+
+        let arr = match args[0].clone() {
+            ColumnarValue::Array(arr) => arr,
+            ColumnarValue::Scalar(v) => v.to_array()?,
+        };
+
+        let mut res = Date64Builder::with_capacity(arr.len());
+        for i in 0..arr.len() {
+            let v = ScalarValue::try_from_array(&arr, i)?
```
rampage644, line 102: Why casting to nanosecond precision? I think we're currently limited to microseconds only (Iceberg has no nanosecond support as of v2).

crates/embucket-functions/src/datetime/last_day.rs
```diff
@@ -0,0 +1,203 @@
...
+
+#[allow(
+    clippy::unwrap_used,
+    clippy::as_conversions,
+    clippy::cast_possible_truncation
+)]
+fn last_day(date: &NaiveDateTime, date_part: &str) -> DFResult<NaiveDateTime> {
+    let date = date.date();
+
+    let new_date = match date_part.to_lowercase().as_str() {
```
rampage644, line 130: Perhaps not in this PR, but does it make sense to split this functions into 3 different implementation for performance reasons? Then rely on `simplify` to replace with correct option during planning? 

@Vedin Something like this to add to the `README`? I.e. where we okay with naive for loop implementation vs. we begin to worry about performance?

crates/embucket-functions/src/datetime/last_day.rs
rampage644: Okay, realized storage doesn't matter here (but just for the context - https://github.com/Embucket/embucket/blob/main/crates/core-executor/src/datafusion/type_planner.rs#L26 )

crates/embucket-functions/src/datetime/last_day.rs
ravlio: > Perhaps not in this PR, but does it make sense to split this functions into 3 different implementation for performance reasons?

@rampage644 Do you mean `last_day` (you quoted it) here function or you want to process `ColumnarValue::Array` and `ColumnarValue::Scalar` separately. With second we can gain a little more performance spending more time on development. Based on README we have to implement separate logic for ColumnarValue::Array and ColumnarValue::Scalar So I started doing this in the latest PRs.

Vedin: > Perhaps not in this PR, but does it make sense to split this functions into 3 different implementation for performance reasons? Then rely on `simplify` to replace with correct option during planning?
> 
> @Vedin Something like this to add to the `README`? I.e. where we okay with naive for loop implementation vs. we begin to worry about performance?

I'd say that we always need to worry about performance but I didn't fully get your comment. I don't think that `simplify` is always required. Perheps,  we need some investigation into it and then change our approach.

Vedin (approved):

PR #1024, opened 2025-06-05
Implement listagg

Vedin:


rampage644:
@andheroe @YevheniiNiestierov No correct slt tests were detected

crates/embucket-functions/src/aggregate/listagg.rs
```diff
@@ -0,0 +1,244 @@
...
+
+        let arr = &values[0];
+
+        // Handle delimiter if present
+        if values.len() > 1 && !self.delimiter_set {
+            if values[1].data_type() == &DataType::Null {
+                // Delimiter is NULL, keep default empty string
+            } else {
+                // Convert delimiter to string
+                let delim_string_arr = compute::cast(&values[1], &DataType::Utf8)?;
```
rampage644, line 131: Nice! I wonder if kernels (`compute::cast) deserve to be menitoned in the README (or instructions - I don't know where exactly we expect agents to pick up context from exactly).

rampage644 (approved):

crates/embucket-functions/src/aggregate/listagg.rs
Vedin: I'll add it to the instructions for now.

PR #1029, opened 2025-06-06
Inline Aliases in SELECT Expressions Using VisitorMut

### LINKED ISSUE #1028: [BUG]: Cannot use aliased expressions within same query
```sql
SELECT 
       '2013-05-08T23:39:20.123-07:00'::TIMESTAMP AS tstamp,
       YEAR(tstamp) AS "YEAR", 
       QUARTER(tstamp) AS "QUARTER OF YEAR",
       MONTH(tstamp) AS "MONTH", 
       DAY(tstamp) AS "DAY",
       DAYOFMONTH(tstamp) AS "DAY OF MONTH",
       DAYOFYEAR(tstamp) AS "DAY OF YEAR";
```
```rust
Query execution error: DataFusion error: Schema error: No field named tstamp.
```
This error occurs because DataFusion does not yet know that tstamp is an alias for '2013-05-08T23:39:20.123-07:00'::TIMESTAMP at the time the query is parsed/planned. Unlike some SQL engines (e.g. PostgreSQL), DataFusion does not support referencing an alias in the same projection SELECT clause where it is defined.

Related to https://github.com/apache/datafusion/issues/16291
Related to https://github.com/apache/datafusion/issues/6543

osipovartem:
Related to https://github.com/apache/datafusion/issues/16291
Related to https://github.com/apache/datafusion/issues/6543
Closes #1028 

This PR introduces a new SQL AST visitor, InlineAliasesInSelect, which recursively replaces alias references inside SELECT projections with their corresponding expressions. This is useful for ensuring correct behavior when aliased expressions are referenced immediately in the same projection.

**📌 Motivation**
Some queries rely on alias reuse within the same SELECT clause, like:
```sql
SELECT 'testxt' AS alias, LENGTH(alias)
```
DataFusion currently does not resolve alias in such cases during planning, which results in:
```rust
Schema error: No field named alias
```

This visitor solves the problem by inlining aliases before query execution planning.

🧠 How It Works
- For every SELECT in the query:
- Collect all alias-to-expression mappings (ExprWithAlias)
- Traverse every expression in the projection using visit_expressions_mut
- Replace any Expr::Identifier matching an alias with the corresponding original expression

✅ Example

Transforms:
```sql
SELECT 'test' AS alias, LENGTH(alias)
```
Into an equivalent form:
```sql
SELECT 'test' AS alias, LENGTH('test')
```

🚨 Notes
- Aliases are only inlined inside the same SELECT clause where they are defined.
- This transformation is safe and local to the projection clause; it does not alter query semantics outside of that.


ravlio (dismissed):

DanCodedThis (dismissed):
@osipovartem LGTM

rampage644 (approved):

PR #1042, opened 2025-06-09
Split year.slt into multiple files - one per function.

### LINKED ISSUE #1021: Split "year.slt" into separate files - one file per function
In the file https://github.com/Embucket/embucket/blob/main/test/sql/sql-reference-functions/Date_%26_time/year.slt there are many functions bundled together. That is because in Snowflake reference it looks like this: https://docs.snowflake.com/en/sql-reference/functions/year.
For our needs, we need to split this SLT file into multiple files - one file per function.

andheroe:


osipovartem (approved):

andheroe:
@osipovartem do you think SQL Logic Test results above ^^ looks right?

Issue #1037, opened 2025-06-08
`dbt run` and `dbt seed` unable to figure schema when run with `iceberg_materialization`


rampage644:
When `dbt` is run with iceberg_materialization set to true it fails due to inability to get the state of tables already created.

First, it fails with 
```
Snowflake adapter: Snowflake error: 000200: DataFusion error: This feature is not implemented: Function 'result_scan' is not implemented yet. Details: https://docs.snowflake.com/en/sql-reference/functions/result_scan
```

which was introduced in #959

Commenting this out it then fails with
```
18:48:31  On list_snowplow_public_snowplow_manifest: /* {"app": "dbt", "dbt_version": "1.9.4", "profile_name": "my_profile", "target_name": "dev", "connection_name": "list_snowplow_public_snowplow_manifest"} */
select all_objects.*, all_tables.IS_ICEBERG as "is_iceberg"
from table(result_scan(last_query_id(-1))) all_objects
left join snowplow.INFORMATION_SCHEMA.tables as all_tables
on all_tables.table_name = all_objects."name"
and all_tables.table_schema = all_objects."schema_name"
and all_tables.table_catalog = all_objects."database_name"
;
18:48:31  Snowflake adapter: Snowflake query id: None
18:48:31  Snowflake adapter: Snowflake error: 000200: DataFusion error: Schema error: No field named all_tables.is_iceberg. Valid fields are all_objects.database_name, all_objects.schema_name, all_objects.name, all_objects.kind, all_objects.comment, all_objects.is_iceberg, all_objects.is_dynamic, all_tables.table_catalog, all_tables.table_schema, all_tables.table_name, all_tables.table_type.
```

Note `all_objects.is_iceberg` is present whereas it tries to find `all_tables.is_iceberg`, this can be quickly fixed by updating information_schema.tables schema.

However, bigger question is how to refactor existing code that relied on hardcoding `select * from table(result_scan(last_query_id(-1)))` statement.

PR #1048, opened 2025-06-10
Table func result_scan

osipovartem:
Implementation for table func [result_scan](https://docs.snowflake.com/en/sql-reference/functions/result_scan)
The func returns the result set of a previous command (within 24 hours of when you executed the query) as if the result was a table.
Related to #1037


rampage644:
I think `result_scan` has to be removed from unimplemented functions visitor - `000200: DataFusion error: This feature is not implemented: Function 'result_scan' is not implemented yet.`

crates/embucket-functions/src/table/result_scan.rs
```diff
@@ -0,0 +1,143 @@
...
+use std::io::{BufReader, Cursor};
+use std::sync::Arc;
+
+#[derive(Debug, Clone)]
+pub struct ResultScanFunc {
+    history_store: Arc<dyn HistoryStore>,
+}
+
+/// `RESULT_SCAN`
+/// Returns the result set of a previous command (within 24 hours of when you executed the query)
```
rampage644: Is it copied from Snowflake documentation? There is no feature at the moment, that will remove history entries after 24 hours

osipovartem: Copied from Snowflake, let's remove non relevant text

crates/embucket-functions/src/table/result_scan.rs
```diff
@@ -0,0 +1,143 @@
...
+
+impl TableFunctionImpl for ResultScanFunc {
+    fn call(&self, args: &[(Expr, Option<String>)]) -> DFResult<Arc<dyn TableProvider>> {
+        let query_id = match args.first() {
+            Some((Expr::Literal(ScalarValue::Utf8(Some(query_id))), _)) => utf8_val(query_id),
+            Some((Expr::ScalarFunction(fun), _)) => {
+                if fun.name().to_lowercase() == "last_query_id" {
+                    let index = match fun.args.first() {
+                        Some(Expr::Literal(value)) => value.clone().try_into().unwrap_or(-1),
+                        _ => -1,
```
rampage644, line 108: Does this arm warrant an error instead of default value? Or is it for `last_query_id()` call?

osipovartem: This is only for last_query_id

crates/embucket-functions/src/table/result_scan.rs
```diff
@@ -0,0 +1,143 @@
...
+}
+
+impl TableFunctionImpl for ResultScanFunc {
+    fn call(&self, args: &[(Expr, Option<String>)]) -> DFResult<Arc<dyn TableProvider>> {
+        let query_id = match args.first() {
+            Some((Expr::Literal(ScalarValue::Utf8(Some(query_id))), _)) => utf8_val(query_id),
+            Some((Expr::ScalarFunction(fun), _)) => {
+                if fun.name().to_lowercase() == "last_query_id" {
+                    let index = match fun.args.first() {
+                        Some(Expr::Literal(value)) => value.clone().try_into().unwrap_or(-1),
```
rampage644, line 107: I've written next comment before this one, so should we error out here instead? Am I correct `try_into()` will try to convert str into integeger?

crates/embucket-functions/src/tests/table/result_scan.rs
```diff
@@ -0,0 +1,22 @@
...
+    result_scan_last_query_id_index,
+    "SELECT * FROM result_scan(last_query_id(-1))"
+);
+test_query!(result_scan_query_id, "SELECT * FROM result_scan('1')");
+test_query!(
+    result_scan_query_id_filter,
+    "SELECT * FROM result_scan('1') WHERE a > 1"
+);
+// Missing query
+test_query!(
```
rampage644, line 17: Let's have a test for `result_scan("a")` ?

rampage644 (approved):
overall lgtm

crates/embucket-functions/src/table/result_scan.rs
osipovartem: Last_query_id takes an integer index as param, result_scan takes query_id string

crates/embucket-functions/src/tests/table/result_scan.rs
osipovartem: It should fail since history supports only i64 query_id. Will add

Issue #1035, opened 2025-06-08
`last_value` window function stopped working


rampage644:
Following query doesn't work:

```sql
WITH snowplow_events_sample AS (
  -- This CTE simulates a table of web events.
  -- Each row represents an event with a user, a domain user, and a timestamp.
  SELECT 'd1' AS domain_userid, 'user_a' AS user_id, CAST('2023-10-25 10:00:00' AS TIMESTAMP) AS collector_tstamp UNION ALL
  SELECT 'd1' AS domain_userid, 'user_b' AS user_id, CAST('2023-10-25 12:30:00' AS TIMESTAMP) AS collector_tstamp UNION ALL -- This is the last event for d1
  SELECT 'd1' AS domain_userid, NULL AS user_id, CAST('2023-10-25 11:00:00' AS TIMESTAMP) AS collector_tstamp UNION ALL
  SELECT 'd2' AS domain_userid, 'user_x' AS user_id, CAST('2023-10-26 09:00:00' AS TIMESTAMP) AS collector_tstamp UNION ALL
  SELECT 'd2' AS domain_userid, 'user_y' AS user_id, CAST('2023-10-26 15:00:00' AS TIMESTAMP) AS collector_tstamp UNION ALL -- This is the last event for d2
  SELECT 'd3' AS domain_userid, 'user_p' AS user_id, CAST('2023-10-27 14:00:00' AS TIMESTAMP) AS collector_tstamp UNION ALL
  SELECT 'd3' AS domain_userid, 'user_q' AS user_id, CAST('2023-10-27 13:00:00' AS TIMESTAMP) AS collector_tstamp
)


select distinct
  domain_userid,
  last_value(user_id) over(
    partition by domain_userid
    order by collector_tstamp
    rows between unbounded preceding and unbounded following
  ) as user_id,
  max(collector_tstamp) over (partition by domain_userid) as end_tstamp
from
snowplow_events_sample
where user_id is not null;
```

with an error 
```
[...]

PR #1052, opened 2025-06-10
Add result_scan visitor to replace table(result_scan..

osipovartem:
Related to #1037 
and https://github.com/Embucket/embucket/pull/1048

Added SQL AST visitor that rewrites `TABLE(RESULT_SCAN(...))` table functions into `RESULT_SCAN(...)` by removing the unnecessary `TABLE(...)` wrapper since result scan already returns data as a Table

```sql
 SELECT * FROM TABLE(RESULT_SCAN(LAST_QUERY_ID())) WHERE value > 1;
```
 ```sql
SELECT * FROM RESULT_SCAN(LAST_QUERY_ID()) WHERE value > 1;
 ```

### DIFF
--- a/crates/core-executor/src/query.rs
+++ b/crates/core-executor/src/query.rs
@@ -26,7 +26,7 @@ use datafusion::execution::session_state::SessionState;
 use datafusion::logical_expr::{LogicalPlan, TableSource};
 use datafusion::prelude::CsvReadOptions;
 use datafusion::scalar::ScalarValue;
-use datafusion::sql::parser::{CreateExternalTable, DFParser, Statement as DFStatement};
+use datafusion::sql::parser::{CreateExternalTable, Statement as DFStatement};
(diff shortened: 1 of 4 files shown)


DanCodedThis (approved):
LGTM

crates/core-executor/src/query.rs
```diff
@@ -184,7 +183,8 @@ impl UserQuery {
             })?;
             copy_into_identifiers::visit(value);
             select_expr_aliases::visit(value);
-            inline_aliases_in_query::visit(value);
```
Vedin, line 187: Is this commented out by a reason?

Vedin (approved):

crates/core-executor/src/query.rs
rampage644: This seems unrelated for this PR, the reason though is #1035 (and other related issues)

rampage644 (approved):
lgtm

PR #1057, opened 2025-06-11
Fix result scan JSON Parsing from ResultSet

osipovartem:
**Fix read_query_batches for Correct JSON Parsing from ResultSet**

- This PR fixes an issue in read_query_batches where the function incorrectly attempted to parse the full ResultSet JSON object instead of extracting and transforming the rows field into Arrow-compatible JSON.
- Fixed DataFusion error: Error during planning: table function 'embucket' not found (#1056)

**Summary of Changes:**
- Fixed JSON parsing logic:
  - Extracted the rows field from ResultSet.
  - Combined it with column names from ResultSet.columns to generate line-delimited JSON objects (Arrow-compatible).
  - Replaced usage of the original result_json with the newly constructed Arrow-compatible JSON string during schema inference and batch reading.
- Added helper function convert_resultset_to_arrow_json_lines:
  - Converts Vec<Row> and Vec<Column> into newline-delimited JSON records:
   **{"a": 1, "b": "2", "c": true}
   {"a": 2.0, "b": "4", "c": false}**
- **Added a check related to #1055 to skip  fully qualified names for table functions. This changes fixes flatten, result_scan and any orher registered table  func**
- Added more tests

PR #1062, opened 2025-06-11
Fix result_scan columns for empty records batch

### LINKED ISSUE #1058: [BUG] Fix show objects column names
000200: DataFusion error: Schema error: No field named all_objects.name. Valid fields are all_tables.table_catalog, all_tables.table_schema, all_tables.table_name, all_tables.table_type, all_tables.is_iceberg.

for 
```sql
show objects in snowplow.public_derived
    limit 10000;

select all_objects.*, all_tables.IS_ICEBERG as "is_iceberg"
from table(result_scan(last_query_id(-1))) all_objects
left join snowplow.INFORMATION_SCHEMA.tables as all_tables
on all_tables.table_name = all_objects."name"
and all_tables.table_schema = all_objects."schema_name"
and all_tables.table_catalog = all_objects."database_name";
```

osipovartem:
- **Added schema serialization to ResultSet to handle empty record batches**

  - Previously, when the ResultSet contained no rows (empty record batches), the schema was inferred at runtime using infer_json_schema, which sometimes resulted in an **empty schema**. This caused issues downstream because a valid schema is required even when no data rows are present (e.g., for joins).
  - To address this,  extended the ResultSet struct to include a serialized schema field. This allows the schema to be preserved and reused directly, ensuring consistent and complete schema information regardless of the presence of rows.

- Additionally, **enabled the serde feature for arrow-schema** to support serialization and deserialization of Schema. This is necessary because DataFusion itself does not enable this feature by default, so we had to explicitly activate it in our build to make schema (de)serialization possible.

Closes #1058 #1037
Related to #1057

### DIFF
--- a/crates/embucket-functions/Cargo.toml
+++ b/crates/embucket-functions/Cargo.toml
@@ -5,6 +5,7 @@ edition = "2024"
 license-file.workspace = true
 
 [dependencies]
+arrow-schema = { version = "55", features = ["serde"] }
 chrono = { workspace = true }
 core-history = { path = "../core-history" }
 datafusion = { workspace = true }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -599,6 +599,9 @@ name = "arrow-schema"
 version = "55.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "73a47aa0c771b5381de2b7f16998d351a6f4eb839f1e13d48353e17e873d969b"
+dependencies = [
+ "serde",
+]
 
 [[package]]
 name = "arrow-select"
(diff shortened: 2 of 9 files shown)


crates/embucket-functions/Cargo.toml
```diff
@@ -5,6 +5,7 @@ edition = "2024"
 license-file.workspace = true
 
 [dependencies]
+arrow-schema = { version = "55", features = ["serde"] }
```
rampage644, line 8: I believe you can skip adding this as extra dependency and rely on `datafusion::arrow::schema` transiently

rampage644: Or is it here due to `serde` feature?

osipovartem: This is necessary because DataFusion itself does not enable this feature by default, so we had to explicitly activate it in our build to make schema (de)serialization possible.

rampage644 (approved):

PR #1086, opened 2025-06-12
api: UI Query history for non-existent Worksheet

DanCodedThis:
- Checking if no worksheet exists
- If not, only adding the query record to the global history
- Tests for this already exist

### DIFF
--- a/crates/api-ui/src/queries/handlers.rs
+++ b/crates/api-ui/src/queries/handlers.rs
@@ -97,7 +97,13 @@ pub async fn query(
             .context
             .as_ref()
             .and_then(|c| c.get("schema").cloned()),
-        payload.worksheet_id,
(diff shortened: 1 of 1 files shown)


DanCodedThis:
@YaroslavLitvinov We should probably change the return type of the `.get_worksheet` to `Option<>`, or to `Result<Option<>>`, no need to error if no worksheet found, it could simply be `Ok(None)` what do you think?

DanCodedThis:
This however will make pattern matching a bit more difficult, so maybe not a good idea

YaroslavLitvinov:
If wrong worksheet id is ignored there is no need returning error since query finally was executed. @nikitastryuk what do you think?

YaroslavLitvinov (approved):
LGTM

PR #1105, opened 2025-06-16
Align input logical plan schema with target table schema for CTAs/INSERT

### LINKED ISSUE #1088: CTAS fails when one column being cast to `TIMESTAMP` type
Integration tests fail when running certain models with `000200: DataFusion error: Error during planning: Inserting query must have the same schema with the table.`

Upon digging deeper I was able to reproduce it with simple query:
```sql
CREATE OR REPLACE TABLE t1 AS
SELECT * FROM (VALUES
    ('2021-03-02 15:55:18.539000'::TIMESTAMP)
) AS t(start_tstamp);
```

However, it works if `TIMESTAMP` replaced with `TIMESTAMP_NTZ(9)`

For reference this is integration test failure:

Sample query:
```sql
create or replace transient table embucket.public_snplw_web_int_tests.snowplow_web_consent_cmp_stats_expected_stg
    as (
select
  event_id,
  domain_userid,
  original_domain_userid,
  cast(page_view_id as TEXT) as page_view_id,
  domain_sessionid,
  original_domain_sessionid,
  cmp_load_time,
  cast(cmp_tstamp as timestamp) as cmp_tstamp,
  cast(first_consent_event_tstamp as timestamp) as first_consent_event_tstamp,
  first_consent_event_type,
  cmp_interaction_time
from embucket.public_snplw_web_int_tests.snowplow_web_consent_cmp_stats_expected
    );
```

This is likely due to timestamp fields: removing those lines or just removing casts fixes the problem.

```sql
  cast(cmp_tstamp as timestamp) as cmp_tstamp,
  cast(first_consent_event_tstamp as timestamp) as first_consent_event_tstamp,
```

osipovartem:
Closes #1088

This PR adds a helper function to align input logical plan schemas to the target table schema when executing CTAsor INSERT operations. It automatically inserts projection and casts where necessary to ensure schema compatibility, avoiding runtime type mismatches (e.g. TIMESTAMP(NANOSECOND) → TIMESTAMP(MICROSECOND)).


crates/core-executor/src/query.rs
```diff
@@ -1940,3 +1941,29 @@ fn apply_show_filters(sql: String, filters: &[String]) -> String {
...
+    for field in target_schema.fields() {
+        let name = field.name();
+        let data_type = field.data_type();
+        let input_field = input_schema
+            .field_with_name(name)
+            .context(ex_error::ArrowSnafu)?;
+        if input_field.data_type() == data_type {
+            projections.push(col(name));
+        } else {
+            projections.push(DFExpr::TryCast(TryCast::new(
```
rampage644, line 1961: Perhaps this can be achieved with type coercion update - it does exactly that

crates/core-executor/src/datafusion/type_planner.rs
```diff
@@ -23,14 +23,16 @@ impl TypePlanner for CustomTypePlanner {
...
                     match parsed_b {
                         Some(0) => Ok(Some(DataType::Timestamp(TimeUnit::Second, None))),
                         Some(3) => Ok(Some(DataType::Timestamp(TimeUnit::Millisecond, None))),
                         // We coerce nanoseconds to microseconds as Apache Iceberg v2 doesn't support nanosecond precision
-                        Some(6 | 9) => Ok(Some(DataType::Timestamp(TimeUnit::Microsecond, None))),
-                        _ => not_impl_err!("Unsupported SQL TIMESTAMP_NTZ precision {parsed_b:?}"),
+                        None | Some(6 | 9) => {
+                            Ok(Some(DataType::Timestamp(TimeUnit::Microsecond, None)))
+                        }
+                        _ => not_impl_err!("Unsupported SQL TIMESTAMP_* precision {parsed_b:?}"),
```
rampage644, line 35: Really curios to check if Snowflake actually supports other precision and what the error if not (i.e. 2,5,7)

crates/core-executor/src/query.rs
osipovartem: what do you mean by type coercion?
By analyser rule?

rampage644 (approved):
Is it going to conflict with #1107?

crates/core-executor/src/datafusion/type_planner.rs
osipovartem: All timestamp variations, as well as the TIMESTAMP alias, support an optional precision parameter for fractional seconds (for example, TIMESTAMP(3)). Timestamp precision can range from 0 (seconds) to 9 (nanoseconds). The default precision is 9.
https://docs.snowflake.com/en/sql-reference/data-types-datetime#timestamp

PR #1124, opened 2025-06-17
`lower` scalar function

### LINKED ISSUE #1119: BUG: `lower` scalar function doesn't work with numbers
DDL
```sql
create table EMBUCKET.salesforce_v2_stitch.casehistory
(
    caseid             integer,
    createdbyid        integer,
    createddate        timestamp without time zone,
    datatype           integer,
    field              integer,
    id                 integer,
    isdeleted          boolean,
    newvalue           integer,
    oldvalue           integer,
    _sdc_batched_at    timestamp without time zone,
    _sdc_extracted_at  timestamp without time zone,
    _sdc_received_at   timestamp without time zone,
    _sdc_sequence      integer,
    _sdc_table_version integer,
    oldvalue__fl       float8
)
```

query 
```sql
WITH base AS (

    SELECT *
    FROM {{ source('salesforce', 'casehistory') }}

), renamed AS (

    SELECT
      caseid::VARCHAR                AS case_id,
      id::VARCHAR                    AS case_history_id,
      createddate::DATE           AS field_modified_at,
      LOWER(field)::VARCHAR          AS case_field,
      LOWER(datatype)::VARCHAR       AS data_type,
      newvalue::VARCHAR              AS new_value,
      oldvalue::VARCHAR              AS old_value,
      isdeleted::BOOLEAN             AS is_deleted,
      createdbyid::VARCHAR           AS created_by_id
    FROM base

)

SELECT *
FROM renamed
```

produces error

```
DataFusion error: Error during planning: Internal error: Expect TypeSignatureClass::Native(LogicalType(Native(String), String)) but received NativeType::Int32, DataType: Int32.
This was likely caused by a bug in DataFusion's code and we would welcome that you file an bug report in our issue tracker No function matches the given name and argument types 'lower(Int32)'. You might need to add explicit type casts.
    Candidate functions:
    lower(Coercion(TypeSignatureClass::Native(LogicalType(Native(String), String))))
```

Why?

`lower` doesn't support numeric data types, only strings. But for some reason some of the workloads contain lower function applied to integers. Snowflake doc https://docs.snowflake.com/en/sql-reference/functions/lower states that argument must be a string but it also work with numbers.

ravlio:
closes #1119


Vedin (approved):

crates/embucket-functions/src/tests/string_binary/lower.rs
```diff
@@ -0,0 +1,54 @@
...
+        LOWER('ΕΛΛΗΝΙΚΆ') AS greek,
+        LOWER('Ελληνικά') AS mixed_greek",
+    snapshot_path = "lower"
+);
+
+// Numeric and non-string types
+test_query!(
+    non_string_types,
+    "SELECT 
+        LOWER(123) AS number,
```
Vedin: I'm just curious: Why do you cast here? Maybe it's worth testing that it really returns a string type after lower and not returns the same type as the original value?

rampage644 (approved):

crates/embucket-functions/src/tests/string_binary/lower.rs
ravlio: The function always returns a string, whether it's originally a string or a number. Tests cover all cases. Or maybe I misunderstood your question.

rampage644: I think @Vedin asked why have `as number/float/decimal` in the query?

rampage644: Perhaps this is to easier test in snapshot, and this is not a cast afaiu

rampage644 (commented):
I've played with snowflake a bit and read the docs (which state btw that it only accepts string expr), but actually works with other datatypes as well.

Would it be perhaps more interesting to use `simplify` here to return explicit cast?

crates/embucket-functions/src/tests/string_binary/lower.rs
ravlio: removed aliases

ravlio:
@rampage644 I made a simplification

rampage644 (approved):

PR #1160, opened 2025-06-19
Add auth headers to upload script

osipovartem:


rampage644:
I think I was fixing another upload script somewhere in `test/` folder. There I dropped UI endpoints altogether and instead used snowflake connector to do last mile query setup (create schema).

rampage644:
#1067

PR #1163, opened 2025-06-19
Succeed if already set up

rampage644:


rampage644 (commented):
There is another failure mode in the setup routine I couldn't fix quickly - UI endpoints require authentication token. The solution is to use snowflake connector to run queries.

rampage644:
Had to restore bad commit from claude

andheroe (approved):

YaroslavLitvinov (approved):

PR #1166, opened 2025-06-20
Issues/0921 execution time of query grows linearly

### LINKED ISSUE #921: Execution time of query grows linearly
When `Embucket` service is executing many requests, the execution time of each individual request constantly grows.

It starts from milliseconds, but after handling around 1000 requests, the execution time increases to as much as 2 seconds.

YaroslavLitvinov:
Closes #921


crates/df-catalog/src/catalog_list.rs
```diff
@@ -31,6 +31,12 @@ use url::Url;
 
 pub const DEFAULT_CATALOG: &str = "embucket";
 
+#[derive(Clone, Debug, PartialEq)]
+pub enum CachedEntity {
```
osipovartem, line 36: Lets add ident to this enum
like Schema(SchemaIdent), Table(TableIdent) 
to remove names: (&str, &str, &str)
Maybe Schema(Ident), Table(Ident) with optional fields inside ident

crates/df-catalog/src/catalog_list.rs
```diff
@@ -180,15 +186,92 @@ impl EmbucketCatalogList {
...
+        }
+        let schema_ref = catalog_ref.schemas_cache
+            .get(schema)
+            .ok_or_else(schema_not_found)?;
+
+        let table_not_found = ||InvalidCacheSnafu {
+            entity: "table", name: format!("{catalog}.{schema}.{table}") }.build();
+        if entity == CachedEntity::Table {
+            if drop {
+                schema_ref.tables_cache.remove(table).ok_or_else(table_not_found)?;
```
osipovartem: we don't need to call ok_or_else here (just remove even if missing)

YaroslavLitvinov: The idea here to raise error if nothing removed , that is why .ok_or_else(table_not_found)?

Since cache updated granularly it should be consistent, I think missing cache values would mean our cache is not valid.

osipovartem: The cache can be already empty and it is ok

crates/df-catalog/src/catalog_list.rs
```diff
@@ -31,6 +32,18 @@ use url::Url;
...
+#[derive(Debug, Eq, PartialEq)]
+pub enum CachedEntity {
+    Schema(SchemaIdent),
+    Table(TableIdent),
+}
+
+#[derive(Debug, Eq, PartialEq)]
+pub enum CachedEntityOp {
+    Create,
+    Delete,
```
osipovartem: Let's align with statements (delete->drop)

crates/core-executor/src/query.rs
```diff
@@ -134,6 +136,42 @@ impl UserQuery {
...
+            .downcast_ref::<EmbucketCatalogList>()
+        {
+            let res = catalog_list_impl
+                .refresh_cache(op, entity)
+                .await
+                .context(RefreshCatalogListSnafu);
+
+            // Not sure we need this, but just in case
+            // TODO: Remove following block when refresh_catalog removed
+            if let Err(ExecutionError::RefreshCatalogList { source, .. }) = res {
```
osipovartem, line 161: Do we you async fn refresh_catalog(&self) -> ExecutionResult<()> after this change somewhere?
if not let's remove and keep it clear (same as refresh_catalog)

crates/core-executor/src/query.rs
```diff
@@ -286,13 +333,13 @@ impl UserQuery {
...
                 Statement::CreateSchema { .. } => {
                     let result = Box::pin(self.create_schema(*s)).await;
-                    self.refresh_catalog().await?;
                     return result;
                 }
                 Statement::CreateStage { .. } => {
                     // We support only CSV uploads for now
                     let result = Box::pin(self.create_stage_query(*s)).await;
-                    self.refresh_catalog().await?;
+                    // It seems `refresh_catalog()` is not needed here since we already
```
osipovartem: let's just remove refresh_catalog from here and remove the comment

crates/core-executor/src/query.rs
YaroslavLitvinov: If I understood correctly question about  remained refresh_catalog.
Since it may be some surprises with granular updates, so in case of cache error this fallback implementation will be called to minimize chances something will break. 
In case it proves its durability `refresh_catalog` could be removed

crates/df-catalog/src/catalog_list.rs
```diff
@@ -181,14 +194,114 @@ impl EmbucketCatalogList {
...
+    }
+
+    #[allow(clippy::as_conversions, clippy::too_many_lines)]
+    #[tracing::instrument(
+        name = "EmbucketCatalogList::refresh_schema",
+        level = "debug",
+        skip(self),
+        err
+    )]
+    pub async fn refresh_cache(&self, op: CachedEntityOp, entity: CachedEntity) -> Result<()> {
```
osipovartem: WDYT about just refresh_entity() without CachedEntityOp
we just need to take latest state of table/schema

DROP table -> just check if table exists in original source. remove from cache if not, update cache if exists
CREATE table ->  just check if table exists in original source. remove from cache if not, update cache if exists
Same for Schema, this is related to other statements like ALTER TABLE where we just update the table withou creating/removing. We just need to get latest state

YaroslavLitvinov: So behavior for `.invalidate_cache(entity)` could be:
For schema: Do nothing - if exists in both metastore and cache; Delete from cache - if not exists in metastore.
For table: Recreate in cache - if exists in both metastore and cache; Delete from cache - if not.

YaroslavLitvinov: done

rampage644 (approved):

PR #1171, opened 2025-06-20
Allow view registration

### LINKED ISSUE #1168: [BUG] Allow view registration
```sql
 create or replace  view
```

errror `schema provider does not support registering table`

Currently we use default implementation which calls table registration (usually in memory catalog).
We need to allow view registration in CachingSchema even if ScemaProvider doesn't support table registration since this is only in-memory view
```rust
async fn create_view(&self, cmd: CreateView) -> Result<DataFrame> {
        let CreateView {
            name,
            input,
            or_replace,
            definition,
            temporary,
        } = cmd;

        let view = self.table(name.clone()).await;

        if temporary {
            return not_impl_err!("Temporary views not supported");
        }

        match (or_replace, view) {
            (true, Ok(_)) => {
                self.deregister_table(name.clone())?;
                let input = Self::apply_type_coercion(input.as_ref().clone())?;
                let table = Arc::new(ViewTable::new(input, definition));
                self.register_table(name, table)?;
                self.return_empty_dataframe()
            }
            (_, Err(_)) => {
                let input = Self::apply_type_coercion(input.as_ref().clone())?;
                let table = Arc::new(ViewTable::new(input, definition));
                self.register_table(name, table)?;
                self.return_empty_dataframe()
            }
            (false, Ok(_)) => exec_err!("Table '{name}' already exists"),
        }
    }
```

osipovartem:
**Closes #1168**

This change enables registering views within the i**n-memory session cache** (tables_cache) of a schema. Views are stored only in memory and are not registered in the underlying schema/catalog, which aligns with expected behavior for session-scoped objects.

### Changes
- When register_table(...) is called with a table of type TableType::View, it is inserted only into tables_cache, without calling self.schema.register_table(...).
- For all other table types, the table is cached and also delegated to the underlying schema.

### Motivation
- In many SQL engines, views created within a session (e.g. via CREATE VIEW) are ephemeral and should not persist beyond the session scope. This implementation provides initial support for such session-local views.

### Limitations
- These views will disappear after catalog list recreation, as they are not persisted across sessions or stored in any durable metastore.
- To support persistent or shared views, we would need to make the catalog list itself shared and stateful across sessions — e.g., via long-lived application state or shared metadata store. (related to #699)

This code is used by default for VIEWs
```rust
    async fn create_view(&self, cmd: CreateView) -> Result<DataFrame> {
        let CreateView {
            name,
            input,
            or_replace,
            definition,
            temporary,
        } = cmd;

        let view = self.table(name.clone()).await;

        if temporary {
            return not_impl_err!("Temporary views not supported");
        }

        match (or_replace, view) {
            (true, Ok(_)) => {
                self.deregister_table(name.clone())?;
                let input = Self::apply_type_coercion(input.as_ref().clone())?;
                let table = Arc::new(ViewTable::new(input, definition));
                self.register_table(name, table)?;
                self.return_empty_dataframe()
            }
            (_, Err(_)) => {
                let input = Self::apply_type_coercion(input.as_ref().clone())?;
                let table = Arc::new(ViewTable::new(input, definition));
                self.register_table(name, table)?;
                self.return_empty_dataframe()
            }
            (false, Ok(_)) => exec_err!("Table '{name}' already exists"),
        }
    }


DanCodedThis (approved):
@osipovartem LGTM

PR #1174, opened 2025-06-20
api: UI inconsistent schemas returned by handler

### LINKED ISSUE #1084: api: Backend returns inconsistent list of schemas. Different result by /ui/navigation-tree vs /ui/databases/{database}/schemas
Our `/ui/navigation-tree` handler returns 6 schemas in selected database, but GET /ui/databases/{database}/schemas returns just 5:  `information_schema` is missing, like showed on picture:
Image

In case we select 'slatedb' /ui/databases/{database}/schemas return 0 schemas, but navigation-tree returned 3.
Image

DanCodedThis:
Closes #1084

- Added slatedb schemas (`meta`, `history`) to the handler
- Added `information_schema` to the handler
- Used `UNION ALL` in a subquery to stitch everything together, to be able to use search, ordering, etc, instead of just modifying the json result
- Fixed tests to accommodate +1 information schema to each database (catalog)

**Would be great** to hear about using the `chrono::UTC::now()` since `slatedb` schemas are recrearted each time (views) I figured it's okay. They don't really have a consistent creation time (maybe the start of service, and only change the updated_at).

### DIFF
--- a/crates/api-ui/src/schemas/handlers.rs
+++ b/crates/api-ui/src/schemas/handlers.rs
@@ -291,8 +291,31 @@ pub async fn list_schemas(
     Path(database_name): Path<String>,
 ) -> Result<Json<SchemasResponse>> {
     let context = QueryContext::new(Some(database_name.clone()), None, None);
+    let now = chrono::Utc::now().to_string();
+    let sql_history_schema = format!(
+        "UNION ALL SELECT 'history' AS schema_name, 'slatedb' AS database_name, '{}' AS created_at, '{}' AS updated_at",
+        now.clone(),
+        now.clone()
+    );
+    let sql_meta_schema = format!(
+        "UNION ALL SELECT 'meta' AS schema_name, 'slatedb' AS database_name, '{}' AS created_at, '{}' AS updated_at",
+        now.clone(),
+        now.clone()
+    );
+    let sql_information_schema = match database_name.as_str() {
+        "slatedb" => format!(
+            "UNION ALL SELECT 'information_schema' AS schema_name, 'slatedb' AS database_name, '{}' AS created_at, '{}' AS updated_at",
+            now.clone(),
+            now.clone()
+        ),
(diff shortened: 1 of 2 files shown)


crates/api-ui/src/schemas/handlers.rs
```diff
@@ -291,8 +291,31 @@ pub async fn list_schemas(
     Path(database_name): Path<String>,
 ) -> Result<Json<SchemasResponse>> {
     let context = QueryContext::new(Some(database_name.clone()), None, None);
+    let now = chrono::Utc::now().to_string();
```
YaroslavLitvinov, line 294: I asked Copilot ro review the PR, and here is the suggestion:

```
let now = chrono::Utc::now().to_rfc3339();

fn synthetic_schema_sql(name: &str, database_name: &str, now: &str) -> String {
    format!(
        "UNION ALL SELECT '{}' AS schema_name, '{}' AS database_name, '{}' AS created_at, '{}' AS updated_at",
        name, database_name, now, now
    )
}

let sql_history_schema = synthetic_schema_sql("history", "slatedb", &now);
let sql_meta_schema = synthetic_schema_sql("meta", "slatedb", &now);

let sql_information_schema = match database_name.as_str() {
    "slatedb" => synthetic_schema_sql("information_schema", "slatedb", &now),
    _ => "UNION ALL SELECT 'information_schema' AS schema_name, database_name, created_at, updated_at FROM slatedb.meta.databases".to_string(),
};

let sql_string = format!(
    "SELECT * FROM (SELECT * FROM slatedb.meta.schemas {sql_history_schema} {sql_meta_schema} {sql_information_schema})"
);

let sql_string = format!(
    "{} WHERE database_name = '{}'",
    sql_string, database_name
);
```
```suggestion
    let now = chrono::Utc::now().to_string();
```

YaroslavLitvinov (approved):

Issue #781, opened 2025-05-13
Add support for `LATERAL FLATTEN`


Vedin:
It's no separate table function `FLATTEN` https://docs.snowflake.com/en/sql-reference/functions/flatten. To use it we need `TABLE` or `LATERAL`. 
We have such steps to implement it:
1. Support syntax for FLATTEN (done).
2. Make sure Flatten works with JOIN.
3. Support for `LATERAL` with correlated subqueries.
4. Support for `LATERAL FLATTEN`

ravlio:
> It's no separate table function FLATTEN

@Vedin we have https://github.com/Embucket/embucket/blob/main/crates/df-builtins/src/table/flatten.rs Or what do you mean?

Vedin:
> > It's no separate table function FLATTEN
> 
> [@Vedin](https://github.com/Vedin) we have https://github.com/Embucket/embucket/blob/main/crates/df-builtins/src/table/flatten.rs Or what do you mean?

I meant that in Snowflake, you can't use FLATTEN directly. To see a result, you should wrap it in the TABLE function or use it with LATERAL.

osipovartem:
The reason of the issue that we doesn't support table faction as relation.
https://github.com/apache/datafusion/blob/main/datafusion/sql/src/relation/mod.rs#L157
Both TableFunction and Function

```sql
create table test (jsontext text);
insert into table test (jsontext) values('{"key": "value"}');
 SELECT
  jsontext
FROM  test
INNER JOIN LATERAL  flatten(INPUT => PARSE_JSON(jsontext)) d;
```

can be implemented like
```rust
            TableFactor::Function {
                lateral,
                name,
                args,
                alias,
            } => {
                let tbl_func_name = self.object_name_to_table_reference(name)?;
                let schema = DFSchema::empty();
                let func_args = args
                    .into_iter()
                    .map(|arg| match arg {
                        FunctionArg::Unnamed(FunctionArgExpr::Expr(expr)) => {
                            let expr = self.sql_expr_to_logical_expr(expr, &schema, planner_context)?;
                            Ok((expr, None))
                        }
                        FunctionArg::Named { name, arg: FunctionArgExpr::Expr(expr), .. } => {
                            let expr = self.sql_expr_to_logical_expr(expr, &schema, planner_context)?;
                            Ok((expr, Some(name.value.clone())))
                        }
                        _ => plan_err!("Unsupported function argument: {arg:?}"),
                    })
[...]

PR #1202, opened 2025-06-25
Extend flatten to work as relation 

osipovartem:
Related to #781


crates/core-executor/src/query.rs
```diff
@@ -1368,7 +1368,7 @@ impl UserQuery {
                     let is_table_func = self
                         .session
                         .ctx
-                        .table_function(table_name.to_string().to_lowercase().as_str())
+                        .table_function(table_name.to_string().to_ascii_lowercase().as_str())
```
Vedin, line 1371: is this still needed after DF change?

crates/embucket-functions/src/table/flatten/func.rs
```diff
@@ -512,14 +406,14 @@ mod tests {
                 "+-----+-----+------+-------+-------+-----------+",
                 "| SEQ | KEY | PATH | INDEX | VALUE | THIS      |",
                 "+-----+-----+------+-------+-------+-----------+",
-                "| 1   | a   | a    |       | 1     | {         |",
+                "| 1   | a   | a    | None  | 1     | {         |",
```
Vedin: I can be wrong, but I assumed that this should be empty in case of null. None is a little bit confusing as output here

test/dbt_integration_tests/dbt-gitlab/charts/dbt_run_status.png
Vedin, line 1: Accidentally added?

crates/embucket-functions/src/table/flatten/func.rs
osipovartem: Snowflake cli returns None for this case

crates/core-executor/src/query.rs
osipovartem: DF change was related to ContextProvider
here we call
```rust
    pub fn table_function(&self, name: &str) -> Result<Arc<TableFunction>> {
        self.state
            .read()
            .table_functions()
            .get(name)
            .cloned()
            .ok_or_else(|| plan_datafusion_err!("Table function '{name}' not found"))
    }

Vedin (approved):

Issue #1152, opened 2025-06-19
[BUG] Tables sorted in reverse order in Data table preview


osipovartem:
<img width="846" alt="Image" src="https://github.com/user-attachments/assets/d9eb30f8-3b74-4f03-a095-fd75fc412c8d" />

PR #1206, opened 2025-06-25
api: UI Fixed default sort order for list handlers

### LINKED ISSUE #1204: api: UI Default sort order
Default sort order for UI list handlers should be not `DESC` by `name`, but by `created_at`

DanCodedThis:
Closes #1204 & #1152 

- Changed the `fn apply_parameters` from using the first `search_column` as the default `order by` to `created_at`
- All handlers that use the `SearchParameters` are now affected (volumes, databases, schemas, tables)
- Fixed tests for databases, since they are the ones that delete and then add back the same value, so by `order by created_at` the value that was first and was deleted and added back should be at the back (which it is)

### DIFF
--- a/crates/api-ui/src/lib.rs
+++ b/crates/api-ui/src/lib.rs
@@ -113,12 +113,7 @@ fn apply_parameters(
     );
     //Default order by is the first search column or created at
     let sql_string = parameters.order_by.map_or_else(
-        || {
-            format!(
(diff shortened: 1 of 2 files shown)


DanCodedThis:
@nikitastryuk @osipovartem would be great if you guys checked

osipovartem (approved):

PR #1207, opened 2025-06-25
embucketd: Bootstrapping default: volume, database & schema

### LINKED ISSUE #1197: Add an option to create default volume, database and schema
For seamless user experience it would be great to have a flag (or have a flag to opt-out) to create an in-memory volume, default database (`embucket`) and default schema (`public`).

DanCodedThis:
Closes #1197 

- Added new Cli flag: `--no-bootstrap`
- If `--no-bootstrap` nothing happens
- If there is no flag, then we create "embucket" memory volume, "embucket" database and "public" schema
- With `cargo run` it's `cargo run -- --no-bootstrap`
- There are errors in tracing even though I don't error if it already exists for obvious reasons. Would be great if it didn't trace this I guess
- Also, should we fail, if for any other reason the bootstrapping fails? For now, it only traces any other error

### DIFF
--- a/crates/embucketd/src/main.rs
+++ b/crates/embucketd/src/main.rs
@@ -26,6 +26,7 @@ use axum::{
     routing::{get, post},
 };
 use clap::Parser;
+use core_executor::catalog::catalog_list::DEFAULT_CATALOG;
 use core_executor::service::CoreExecutionService;
 use core_executor::utils::Config as ExecutionConfig;
 use core_history::SlateDBHistoryStore;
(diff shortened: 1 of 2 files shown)


DanCodedThis:
@andheroe For some reason even though it didn't do the SLT testing it wrote a negative comment :)

andheroe:
I see that 6 tests were executed

DanCodedThis:
@andheroe I was changing **embucked** crate. Could the default volume, database & schema affect the SLT's?

crates/embucketd/src/main.rs
```diff
@@ -125,6 +130,53 @@ async fn main() {
     ));
 
     let metastore = Arc::new(SlateDBMetastore::new(db.clone()));
+    if !no_bootstrap {
```
rampage644: Let's refactor this out to the separate function not to clutter the wiring logic in the `main()`

crates/embucketd/src/main.rs
```diff
@@ -125,6 +130,53 @@ async fn main() {
...
+        let ident = "embucket".to_string();
```
rampage644: There should be constant in `catalog` crate with default ident

andheroe:
@DanCodedThis there are false positives in SLT selection automation. We are improving it step by step.
In this case, this is how LLM explained the selection of tests:
`EXPLANATION: The code changes involve the addition of a new command-line option to disable bootstrap functionality and modifications to the main function to conditionally execute database, volume, and schema creation based on this new option. The changes specifically affect the creation and initialization of volumes, databases, and schemas in the metastore. Therefore, tests related to database, volume, and schema operations are relevant to ensure that these functionalities are not adversely affected by the new bootstrap option.`
You can always check the logs of the job if you're not sure why SLTs were selected / not selected. For this PR the job is https://github.com/Embucket/embucket/actions/runs/15884985191/job/44794597333. 
In this particular case, SLTs selection is false positive.

DanCodedThis:
@andheroe Thanks!

crates/embucketd/src/main.rs
```diff
@@ -419,3 +376,59 @@ fn load_openapi_spec() -> Option<openapi::OpenApi> {
...
 }
+
+///This function bootstraps the service if no flag is present (`--no-bootstrap`) with:
+/// 1. Creation of a default in-memory volume named `embucket`
+/// 2. Creation of a default database `embucket` in the volume `embucket`
+/// 3. Creation of a default schema `public` in the database `embucket`
+///
+/// Only traces the errors, doesn't panic.
+async fn bootstrap(metastore: Arc<dyn Metastore>, no_bootstrap: bool) {
+    if !no_bootstrap {
```
rampage644:
```rust
if no_boostrap {
  return
}
...
code continues with one ident less
...
```

rampage644 (approved):

osipovartem (approved):

PR #1219, opened 2025-06-26
Split functions registration to categories 

### LINKED ISSUE #1145: Split udf registration into proper categories
To reduce the size of the register function, I propose registering them in the module where they are placed and then registering the whole module.

Vedin:
closes #1145 

I have now put everything in alphabetical order. I'm not sure that's the right approach, and is it worth being tested?


DanCodedThis (approved):
@Vedin LGTM

PR #1220, opened 2025-06-26
api: sessions snowflake v1 rest auth with demo credentials

### LINKED ISSUE #942: Add authorization for iceberg rest api and v1 rest api

DanCodedThis:
Closes #942 

- v1 snowflake rest api now uses demo credentials, just as the **api-ui**
- Added credentials to an `Auth` struct
- Added `Auth` to the rest config 
- In profiles for `snowplow`, for password we use `SNOWFLAKE_SECRET` but my **.env** was having `SNOWFLAKE_PASSWORD`, change one or the other for easy of use
- Added logic to check the credentials just like in the **auth** for **api-ui**
- Changed session impl to still rely on the api/interface of the tower_sessions, but inside it doesn't use it.
- We only use our list of sessions from the execution_svc. No list discrepancies from the tower_sessions.
- We don't send cookies in snowflake rest, only in UI rest for sessions.
- Snowflake rest auth test is added.
- Session expiration test is fixed.
- UI auth incorporates the new session id use without relying on it for auth (no unification of auths).


crates/api-snowflake-rest/src/handlers.rs
```diff
@@ -49,9 +49,16 @@ pub async fn login(
...
 
     // Deserialize the JSON body
-    let _body_json: LoginRequestBody =
+    let body_json: LoginRequestBody =
         serde_json::from_str(&s).context(api_snowflake_rest_error::LoginRequestParseSnafu)?;
 
+    if body_json.data.login_name != *state.config.auth.demo_user
+        || body_json.data.password != *state.config.auth.demo_password
+    {
+        tracing::error!("{} - {}", body_json.data.login_name, body_json.data.password);
```
osipovartem: let's remove any passwords from tracing

YaroslavLitvinov:
It's worth adding some tests for auth feature

DanCodedThis:
@osipovartem are there any tests for this rest?

DanCodedThis:
> It's worth adding some tests for auth feature

@YaroslavLitvinov not sure if it's gonna be easy to replicate the dbt rest call, but I did test the dbt with it, the login sequence works as expected, if the credentials differ we can't login and fail to execute even **dbt seed**

YaroslavLitvinov:
> > It's worth adding some tests for auth feature
> 
> @YaroslavLitvinov not sure if it's gonna be easy to replicate the dbt rest call, but I did test the dbt with it, the login sequence works as expected, if the credentials differ we can't login and fail to execute even **dbt seed**

You can mock rest-client, with automock for instance.

DanCodedThis:
> > > It's worth adding some tests for auth feature
> > 
> > 
> > @YaroslavLitvinov not sure if it's gonna be easy to replicate the dbt rest call, but I did test the dbt with it, the login sequence works as expected, if the credentials differ we can't login and fail to execute even **dbt seed**
> 
> You can mock rest-client, with automock for instance.

I'll look into it, thanks

crates/api-snowflake-rest/src/handlers.rs
```diff
@@ -49,9 +49,15 @@ pub async fn login(
...
 
     // Deserialize the JSON body
-    let _body_json: LoginRequestBody =
+    let body_json: LoginRequestBody =
         serde_json::from_str(&s).context(api_snowflake_rest_error::LoginRequestParseSnafu)?;
 
+    if body_json.data.login_name != *state.config.auth.demo_user
+        || body_json.data.password != *state.config.auth.demo_password
+    {
+        return api_snowflake_rest_error::InvalidAuthDataSnafu.fail()?;
```
rampage644, line 56: What happens when `query` endpoint is called after failed `login`? Does it make sense to reuse other endpoints auth flow or code?

rampage644: `let token = Uuid::new_v4().to_string();` is not stored anywhere to test against later

crates/api-snowflake-rest/src/schemas.rs
```diff
@@ -151,15 +151,29 @@ impl From<ColumnInfoModel> for ColumnInfo {
...
 }
 
 impl Config {
     pub fn new(data_format: &str) -> std::result::Result<Self, strum::ParseError> {
         Ok(Self {
             dbt_serialization_format: DataSerializationFormat::try_from(data_format)?,
+            ..Self::default()
         })
     }
+    pub fn with_demo_credentials(&mut self, demo_user: String, demo_password: String) {
```
rampage644: `with_` usually returns itself to allow chaining:
```rust
let a = Config::default().with_something().with_something_else().with_another_thing()
```

YaroslavLitvinov:
BTW, I don't know all the requirements, but wouldn't be enough just moving auth/layer middleware `require_auth` prior v1, so it could transparently enabvle auth model for v1 too ?

crates/api-snowflake-rest/src/handlers.rs
DanCodedThis: I'll try to use the auth from UI, and see what happens

DanCodedThis:
> BTW, I don't know all the requirements, but wouldn't be enough just moving auth/layer middleware `require_auth` prior v1, so it could transparently enabvle auth model for v1 too ?

Possibly, I'm not entirely sure

DanCodedThis:
@YaroslavLitvinov not exactly the same, but at least I'll take inspiration from there

crates/api-snowflake-rest/src/handlers.rs
DanCodedThis: @rampage644 We can't just add our **api-ui** middleware. If the `query` is called after `login` fails, it won't have a token and will fail. But we currently don't check if the token is correct or not, we just check its existence.

DanCodedThis:
> > > It's worth adding some tests for auth feature
> > 
> > 
> > @YaroslavLitvinov not sure if it's gonna be easy to replicate the dbt rest call, but I did test the dbt with it, the login sequence works as expected, if the credentials differ we can't login and fail to execute even **dbt seed**
> 
> You can mock rest-client, with automock for instance.

This however won't be so easy, it sends bytes in body. I'll try and see

DanCodedThis:
@osipovartem dbt snowplow creates 3-4 `DedicatedExecutors` for some reason instead of just reusing one. (file handle leakage flashbacks)

YaroslavLitvinov:
It seems PR name : "api: v1 rest auth with demo credentials" is misleading, since no changes to v1 rest are  made, but related to snowflake-rest

crates/api-sessions/src/session.rs
```diff
@@ -152,21 +146,85 @@ where
...
+            .await
+            .context(SessionPersistSnafu)?;
+        session.save().await.context(SessionPersistSnafu)?;
+        Ok(Self(id))
+    }
+}
+
+#[must_use]
+pub fn extract_token(headers: &HeaderMap) -> Option<String> {
+    //First we check the header, second the cookie
```
YaroslavLitvinov, line 189: Could we reuse cookie parser?
crates/api-ui/src/auth/handlers.rs

crates/api-ui/src/auth/layer.rs
```diff
@@ -47,5 +75,28 @@ pub async fn require_auth(
...
+
+        // Step 3: Check if the token is in the session data
+        if !sessions.contains_key(&token) {
+            drop(sessions);
+            //If no session_id, get a new one to the extractor,
+            // in a way that we can provide it to the response header for the browser also
+            let session_id = uuid::Uuid::new_v4().to_string();
+            req.extensions_mut().insert(DFSessionId(session_id.clone()));
+            let mut res = next.run(req).await;
+            set_headers_in_flight(
```
YaroslavLitvinov, line 92: set_cookies function seems doing the same

crates/api-ui/src/auth/handlers.rs
```diff
@@ -111,11 +114,11 @@ fn ensure_jwt_secret_is_valid(jwt_secret: &str) -> Result<()> {
     Ok(())
 }
 
-fn set_cookies(headers: &mut HeaderMap, refresh_token: &str) -> Result<()> {
+fn set_cookies(headers: &mut HeaderMap, name: &str, token: &str) -> Result<()> {
```
YaroslavLitvinov, line 117: it makes sense to extend this cookie setter (which I forgot adding initially) with .max_age(Duration)

crates/api-snowflake-rest/src/auth.rs
YaroslavLitvinov, line 1: I would follow naming convention for naming middleware related files directly with "layer" in its name auth.rs -> layer_auth.rs or so

DanCodedThis:
Since `UserSession` is in an `Arc`, we can probably avoid an `Arc` around the expiry `Mutex`, not sure

YaroslavLitvinov (approved):
LGTM with some notes I left

crates/api-sessions/src/session.rs
DanCodedThis: @YaroslavLitvinov It uses it internally, it's the initial function that Artem wrote. It works correctly for the Auth header of Snowflake rest, I extended it with your `fn cookier_parser`. 

P.S. If you're talking about just the code duplication, it's the case of what's whose dependency. **Api-ui** depends on **api-session**, not the other way around. We can take this function from session for the auth layer in UI (remove it from auth), but that's a level of indirection I decided to avoid in turn for code duplication for now.

crates/api-ui/src/auth/layer.rs
DanCodedThis: @YaroslavLitvinov The result type is different, I had a hard time with trying to make it work with `.context()`. Since it will do something like. Return **api-ui** error, map it to **ui auth** error inside the function itself, and then map it again to **api-ui** error. Or the other way around, in either case it won't provide the right error type because of different result type. 

tl;dr error type doesn't map correctly

crates/api-ui/src/auth/handlers.rs
DanCodedThis: @YaroslavLitvinov Won't we have two sources of truth for the expiration? It'll be similar to what we had with the `tower_sessions`. It could be an edge case with premature deletion of the cookie.

crates/api-snowflake-rest/src/auth.rs
DanCodedThis: @YaroslavLitvinov Since we provide the router here and the middleware. I decided to name it auth, since in UI we have a folder for auth and in main when we use for ui:
- `use api-ui::auth::router as ui_auth_router;`
- `use api-snowflake-rest::auth::router as snowflake_auth_router;`

We can split the file for layer of middleware and router for the auth file. Since we already have a router file (or put it ther under a different name).

DanCodedThis:
For anybody reviewing and who checked and got the idea of the obscure/tricky branch (`DanCodedThis/942-v1-rest-auth-self-contained`), I would also add, that we can propagate such idea to here. For example, I can make a middleware that will from the extension of the request propagate the Set Cookie if needed. So we could possibly be able to remove the id generation from the auth middleware layer. But It would add a level of indirection where yes, the id generation will be self-contained for the extractor. We would just pass an enum to it like: `SessionManagment::Create` or `SessionManagement::GetOrCreate`. For the sake of brevity, I can already think of a many ways to improve it from the logical (self-contained id generation), developer (message passing for the session layer), performance (no locking in the middleware) perspective, etc. 

tl;dr let's just go with an iterative approach (merge this) instead of a completionsist one, and not make perfect be the enemy of good, because it won't change the approach in a meaningful (correctness) type of way

YaroslavLitvinov:
This PR expected to be a breaking change since it uses a different cookie name.
Session cookie name changed in this PR from "id" to "session_id"
All the existing code supporting old sessions cookie will stop working.
@DanCodedThis

crates/api-sessions/src/session.rs
```diff
@@ -1,35 +1,36 @@
...
+    ExpiredDeletion, Expiry, Session, SessionStore,
     session::{Id, Record},
     session_store,
 };
 use uuid;
 
-pub type RequestSessionMemory = Arc<Mutex<HashMap<Id, Record>>>;
+pub const SESSION_ID_COOKIE_NAME: &str = "session_id";
+
+pub const SESSION_EXPIRATION_SECONDS: u64 = 60;
```
osipovartem, line 22: why only 60 seconds?

crates/api-sessions/src/session.rs
```diff
@@ -1,35 +1,36 @@
...
 #[allow(clippy::missing_const_for_fn)]
 impl RequestSessionStore {
-    pub fn new(
-        store: Arc<Mutex<HashMap<Id, Record>>>,
-        execution_svc: Arc<dyn ExecutionService>,
-    ) -> Self {
+    pub fn new(execution_svc: Arc<dyn ExecutionService>) -> Self {
         Self {
-            store,
+            //store,
```
osipovartem: remove or uncomment

crates/api-sessions/src/session.rs
```diff
@@ -51,51 +52,38 @@ impl RequestSessionStore {
...
-                .create_session(df_session_id.to_string())
-                .await
-                .map_err(|e| session_store::Error::Backend(e.to_string()))?;
+            let sessions = self.execution_svc.get_sessions().await;
+
+            let mut sessions = sessions.write().await;
+
+            if let Some(session) = sessions.get_mut(df_session_id) {
+                let mut expiry = session.expiry.lock().await;
+                *expiry = record.expiry_date;
```
osipovartem, line 59: do you need to drop lock here?

crates/api-sessions/src/session.rs
```diff
@@ -51,51 +52,38 @@ impl RequestSessionStore {
...
-    async fn load(&self, id: &Id) -> session_store::Result<Option<Record>> {
-        Ok(self
-            .store
-            .lock()
-            .await
-            .get(id)
-            .filter(|Record { expiry_date, .. }| *expiry_date > OffsetDateTime::now_utc())
-            .cloned())
+    async fn load(&self, _id: &Id) -> session_store::Result<Option<Record>> {
+        Ok(None)
```
osipovartem, line 79: Maybe move locking logic back to this method?

crates/api-snowflake-rest/src/router.rs
```diff
@@ -1,11 +1,10 @@
-use crate::handlers::{abort, login, query};
+use crate::handlers::{abort, query};
 use crate::state::AppState;
 use axum::Router;
 use axum::routing::post;
 
 pub fn create_router() -> Router<AppState> {
     Router::new()
-        .route("/session/v1/login-request", post(login))
```
osipovartem, line 8: why do you need separate router for a single path? Can you register here everything for auth?

crates/core-executor/src/session.rs
```diff
@@ -35,6 +35,10 @@ use snafu::ResultExt;
 use std::collections::HashMap;
 use std::env;
 use std::sync::Arc;
+use time::{Duration, OffsetDateTime};
+use tokio::sync::Mutex;
+
+pub const SESSION_INACTIVITY_EXPIRATION_SECONDS: i64 = 5 * 60;
```
osipovartem, line 41: why only 5 mins?

osipovartem:
Run snowflake DBT for tests

crates/core-executor/src/session.rs
DanCodedThis: This was the value it had before.

crates/api-snowflake-rest/src/router.rs
DanCodedThis: There is only one path for snowflake rest auth. Before I actually had it wired directly in `main` but decided later on this approach.

DanCodedThis: Because the auth middleware, shouldn't affect this path

crates/api-sessions/src/session.rs
DanCodedThis: Load method, loads by Id of the `tower_sessions` which we no longer use, to not be affected by their cookie (on every route, including ones that don't use the cookie)

crates/api-sessions/src/session.rs
DanCodedThis: If you're talking about:
- Expiry lock drops automatically at the end of the **if** scope
- And the sessions either are dropped right after **if**, or we manually drop them to not create a deadlock in the **else** scope

crates/api-sessions/src/session.rs
DanCodedThis: This is the cleanup job, it was the value it had before.

DanCodedThis:
> This PR expected to be a breaking change since it uses a different cookie name. Session cookie name changed in this PR from "id" to "session_id" All the existing code supporting old sessions cookie will stop working. @DanCodedThis

@YaroslavLitvinov To my knowledge UI endpoints are either:
-  Used by the browser, which manges cookies automatically
- Our own unit tests, which I mentioned, but we decided that it's not a big deal since they don't fall with segfaults 
- Or need to be used by the `python slt runner` which id wasn't used there before (since they got increasing file descriptors and segfaults). Also, a little bit more info there: https://github.com/Embucket/embucket/issues/435#issuecomment-3043084216
- The Iceberg rest doesn't use the extractor or the sessions in any way
- The internal rest also doesn't use the extractor or sessions in any way

crates/api-snowflake-rest/src/router.rs
osipovartem: I mean create second router here and return merged router

DanCodedThis:
> Run snowflake DBT for tests

@osipovartem The only error I got is I think unrelated. Also, I changed the name of the schema to make it run correctly (the Iceberg public no schema bug you fixed, this branch is from the before):

```
(.venv) daniil@ThinkPad-E14-Gen-6:~/RustroverProjects/icebucket/dbt/snowplow$ dbt run -m snowplow_web
12:46:48  Running with dbt=1.9.6
12:46:49  Registered adapter: snowflake=1.9.4
12:46:49  Found 18 models, 103 data tests, 3 seeds, 2 operations, 8 sources, 771 macros
12:46:49  
12:46:49  Concurrency: 1 threads (target='dev')
12:46:49  
12:46:51  1 of 1 START hook: snowplow_web.on-run-start.0 ................................. [RUN]
12:46:51  1 of 1 OK hook: snowplow_web.on-run-start.0 .................................... [OK in 0.02s]
12:46:51  
12:46:51  1 of 18 START sql incremental model goodpublic_snowplow_manifest.snowplow_web_base_quarantined_sessions  [RUN]
12:46:51  1 of 18 OK created sql incremental model goodpublic_snowplow_manifest.snowplow_web_base_quarantined_sessions  [ok 1 in 0.39s]
12:46:51  2 of 18 START sql incremental model goodpublic_snowplow_manifest.snowplow_web_incremental_manifest  [RUN]
12:46:52  2 of 18 OK created sql incremental model goodpublic_snowplow_manifest.snowplow_web_incremental_manifest  [ok 1 in 0.38s]
12:46:52  3 of 18 START sql table model goodpublic_scratch.snowplow_web_base_new_event_limits  [RUN]
12:46:52  15:46:52 + Snowplow: New Snowplow incremental model. Backfilling
12:46:52  15:46:52 + Snowplow: Processing data between '2022-08-19 00:00:00' and '2022-09-18 00:00:00' (snowplow_web)
12:46:52  3 of 18 OK created sql table model goodpublic_scratch.snowplow_web_base_new_event_limits  [ok 1 in 0.60s]
12:46:52  4 of 18 START sql incremental model goodpublic_snowplow_manifest.snowplow_web_base_sessions_lifecycle_manifest  [RUN]
12:46:53  4 of 18 OK created sql incremental model goodpublic_snowplow_manifest.snowplow_web_base_sessions_lifecycle_manifest  [ok 1 in 0.69s]
12:46:53  5 of 18 START sql table model goodpublic_scratch.snowplow_web_base_sessions_this_run  [RUN]
12:46:54  5 of 18 OK created sql table model goodpublic_scratch.snowplow_web_base_sessions_this_run  [ok 1 in 0.50s]
12:46:54  6 of 18 START sql table model goodpublic_scratch.snowplow_web_base_events_this_run  [RUN]
12:46:54  6 of 18 OK created sql table model goodpublic_scratch.snowplow_web_base_events_this_run  [ok 1 in 0.50s]
12:46:54  7 of 18 START sql table model goodpublic_scratch.snowplow_web_pv_engaged_time .. [RUN]
12:46:54  7 of 18 OK created sql table model goodpublic_scratch.snowplow_web_pv_engaged_time  [ok 1 in 0.30s]
12:46:54  8 of 18 START sql table model goodpublic_scratch.snowplow_web_pv_scroll_depth .. [RUN]
12:46:55  8 of 18 OK created sql table model goodpublic_scratch.snowplow_web_pv_scroll_depth  [ok 1 in 0.30s]
12:46:55  9 of 18 START sql table model goodpublic_scratch.snowplow_web_sessions_this_run  [RUN]
12:46:55  9 of 18 ERROR creating sql table model goodpublic_scratch.snowplow_web_sessions_this_run  [ERROR in 0.21s]
12:46:55  10 of 18 START sql incremental model goodpublic_derived.snowplow_web_user_mapping  [RUN]
12:46:55  10 of 18 OK created sql incremental model goodpublic_derived.snowplow_web_user_mapping  [ok 1 in 0.38s]
12:46:55  11 of 18 START sql table model goodpublic_scratch.snowplow_web_page_views_this_run  [RUN]
12:46:56  11 of 18 OK created sql table model goodpublic_scratch.snowplow_web_page_views_this_run  [ok 1 in 0.70s]
12:46:56  12 of 18 SKIP relation goodpublic_derived.snowplow_web_sessions ................ [SKIP]
12:46:56  13 of 18 START sql incremental model goodpublic_derived.snowplow_web_page_views  [RUN]
12:46:57  13 of 18 OK created sql incremental model goodpublic_derived.snowplow_web_page_views  [ok 1 in 0.50s]
12:46:57  14 of 18 SKIP relation goodpublic_scratch.snowplow_web_users_sessions_this_run . [SKIP]
12:46:57  15 of 18 SKIP relation goodpublic_scratch.snowplow_web_users_aggs .............. [SKIP]
12:46:57  16 of 18 SKIP relation goodpublic_scratch.snowplow_web_users_lasts ............. [SKIP]
12:46:57  17 of 18 SKIP relation goodpublic_scratch.snowplow_web_users_this_run .......... [SKIP]
12:46:57  18 of 18 SKIP relation goodpublic_derived.snowplow_web_users ................... [SKIP]
12:46:57  
12:46:57  1 of 1 START hook: snowplow_web.on-run-end.0 ................................... [RUN]
12:46:57  1 of 1 OK hook: snowplow_web.on-run-end.0 ...................................... [OK in 0.19s]
12:46:57  
12:46:57  Finished running 7 incremental models, 2 project hooks, 11 table models in 0 hours 0 minutes and 7.74 seconds (7.74s).
12:46:57  
12:46:57  Completed with 1 error, 0 partial successes, and 0 warnings:
12:46:57  
12:46:57    Database Error in model snowplow_web_sessions_this_run (models/sessions/scratch/snowflake/snowplow_web_sessions_this_run.sql)
  000200: DataFusion error: Schema error: No field named derived_tstamp. Did you mean 'model_tstamp'?.
  compiled code at target/run/snowplow_web/models/sessions/scratch/snowflake/snowplow_web_sessions_this_run.sql
12:46:57  
12:46:57  Done. PASS=13 WARN=0 ERROR=1 SKIP=6 TOTAL=20
```

YaroslavLitvinov:
As @DanCodedThis  mentioned in issue #435 , api-ui used by SLT tests; I would add it also used by embucket-seed.
Breaking changes are acceptable in case if we are agree that it was a misuse of `api-ui`.
@andheroe

crates/api-snowflake-rest/src/router.rs
DanCodedThis: > I mean create second router here and return merged router

We can't do this, because we need the state for the auth layer fn. Only after we can merge the auth router path.

DanCodedThis: @osipovartem 

What I meant (in main with comments):

```rust
//The normal router
let snowflake_router = create_snowflake_router()
        .with_state(snowflake_state.clone())
        //This needs to be done before the merging of routers
        .layer(middleware::from_fn_with_state(
            snowflake_state.clone(),
            //The auth layer (`fn require_auth`)
            snowflake_require_auth,
        ));
    //That one path router
    let snowflake_auth_router = create_snowflake_auth_router().with_state(snowflake_state.clone());
    //The merge
    let snowflake_router = snowflake_router.merge(snowflake_auth_router);
```

osipovartem (approved):

PR #1221, opened 2025-06-26
conversion: Json format for `Time` type

### LINKED ISSUE #555: Fix `current_time()` and `current_date()` return type

DanCodedThis:
Closes #555

General info:
- ~Reimplemented `current_time()`~
- ~Added a test~
- Will have to extend it for when we wil be able to do `ALTER SESSION SET TIME_OUTPUT_FORMAT = 'HH24:MI:SS.FF';` to do `SELECT CURRENT_TIME(2);`
- We can support this right now, but with only limited: 0, 3,6 & 9 scale
- ~I added the signature for this, but right now it doesn't do anything~
- Also, we should support **_Timezones_** `ALTER SESSION SET TIMEZONE = 'UTC';`

Definition of done:
- [x] New `fn convert_time` with leading zeroes for **snow sql** `Time` Json format type
- [x] Unit tests for the new conversion
- ~[ ] SLT tests for the `DATA_FORMAT=JSON`~

~Docs: https://docs.snowflake.com/en/sql-reference/functions/current_time~


crates/embucket-functions/src/datetime/current_time.rs
```diff
@@ -0,0 +1,132 @@
...
+        &self.aliases
+    }
+
+    #[allow(clippy::unwrap_used)]
+    fn simplify(&self, _args: Vec<Expr>, info: &dyn SimplifyInfo) -> Result<ExprSimplifyResult> {
+        let now_ts = info.execution_props().query_execution_start_time;
+        let time = now_ts.timestamp_nanos_opt().map(|ts| {
+            i32::try_from((ts % 86_400_000_000_000) / 1_000_000_000)
+                .ok()
+                .unwrap()
```
osipovartem: lets remove unwrap

DanCodedThis: I went the normal `.timestamp()` - because it's seconds (and no `Option`), but then the UI has a problem. I got an error instead of a value in UI specifically. Something about cast to seconds. But the query itself was successful.

cc @nikitastryuk @YaroslavLitvinov 

It's either `ResultSet` or the UI UI part.

DanCodedThis: image

This happens if the logic is:

```rust
fn simplify(&self, _args: Vec<Expr>, info: &dyn SimplifyInfo) -> Result<ExprSimplifyResult> {
        let now_ts = info.execution_props().query_execution_start_time.timestamp();
        let time = 
            i32::try_from(now_ts)
                .ok();
        Ok(ExprSimplifyResult::Simplified(Expr::Literal(
            ScalarValue::Time32Second(time),
        )))
    }
```

Instead of:

```rust
 #[allow(clippy::unwrap_used)]
    fn simplify(&self, _args: Vec<Expr>, info: &dyn SimplifyInfo) -> Result<ExprSimplifyResult> {
        let now_ts = info.execution_props().query_execution_start_time;
        let time = now_ts.timestamp_nanos_opt().map(|ts| {
            i32::try_from((ts % 86_400_000_000_000) / 1_000_000_000)
                .ok()
                .unwrap()
        });
        Ok(ExprSimplifyResult::Simplified(Expr::Literal(
            ScalarValue::Time32Second(time),
        )))
    }
```

And this is the expected and working result right now:

image

DanCodedThis: We can simply remove the `.unwrap()`, but the underline issue of why I went this way, won't be solved

rampage644: what happens in `snow sql`? Also not a fan of unwrap - error should be returned instead

DanCodedThis: The problem was that it expected the time in one day (so `% 86_400`). Removed `.unwarp()`, we handle it like an option `.ok()` (to not introduce magical values or an error to function with no parameters (for now))

DanCodedThis:
Checked slt's, they don't use `current_time()` in any way, so seems like a false positive.

DanCodedThis:
**TIME** in snow sql:

```python
def _TIME_to_python(self, ctx: dict[str, Any]) -> Callable:
        """TIME to formatted string, SnowflakeDateTime, or datetime.time with no timezone attached."""
        scale = ctx["scale"]

        def conv0(value: str) -> time:
            return (
                datetime.fromtimestamp(float(value), timezone.utc)
                .replace(tzinfo=None)
                .time()
            )

        def conv(value: str) -> dt_t:
            microseconds = float(value[0 : -scale + 6])
            return (
                datetime.fromtimestamp(microseconds, timezone.utc)
                .replace(tzinfo=None)
                .time()
            )

        return conv if scale > 6 else conv0
```

**TIMESTAMP** in snow sql:

```python
def _TIMESTAMP_TZ_to_python(self, ctx: dict[str, Any]) -> Callable:
        """Converts TIMESTAMP TZ to datetime.

        The timezone offset is piggybacked.
        """
        scale = ctx["scale"]

        def conv(encoded_value: str) -> datetime:
            value, tz = encoded_value.split()
            tzinfo = _generate_tzinfo_from_tzoffset(int(tz) - 1440)
            return SnowflakeConverter.create_timestamp_from_string(
                value=value, scale=scale, tz=tzinfo
            )

        return conv

    def _get_session_tz(self) -> tzinfo | UTC:
        """Gets the session timezone or use the local computer's timezone."""
        try:
            tz = self.get_parameter("TIMEZONE")
            if not tz:
                tz = "UTC"
            return pytz.timezone(tz)
        except pytz.exceptions.UnknownTimeZoneError:
            logger.warning("converting to tzinfo failed")
            if tzlocal is not None:
                return tzlocal.get_localzone()
            else:
                return datetime.timezone.utc

    def _pre_TIMESTAMP_LTZ_to_python(
        self,
        value,
        ctx,
    ) -> tuple[datetime, int] | tuple[struct_time, int]:
        """Converts TIMESTAMP LTZ to datetime.

        This takes consideration of the session parameter TIMEZONE if available. If not, tzlocal is used.
        """
        microseconds, fraction_of_nanoseconds = _extract_timestamp(value, ctx)
        tzinfo_value = self._get_session_tz()

        try:
            t0 = ZERO_EPOCH + timedelta(seconds=microseconds)
            t = pytz.utc.localize(t0, is_dst=False).astimezone(tzinfo_value)
            return t, fraction_of_nanoseconds
        except OverflowError:
            logger.debug(
                "OverflowError in converting from epoch time to "
                "timestamp_ltz: %s(ms). Falling back to use struct_time."
            )
            return time.localtime(microseconds), fraction_of_nanoseconds

    def _TIMESTAMP_LTZ_to_python(self, ctx: dict[str, Any]) -> Callable:
        tzinfo = self._get_session_tz()
        scale = ctx["scale"]

        return partial(
            SnowflakeConverter.create_timestamp_from_string, scale=scale, tz=tzinfo
        )

    _TIMESTAMP_to_python = _TIMESTAMP_LTZ_to_python

    def _TIMESTAMP_NTZ_to_python(self, ctx: dict[str, Any]) -> Callable:
        """TIMESTAMP NTZ to datetime with no timezone info is attached."""
        scale = ctx["scale"]

        return partial(SnowflakeConverter.create_timestamp_from_string, scale=scale)

    def _TIMESTAMP_NTZ_numpy_to_python(self, ctx):
        """TIMESTAMP NTZ to datetime64 with no timezone info is attached."""

        def conv(value: str) -> numpy.datetime64:
            nanoseconds = int(decimal.Decimal(value).scaleb(9))
            return numpy.datetime64(nanoseconds, "ns")

        return conv
```

crates/core-executor/src/utils.rs
```diff
@@ -430,6 +443,80 @@ fn convert_uint_to_int_datatypes(
...
+    unit: TimeUnit,
+    data_format: DataSerializationFormat,
+) -> ArrayRef {
+    match data_format {
+        DataSerializationFormat::Json => {
+            let time: Vec<_> = match unit {
+                TimeUnit::Second => downcast_and_iter!(column, Time32SecondArray)
+                    .map(|time| {
+                        time.map(|ts| {
+                            let ts = DateTime::from_timestamp(i64::from(ts), 0).unwrap();
```
osipovartem: Let's remove unwap (return error or default)

rampage644:
Let's please add some test coverage for this, I presume several SLT tests would suffice?

DanCodedThis:
> Let's please add some test coverage for this, I presume several SLT tests would suffice?

Working on it, do SLTs run in Json data format? Artem suggested insta tests. Right now I'm working on adding unit tests.

crates/core-executor/src/utils.rs
```diff
@@ -198,6 +199,18 @@ pub fn convert_record_batches(
                     );
                     Arc::clone(&converted_column)
                 }
+                DataType::Time32(unit) | DataType::Time64(unit) => {
+                    let converted_column = convert_time(column, *unit, data_format);
```
osipovartem, line 203: column: &ArrayRef,
    unit: &TimeUnit,
    and pass unit as is

crates/core-executor/src/utils.rs
```diff
@@ -430,6 +439,81 @@ fn convert_uint_to_int_datatypes(
...
+                        })
+                    })
+                    .collect(),
+                TimeUnit::Millisecond => downcast_and_iter!(column, Time32MillisecondArray)
+                    .map(|time| {
+                        time.map(|ts| {
+                            let ts =
+                                DateTime::from_timestamp_millis(i64::from(ts)).unwrap_or_default();
+                            //If millis == 0, 4 zeroes after the `.` instead of 3
+                            format!(
```
osipovartem: move formatting to helper func because looks like duplicated code

osipovartem:
```rust
    TimeUnit::Microsecond => downcast_and_iter!(column, Time64MicrosecondArray)
        .map(|time| time.map(|ts| {
            let ts = DateTime::from_timestamp_micros(ts).unwrap_or_default();
            format_time_string(ts.timestamp(), ts.timestamp_subsec_micros(), 6)
        }))
        .collect(),
                    ....
/// Formats the timestamp and subsecond part into a string with the given scale.
/// `scale` is the number of digits to pad the subsecond value to.
fn format_time_string<T: std::fmt::Display>(
    timestamp: i64,
    subsecond: T,
    scale: usize,
) -> String {
    let sub_str = subsecond.to_string();
    let zeros = leading_zeros(scale, sub_str.len());
    format!("{}.{}{}", timestamp, zeros, sub_str)
}

File: crates/core-executor/src/utils.rs (hunk 3 above)
Comment (DanCodedThis) [reply crates/core-executor/src/utils.rs:203]: `clippy::trivially-copy-pass-by-ref`

File: crates/core-executor/src/utils.rs [hunk 5]
```diff
@@ -430,6 +439,74 @@ fn convert_uint_to_int_datatypes(
...
+                    })
+                    .collect(),
+            };
+            Arc::new(StringArray::from(time)) as ArrayRef
+        }
+        DataSerializationFormat::Arrow => column.clone(),
+    }
+}
+
+fn leading_zeros(target: usize, len: usize) -> String {
```
osipovartem: This func is used just once, lets maybe remove it and keep logic in format_time_string

osipovartem (approved):

osipovartem (approved):

Issue #1130, opened 2025-06-17
Finish integrating error_stack_trace::debug, so error stack trace printed in all crates


YaroslavLitvinov:
The list of crates where `error_stack_trace` should be sent to the logs is limited by our rest apis crates. Which are kind of "entry-points" for incoming requests.
These crates are : 
- [x] `api-snowflake-rest`
- [x] `api-ui`
- [x] `api-internal-rest`
- [x] `api-iceberg-rest`
- [ ] `embucketd` should it use errors instead of expects?

YaroslavLitvinov:
Reopen or create new issue when decided to replace expects by snafu errors in embucketd

PR #1226, opened 2025-06-27
issue # 1130 add error stack trace prints

YaroslavLitvinov:
Closes issue #1130

### DIFF
--- a/crates/api-ui/src/error.rs
+++ b/crates/api-ui/src/error.rs
@@ -98,7 +98,7 @@ impl IntoResponse for Error {
         tracing::error!("{}", self.output_msg());
         let code = self.status_code();
         let error = ErrorResponse {
-            message: self.output_msg(),
+            message: self.to_string(),
             status_code: code.as_u16(),
         };
         match self {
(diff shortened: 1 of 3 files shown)


crates/api-ui/src/error.rs
```diff
@@ -98,7 +98,7 @@ impl IntoResponse for Error {
         tracing::error!("{}", self.output_msg());
         let code = self.status_code();
         let error = ErrorResponse {
-            message: self.output_msg(),
+            message: self.to_string(),
```
DanCodedThis, line 101: @YaroslavLitvinov why the change here? Does `.to_string()` provide more/less info?

DanCodedThis (approved):
@YaroslavLitvinov LGTM

crates/api-ui/src/error.rs
YaroslavLitvinov: output_msg contains error stack trace, which is needed just for our log/tracing and shouldn't be returned in response

rampage644 (approved):

PR #1227, opened 2025-06-28
data types slt

andheroe:
@rampage644 is there anything here that I should remove?


test/sql/bronze_scope/data-types/file.slt
```diff
@@ -0,0 +1,121 @@
+# Test FILE data type for unstructured data
```
rampage644: Let's drop this 🙏

test/sql/bronze_scope/data-types/geospatial.slt
```diff
@@ -0,0 +1,354 @@
+# Test GEOGRAPHY and GEOMETRY geospatial data types
```
rampage644: I think we can postpone geospatial for now as well

rampage644:
All look good - are those 100% new tests or were supposed to replace some other?

rampage644 (approved):
lgtm, let's drop file and geospatial though

andheroe:
Dropped file and geospatial

These are tests written from scratch, but they are not 100% addition - @YevheniiNiestierov is doing setup queries cleanup and we had multiple queries failing because of unsupported column type creation. These tests cover that. But they also cover more functionality of types.

YevheniiNiestierov (approved):

PR #1236, opened 2025-07-01
Support "MERGE INTO" statement

JanKaul:
This PR implements the basic functionality for the MERGE INTO statement.
- Adds LogicalPlan for Merge
- Adds ExecutionPlan for Merge
- Adopts `merge_query` function to create LogicalPlan

Is based on the design doc: https://docs.google.com/document/d/1bVWg_mV7Ipte3IhdEQMffHaYz5sVjzwWHV9JsMrXgwE/edit?tab=t.0#heading=h.1z4gmr8tge25


crates/core-executor/src/query.rs
```diff
@@ -919,86 +936,165 @@ impl UserQuery {
...
-                            .join(", ");
-                    }
-                }
-            }
-        }
-        let select_query = format!(
-            "SELECT {values} FROM {source_query} LEFT JOIN {target_table} {target_alias} ON {on}{where_clause_str}"
-        );
+        let schema = build_join_schema(&target_schema, &source_schema, &JoinType::Full)
+            .context(ex_error::DataFusionSnafu)?;
```
rampage644: Ideally, we want different context selectors for different failures. In this specific example we might create a `ex_error::MergeBuildJoinSchemaSnafu` with `error: DatafusionError` as one of the field. cc @YaroslavLitvinov to validate this.

As far as I understand this will allow to differentiate between different errors in the same function.


```suggestion
            .context(ex_error::MergeBuildJoinSchemaSnafu)?;
```

rampage644: Maybe not one for each `?` but still not a generic `DataFusionSnafu` that I believe is the a wrong approach we still haven't totally fixed.

rampage644 (approved):

rampage644:
Couldn't review the PR in full, skimmed mostly over parts, would be great to have this properly tested as well. SLT fails with 3/8 success rate also.

osipovartem (approved):

JanKaul:
@rampage644 Should I include the databend test file: 09_0036_merge_into_without_distributed_enable?

Otherwise, I would go ahead and merge the PR.

crates/core-executor/src/query.rs
JanKaul: The error variants should be much better now.

rampage644:
@JanKaul 👍 Up to you

rampage644 (approved):

YaroslavLitvinov (approved):
re-approving

PR #1249, opened 2025-07-03
fix: remove hardcoded /catalog suffix from iceberg catalog URL

rampage644:
## Summary
- Remove hardcoded `/catalog` suffix from iceberg catalog URL in handlers.rs line 297
- This change removes the hardcoded path when returning the catalog configuration in `/v1/config` endpoint

## Test plan
- [ ] Verify that the `/v1/config` endpoint returns the correct catalog URL without the hardcoded suffix
- [ ] Test that iceberg catalog functionality still works correctly

### DIFF
--- a/crates/api-iceberg-rest/src/handlers.rs
+++ b/crates/api-iceberg-rest/src/handlers.rs
@@ -294,7 +294,7 @@ pub async fn get_config(
     let config = CatalogConfig {
         defaults: HashMap::new(),
         overrides: HashMap::from([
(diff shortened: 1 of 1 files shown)


osipovartem (approved):

osipovartem:
https://github.com/Embucket/embucket/blob/main/crates/embucketd/src/main.rs#L202
we register all iceberg catalog routes with catalog prefix
that's why we had **/catalog** in config

PR #1269, opened 2025-07-07
Refactor errors in crates: df-catalog, embucket-functions

YaroslavLitvinov:
Closes #1132

- Some refactoring of df-catalog errors
- Refactoring of embucket-functions error - split huge errors file into a separate files
- It also fixes issue with making errors External (because of `PanicWrapper` which was formerly used)

PR #1267, opened 2025-07-07
Old unsplitted PR: Create approach for returning Snowflake like errors

YaroslavLitvinov:
Closes #1126 

### General guidance for making Snowflake errors

1. SnowflakeError is the part of `core-executor::Error` implementing `From<Error> for SnowflakeError`, it also include tests for returning Snowflake errors.
2. Converter gathers all the possible matches needed to identify any Embucket error
3. error_stack_trace strings from `log` and from `slt_errors_stats_embucket.csv` produced by slt_runner, is used as a source of structured data when matching errors
4. Use DataFusionError::External to encode any specific error context, as it can carry any boxed error
5. Use DataFusionError::Collection for returning array of errors (or use own error for it, wrapped as External error)
6. Add tests for added Snowflake errors

### Changes it contains (This is not an atomic PR)
- Some refactoring of df-catalog errors
- Refactoring of embucket-functions error - split huge errors file into a separate files
- Added `/v1/metastore/queries/{query_id}` handler returning QueryResult from History by query_id
- Added `slt_errors_stats_{embucket|snowflake}.csv` report produced by SLT Runner
- SnowflakeError + some tests added into `core-executor::Error`
- Returning Snowflake error from both api-ui, api-snowflake-rest

PR #1271, opened 2025-07-07
Create approach for returning Snowflake like errors [splitted]

YaroslavLitvinov:
Closes #1126 
This PR replaces following PR https://github.com/Embucket/embucket/pull/1267 which will be closed.

Since this PR was a part of bigger one ^^, so now it depends on PR https://github.com/Embucket/embucket/pull/1269 to be merged before it.

### General guidance for making Snowflake errors

1. SnowflakeError is the part of `core-executor::Error` implementing `From<Error> for SnowflakeError`, it also include tests for returning Snowflake errors.
2. Converter gathers all the possible matches needed to identify any Embucket error
3. error_stack_trace strings from `log` and from `slt_errors_stats_embucket.csv` produced by slt_runner, is used as a source of structured data when matching errors
4. Use DataFusionError::External to encode any specific error context, as it can carry any boxed error
5. Use DataFusionError::Collection for returning array of errors (or use own error for it, wrapped as External error)
6. Use DataFusionError::Diagnostic for returning parser context
7. Add tests for added Snowflake errors

### Changes it contains
- Added `/v1/metastore/queries/{query_id}` handler returning QueryResult from History by query_id
- Added `slt_errors_stats_{embucket|snowflake}.csv` report produced by SLT Runner
- SnowflakeError + some tests added into `core-executor::Error`
- Returning Snowflake error from both api-ui, api-snowflake-rest


YaroslavLitvinov:
I also tagged here @YevheniiNiestierov as PR changes SLT runner as well

YaroslavLitvinov:
Marked this as WIP until prerequisite PR https://github.com/Embucket/embucket/pull/1269 is not merged

crates/api-snowflake-rest/src/error.rs
```diff
@@ -106,8 +107,8 @@ impl IntoResponse for Error {
         skip(self)
     )]
     fn into_response(self) -> axum::response::Response<axum::body::Body> {
-        tracing::error!("{}", self.output_msg());
-        let (status_code, message) = if let Self::Execution { source } = &self {
+        tracing::error!(" {}", self.output_msg());
```
rampage644: Why extra whitespace in the message?

YaroslavLitvinov: accidently added

osipovartem:
DF started to use new type of error
```rust
let e = e.with_diagnostic(Diagnostic::new_error(
    format!("table '{}' not found", table_ref),
    Span::try_from_sqlparser_span(relation_span),
));
```

```rust
    /// Error wrapped together with additional contextual information intended
    /// for end users, to help them understand what went wrong by providing
    /// human-readable messages, and locations in the source query that relate
    /// to the error in some way.
    Diagnostic(Box<Diagnostic>, Box<DataFusionError>),
```
Maybe we can use the same approach

rampage644:
> DF started to use new type of error

This is really great, it's stacking the context for the end user. It still relies on `String` as the only source of context, though. Doesn't solve our problem :)

crates/api-snowflake-rest/src/error.rs
```diff
@@ -108,12 +109,13 @@ impl IntoResponse for Error {
...
         level = "info",
-        fields(status_code),
+        fields(error, error_stack_trace, status_code),
         skip(self)
     )]
     fn into_response(self) -> axum::response::Response<axum::body::Body> {
-        tracing::error!("{}", self.output_msg());
-        let (status_code, message) = if let Self::Execution { source } = &self {
+        // Record the result as part of the current span.
+        tracing::Span::current().record("error_stack_trace", self.output_msg());
```
rampage644, line 117: this is interesting, do we still get an logging entry?

crates/core-executor/src/tests/snowflake_errors.rs
```diff
@@ -0,0 +1,84 @@
...
+            error: arrow_schema::ArrowError::DivideByZero,
+            location: location!(),
+        },
+    }));
+    assert_eq!(
+        err.to_string(),
+        "External error: Failed to cast to int: Divide by zero error"
+    );
+
+    // assert!(false);
```
rampage644: remove?

crates/core-executor/src/snowflake_error.rs
```diff
@@ -0,0 +1,277 @@
...
+fn diagnostic_location_error(diagnostic: &Diagnostic) -> Option<String> {
+    diagnostic.span.map(|span| {
+        format!(
+            "error line {} at position {}\n",
+            span.start.line, span.start.column
+        )
+    })
+}
+
+fn datafusion_error(datafusion_error: DataFusionError) -> SnowflakeError {
```
rampage644, line 223: would it be possible to convert into `Into`/`From` trait?

crates/core-executor/src/snowflake_error.rs
```diff
@@ -0,0 +1,277 @@
...
+                // SQL compilation error: Object 'DATABASE.PUBLIC.ARRAY_DATA' does not exist or not authorized.
+                // TODO: add line and column from diagnostic.span.
+                message: format!(
+                    "SQL compilation error: {location_error}{}",
+                    diagnostic.message
+                ),
+            }
+        }
+        DataFusionError::Execution(err) => SnowflakeError {
+            message: format!("SQL compilation error: {err}"),
```
rampage644, line 250: Is `SQL compilation` correct message here?

rampage644 (approved):

crates/api-snowflake-rest/src/error.rs
YaroslavLitvinov: yes, it should be in our logs

YaroslavLitvinov:
> > DF started to use new type of error
> 
> This is really great, it's stacking the context for the end user. It still relies on `String` as the only source of context, though. Doesn't solve our problem :)

It also has a span -context from parser

crates/core-executor/src/snowflake_error.rs
YaroslavLitvinov: I think it's correct. As "SQL compilation error: " is a generic message for many Snowflake errors.

rampage644 (approved):

crates/core-executor/src/snowflake_error.rs
YaroslavLitvinov: No, I've created a separate function as DataFusionError should be resolved for Boxed and unboxed arms that physically in different closures.

PR #1304, opened 2025-07-09
Support drop database

### LINKED ISSUE #159: engine: Basic support for `drop database`
At the moment `DROP DATABASE demo` returns `200: DataFusion error: This feature is not implemented: Only `DROP TABLE/VIEW/SCHEMA  ...` statement is supported currently`

```bash
snow sql -c local                                                                                                      ✔  24m 27s   13:37:31  
╭─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ Welcome to Snowflake-CLI REPL                                                                                                                               │
│ Type 'exit' or 'quit' to leave                                                                                                                              │
 > drop database demo;

Error occurred: 000200: 200: DataFusion error: This feature is not implemented: Only `DROP TABLE/VIEW/SCHEMA  ...` statement is supported currently
 >
```

osipovartem:
* This pull request introduces support for dropping catalogs, including databases, in the core executor and catalog management components. It also removes functionality related to external catalog registration (duplicated call) and simplifies catalog handling by introducing a new `CatalogType` enum.
* Added test cases for `DROP DATABASE` functionality in `crates/core-executor/src/tests/query.rs` and corresponding snapshot validation.
* Extended `SnowflakeError` to include mappings for the new error variants `InvalidDatabaseIdentifier` and `DropCatalog`.
* Added a new `NotImplemented` error variant to `crates/df-catalog/src/error.rs` for unsupported catalog operations, such as dropping S3 tables catalogs.
* For Embucket catalog we return an error if database in use by schemas

Added tests with
```sql
DROP DATABASE test
```
Closes #159

### DIFF
--- a/crates/core-executor/src/session.rs
+++ b/crates/core-executor/src/session.rs
@@ -11,27 +11,18 @@ use crate::datafusion::query_planner::CustomQueryPlanner;
 use crate::models::QueryContext;
 use crate::query::UserQuery;
 use crate::utils::Config;
-use aws_config::{BehaviorVersion, Region, SdkConfig};
-use aws_credential_types::Credentials;
-use aws_credential_types::provider::SharedCredentialsProvider;
 use core_history::history_store::HistoryStore;
-use core_metastore::error::{self as metastore_error};
-use core_metastore::{AwsCredentials, Metastore, VolumeType as MetastoreVolumeType};
-use core_utils::scan_iterator::ScanIterator;
-use datafusion::catalog::CatalogProvider;
+use core_metastore::Metastore;
 use datafusion::execution::runtime_env::RuntimeEnvBuilder;
 use datafusion::execution::{SessionStateBuilder, SessionStateDefaults};
 use datafusion::prelude::{SessionConfig, SessionContext};
 use datafusion::sql::planner::IdentNormalizer;
 use datafusion_functions_json::register_all as register_json_udfs;
-use datafusion_iceberg::catalog::catalog::IcebergCatalog as DataFusionIcebergCatalog;
 use df_catalog::catalog_list::{DEFAULT_CATALOG, EmbucketCatalogList};
 use df_catalog::information_schema::session_params::{SessionParams, SessionProperty};
 use embucket_functions::expr_planner::CustomExprPlanner;
 use embucket_functions::register_udafs;
 use embucket_functions::table::register_udtfs;
-use iceberg_rust::object_store::ObjectStoreBuilder;
-use iceberg_s3tables_catalog::S3TablesCatalog;
 use snafu::ResultExt;
 use std::collections::HashMap;
 use std::env;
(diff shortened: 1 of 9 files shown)


crates/core-executor/src/session.rs
```diff
@@ -121,65 +112,9 @@ impl UserSession {
...
-            let catalog = DataFusionIcebergCatalog::new(Arc::new(catalog), None)
-                .await
-                .context(ex_error::DataFusionSnafu)?;
-            let catalog_provider = Arc::new(catalog) as Arc<dyn CatalogProvider>;
-
-            self.ctx.register_catalog(volume.name, catalog_provider);
-        }
-        Ok(())
-    }
-
```
DanCodedThis: Where was the duplicated call?

DanCodedThis (approved):
@osipovartem LGTM

crates/core-executor/src/session.rs
osipovartem: During session creation we already call catalogs initialization

crates/df-catalog/src/catalog_list.rs
```diff
@@ -57,6 +59,37 @@ impl EmbucketCatalogList {
...
+                entity: "catalog",
+                name,
+            }
+            .fail();
+        };
+        match catalog.catalog_type {
+            CatalogType::Internal => {
+                // Set cascade to true to delete all tables in the database
+                self.metastore
+                    .delete_database(&name.to_string(), true)
```
rampage644, line 80: Should it be `delete_database(name, cascade)?`

crates/df-catalog/src/catalog.rs
```diff
@@ -6,12 +6,20 @@ use std::{any::Any, sync::Arc};
...
     pub should_refresh: bool,
     pub name: String,
     pub enable_information_schema: bool,
 }
 
+#[derive(Clone, Debug)]
+pub enum CatalogType {
+    Internal,
+    Memory,
+    S3tables,
```
rampage644, line 20: `External`?

crates/df-catalog/src/catalog_list.rs
```diff
@@ -57,6 +59,37 @@ impl EmbucketCatalogList {
...
+                    .await
+                    .context(MetastoreSnafu)?;
+                Ok(())
+            }
+            CatalogType::Memory => Ok(()),
+            CatalogType::S3tables => NotImplementedSnafu {
```
rampage644, line 86: Does it mean self.catalogs doesn't contain the catalog?

crates/df-catalog/src/catalog.rs
osipovartem: External means S3Tables.
We check the type of catalog to call specific logic

crates/df-catalog/src/catalog_list.rs
osipovartem: It means that we don't support removing database (s3-table bucket). We can just add logic to remove it from catalog list

crates/df-catalog/src/catalog_list.rs
osipovartem: In case of cascade false we just remove record from slate db. It means if the user recreate database he will see all previous schemas and tables. Is it ok or we should delete data completely?

rampage644: I think we should error out if database contains tables and schemas, what is snowflake behavior here?

osipovartem: https://docs.snowflake.com/en/sql-reference/sql/drop-database#usage-notes

Dropping a database does not permanently remove it from the system. A version of the dropped database is retained in [Time Travel](https://docs.snowflake.com/en/user-guide/data-time-travel) for the number of days specified by the DATA_RETENTION_TIME_IN_DAYS parameter for the database:

Within the Time Travel retention period, a dropped database can be restored using the [UNDROP DATABASE](https://docs.snowflake.com/en/sql-reference/sql/undrop-database) command.

When the Time Travel retention period ends, the next state for the dropped database depends on whether it is permanent or transient:

A permanent database moves into [Fail-safe](https://docs.snowflake.com/en/user-guide/data-failsafe). In Fail-safe (7 days), a dropped database can be recovered, but only by Snowflake. When the database leaves Fail-safe, it is purged.

A transient database has no Fail-safe, so it is purged when it moves out of Time Travel.

Once a dropped database has been purged, it cannot be recovered; it must be recreated.

crates/df-catalog/src/error.rs
```diff
@@ -79,4 +79,10 @@ pub enum Error {
         #[snafu(implicit)]
         location: Location,
     },
+    #[snafu(display("Feature not implemented: '{details:?}'"))]
```
YaroslavLitvinov, line 82: From perspective of extended error handling this error gives no enough context that can be matched on what exactly unsupported. Maybe it would be better to have additional `enum UnsupportedFeature { S3TableDrop }`. 
And then:
```
    NotImplemented {
        feature: UnsupportedFeature,
        details: String,
        #[snafu(implicit)]
        location: Location,
    },
```

crates/df-catalog/src/catalog_list.rs
osipovartem: Added a logic to return error if database is not empty without cascade enabled

DanCodedThis (approved):
@osipovartem reapprove just in case

rampage644 (approved):

DanCodedThis (approved):
@osipovartem reapproved

PR #1335, opened 2025-07-15
Create database with external volume

### LINKED ISSUE #510: `Discussion` CREATE DATABASE
This issue was added in order to move related discussions here. Please feel free to update and comment it!

# What our requirements?
As it stated in reference implementation, and also mentioned in issue #470 
> Creating a new database automatically creates two schemas in the database:
    PUBLIC: Default schema for the database.
    INFORMATION_SCHEMA: Schema which contains views and table functions that can be used for querying metadata about the objects in the database, as well as across all objects in the account.

# Which way do we choose for implementing it?

Currently we don't support CREATE DATABASE statements. At least reasons are:
* Not enough context as is it depends on underlying volume
* No dependency on volumes in a genuine datafusion's codebase.

Instead we create database from UI creating record directly in metastore. It creates no default schemas:
* No PUBLIC schema created
* No information_schema created

osipovartem:
- added support for `CREATE DATABASE db_test external_volume = 'test_volume'`
- tested with all types of volumes, for s3 tables return unimplemeted error
- Fixed `DROP DATABASE` comments from (**commit was removed during rebase**) https://github.com/Embucket/embucket/pull/1304
  - Fixed catalog type and metastore errors for database in use, added tests
  - added specific unimplemented enum which was mentioned in comments
- Added missing filters for `SHOW DATABASES` to support `SHOW DATABASES STARTS WITH 'db_test'`

Closes #510 
Based on #https://github.com/Embucket/datafusion-sqlparser-rs/pull/14
**This PR allows to move to `CREATE DATABASE` sql instead of internal API call. After migration it is possible to remove [job ](https://github.com/Embucket/embucket/blob/main/crates/df-catalog/src/catalog_list.rs#L397-L440)which checks created catalogs every 10 seconds**.


crates/core-executor/src/query.rs
```diff
@@ -1203,6 +1189,33 @@ impl UserQuery {
...
+            return ex_error::ExternalVolumeRequiredForCreateDatabaseSnafu.fail();
+        }
+        let catalog_name = object_name_to_string(&db_name);
+        let catalog_exist = self.get_catalog(&catalog_name).is_ok();
+        if catalog_exist {
+            if if_not_exists {
+                return self.created_entity_response();
+            }
+            return ex_error::ObjectAlreadyExistsSnafu {
+                type_name: "database".to_string(),
```
rampage644: move to enum?

crates/core-executor/src/query.rs
```diff
@@ -1203,6 +1189,33 @@ impl UserQuery {
...
 
+    #[instrument(name = "UserQuery::create_database", level = "trace", skip(self), err)]
+    pub async fn create_database(
+        &self,
+        db_name: ObjectName,
+        if_not_exists: bool,
+        external_volume: Option<String>,
+    ) -> Result<QueryResult> {
+        if external_volume.is_none() {
+            return ex_error::ExternalVolumeRequiredForCreateDatabaseSnafu.fail();
```
rampage644: capture db name as well

crates/core-metastore/src/error.rs
```diff
@@ -182,6 +182,13 @@ pub enum Error {
         location: Location,
     },
 
+    #[snafu(display("Database in use by schema(s): {schema}"))]
+    DatabaseInUse {
```
rampage644, line 186: Capture database name. I also wonder if schema is strictly required here: going forward it could be an expensive operation to fetch all schemas in the database.

crates/df-catalog/src/catalog_list.rs
```diff
@@ -74,22 +88,77 @@ impl EmbucketCatalogList {
...
+            VolumeType::S3(_) | VolumeType::File(_) => {
+                let database = Database {
+                    ident: catalog_name.to_owned(),
+                    volume: volume_ident.to_owned(),
+                    properties: None,
+                };
+                self.metastore
+                    .create_database(&catalog_name.to_owned(), database)
+                    .await
+                    .context(MetastoreSnafu)?;
```
rampage644, line 137: Let's make an error more explicit here: `MetastoreSnafu` --> something else. @YaroslavLitvinov do you have an idea? I was trying to think about better name. As far as I understand snafu philosophy the error here should provide the context on top of underlying Metastore error, such that we know we tried to create a catalog in our `CatalogList` implementation, had S3 based volume and attempted to create a database entity.

rampage644: @osipovartem Will it be possible to check what error we get (in the logs + in the response) when hit this error code path?

rampage644 (commented):
Overall lgtm, could we experiment with better error handling?

crates/core-metastore/src/error.rs
osipovartem: We already fetched them before returning the error

crates/df-catalog/src/catalog_list.rs
osipovartem: **create_database file/memory based database errors**
- Object of type database with name local already exists
- Failed to create database: Missing volume with ident: '"testvol"'

**create database for s3_tables** 
- Failed to create database: Feature not implemented: CreateS3TablesDatabase, "Creating an S3 table catalog is not supported

rampage644 (approved):

Vedin (approved):

PR #1338, opened 2025-07-15
Many auth DedicatedExecutors fix

### LINKED ISSUE #1290: Embucket opens to many sessions
Root cause: Embucket consumes all the available file descriptors when creating new sessions, and finally fail creating new sessions when no more files can't be opened. It depends on environment, how many FD can be opened, so probably you need run repro more times or change it increasing numbers. 

After 3 minutes or so Embucket recovers and works as usual!

### Snippet from embucketd log after it failed to open new session
```
---- tests::test_seed_client stdout ----

thread 'DedicatedExecutor driver' panicked at crates/core-executor/src/dedicated_executor.rs:600:22:
Creating tokio runtime: Os { code: 24, kind: Uncategorized, message: "Too many open files" }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

thread 'tests::test_seed_client' panicked at crates/core-executor/src/dedicated_executor.rs:627:39:
driver started: RecvError

thread 'tests::test_seed_client' panicked at crates/embucket-seed/src/requests/http.rs:88:24:
called `Result::unwrap()` on an `Err` value: reqwest::Error { kind: Request, url: "http://0.0.0.0:40053/ui/queries", source: hyper_util::client::legacy::Error(SendRequest, hyper::Error(IncompleteMessage)) }


failures:
    tests::test_seed_client
```

### Ways to reproduce
``` bash
# using snowsql
for i in {1..100}; do (snow sql -c local-embucket-bad-pass -q "SELECT 1" &) ; done

#   using api-ui
for i in {1..100}; do (python -c "import requests; requests.request('post', 'http://localhost:3000/ui/auth/login')" &) ; done  
```

Update:
Previous approach `cargo embucket-seed` can't be used as a repro anymore, after merging PR https://github.com/Embucket/embucket/pull/1329 that supports new session_id cookie.

DanCodedThis:
Closes #1290 

- Removed Extractor from `login` in `snowflake-rest` and `api-ui` auths to not call the session generation before log in.
- Also removed Extractor from `refresh_token` in `api-ui`.
- Removed `tower_session` layer since it was used just as a gimmick to get to execution service through a trait. 
- Instead, added an `ExecutionAppState` trait for getting `execution_svc` from any AppState that has it.
- As suggested by @YaroslavLitvinov removed session management from `api-ui` auth flow entirely. This resulted in a somewhat big rewrite (refactor). 
- However, this now means that session flow is completely self-contained, including, but not limited to: session generation, session cookie management, session extraction, etc.

### DIFF
--- a/crates/api-sessions/src/session.rs
+++ b/crates/api-sessions/src/session.rs
@@ -108,6 +108,9 @@ impl DFSessionId {
     }
 }
 
+//Snowflake token extraction lives in the api-session crate (used here also),
+// so to not create a cyclic dependency exporting it from the api-snowflake-rest crate,
+// where it's used in the `require_auth` layer as part of the session flow and was originally from
 #[must_use]
 pub fn extract_token_from_auth(headers: &HeaderMap) -> Option<String> {
     //First we check the header
--- a/crates/api-sessions/src/session.rs
+++ b/crates/api-sessions/src/session.rs
@@ -1,128 +1,54 @@
-use axum::{Json, extract::FromRequestParts, response::IntoResponse};
+use crate::error as session_error;
+use axum::extract::FromRequestParts;
+use core_executor::ExecutionAppState;
 use core_executor::service::ExecutionService;
 use core_executor::session::SESSION_INACTIVITY_EXPIRATION_SECONDS;
 use http::header::COOKIE;
 use http::request::Parts;
 use http::{HeaderMap, HeaderName};
 use regex::Regex;
-use serde::{Deserialize, Serialize};
 use snafu::ResultExt;
-use snafu::prelude::*;
 use std::{collections::HashMap, sync::Arc};
 use time::{Duration, OffsetDateTime};
-use tower_sessions::{
-    ExpiredDeletion, Expiry, Session, SessionStore,
-    session::{Id, Record},
-    session_store,
-};
-use uuid;
 
 pub const SESSION_ID_COOKIE_NAME: &str = "session_id";
 
 pub const SESSION_EXPIRATION_SECONDS: u64 = 60;
 
 #[derive(Clone)]
-pub struct RequestSessionStore {
-    execution_svc: Arc<dyn ExecutionService>,
+pub struct SessionStore {
+    pub execution_svc: Arc<dyn ExecutionService>,
 }
 
-#[allow(clippy::missing_const_for_fn)]
-impl RequestSessionStore {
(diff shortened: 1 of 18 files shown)


crates/api-sessions/src/session.rs
```diff
@@ -131,82 +57,78 @@ pub struct DFSessionId(pub String);
...
-                .context(SessionPersistSnafu)?;
-            session.set_expiry(Some(Expiry::OnInactivity(Duration::seconds(
-                SESSION_INACTIVITY_EXPIRATION_SECONDS,
-            ))));
-            session.save().await.context(SessionPersistSnafu)?;
-            Ok(Self(token))
-        //If the session is dead
+        let execution_svc = state.get_execution_svc();
+
+        if let Some(token) = extract_token_from_auth(&req.headers) {
```
YaroslavLitvinov, line 68: What is extract_token_from_auth extracting SnowflakeToken (api-snowflake-rest) doing in DFSessionId extractor?

If that wasn't a mistake, such unobvious things worth a comment.

crates/api-ui/src/auth/error.rs
```diff
@@ -95,6 +95,12 @@ pub enum Error {
         #[snafu(implicit)]
         location: Location,
     },
+
+    #[snafu(display("Execution error: {error}"))]
+    Execution {
```
YaroslavLitvinov: auth Execution error accidently added?

YaroslavLitvinov (approved):
LGTM, There are just two things I mentioned that may need fixes.

crates/api-ui/src/auth/error.rs
DanCodedThis: Yep, it was from the first iteration I tried.

crates/api-sessions/src/session.rs
DanCodedThis: I'll add a comment. Some explanation: this function is used in the **session.rs** itself. And it's used in the `require_auth` layer as part of the session flow in the **api-snowflake-rest**, but exporting the function from the **api-snowflake-rest** would create a cyclic dependency.

YaroslavLitvinov (approved):

YaroslavLitvinov (approved):

DanCodedThis:
Unit-test flow can't run. I restarted, still it's stuck. No significant changes were done (last time it ran successfully), so I'm merging.

Issue #1148, opened 2025-06-19
[BUG] Function try_to_numeric is not implemented yet


osipovartem:
SQL to test
```sql
SELECT 
    TRY_TO_NUMERIC('1')::VARCHAR AS gl_namespace_id
```

`Query execution error: Query execution error: DataFusion error: This feature is not implemented: Function 'try_to_numeric' is not implemented yet. Details: https://docs.snowflake.com/en/sql-reference/functions/try_to_decimal 0: <transparent> 1: Query execution error: Query execution error: DataFusion error: This feature is not implemented: Function 'try_to_numeric' is not implemented yet. Details: https://docs.snowflake.com/en/sql-reference/functions/try_to_decimal, at crates/api-ui/src/queries/handlers.rs:118:10 2: Query execution error: DataFusion error: This feature is not implemented: Function 'try_to_numeric' is not implemented yet. Details: https://docs.snowflake.com/en/sql-reference/functions/try_to_decimal, at crates/api-ui/src/queries/handlers.rs:117:10 3: DataFusion error: This feature is not implemented: Function 'try_to_numeric' is not implemented yet. Details: https://docs.snowflake.com/en/sql-reference/functions/try_to_decimal, at crates/core-executor/src/query.rs:203:44 4: NotImplemented("Function 'try_to_numeric' is not implemented yet. Details: https://docs.snowflake.com/en/sql-reference/functions/try_to_decimal")`

Issue #604, opened 2025-05-01
[BUG] Snowflake Function: to_decimal


rampage644:
**Function:** `to_decimal`
**Category:** Conversion
**Status:** Failed (Return Code: 1)
**Documentation:** https://docs.snowflake.com/en/sql-reference/functions/to_decimal

**Test SQL:**
```sql
SELECT TO_DECIMAL('98.6', 5, 1);
```

**Expected Result:**
```
98.6
```

**Actual Result:**
```
SELECT TO_DECIMAL('98.6', 5, 1);
```

**Stderr:**
```
╭─ Error ─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ 000200: 200: DataFusion error: Error during planning: Invalid function 'to_decimal'.                                                                        │
│ Did you mean 'to_char'?                                                                                                                                     │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

**Parent EPIC:** #565

PR #1339, opened 2025-07-15
`to_decimal` & `try_to_decimal` udf

### LINKED ISSUE #601: [BUG] Snowflake Function: try_to_decimal
**Function:** `try_to_decimal`
**Category:** Conversion
**Documentation:** https://docs.snowflake.com/en/sql-reference/functions/try_to_decimal

**Test SQL:**
```sql
SELECT TRY_TO_DECIMAL('123.45', 5, 2), TRY_TO_DECIMAL('xyz', 5, 2);
```

**Expected Result:**
```
Row with 123.45, NULL
```

**Actual Result:**
```
SELECT TRY_TO_DECIMAL('123.45', 5, 2), TRY_TO_DECIMAL('xyz', 5, 2);
```

**Stderr:**
```
│ 000200: 200: DataFusion error: Error during planning: Invalid function 'try_to_decimal'.                                                                    │
│ Did you mean 'array_dims'?                                                                                                                                  │
```

**Parent EPIC:** #565

DanCodedThis:
Closes #601, #1148 & #604

- `to_decimal` impl completely made on top of arrow cast
- Added more errors
- Added `try_` support
- Added basic formatting support: comma, whitespace & $ (dollar sign)
- Added insta tests
- Fixed (tried fixing) slt test
- This is the basic impl, to be continued on: 
  - #1340
  - #1341
  - #1342

Reference: https://docs.snowflake.com/en/sql-reference/functions/to_decimal


crates/embucket-functions/src/conversion/errors.rs
```diff
@@ -44,6 +54,58 @@ pub enum Error {
...
+    #[snafu(display("Invalid scale only allowed from 1 to {precision_minus_one} got: {scale:?}"))]
+    InvalidScale {
+        precision_minus_one: u8,
+        scale: ScalarValue,
+        #[snafu(implicit)]
+        location: Location,
+    },
+
+    #[snafu(display("Too little arguments, got: {got}, expected at least: {at_least}"))]
+    TooLittleArguments {
```
osipovartem: Maybe better to have single validation error
with just details, like not enough arguments, too many arguments, etc

DanCodedThis: No params?

crates/embucket-functions/src/conversion/to_decimal.rs
```diff
@@ -0,0 +1,342 @@
...
+                got: other,
+                at_maximum: 4usize,
+            }
+            .fail()?,
+        }
+    }
+    //TODO: formatting <format> second argument
+    #[allow(clippy::unwrap_used, clippy::too_many_lines)]
+    fn invoke_with_args(&self, args: ScalarFunctionArgs) -> DFResult<ColumnarValue> {
+        let DataType::Decimal128(precision, scale) = args.return_type else {
```
osipovartem, line 288: why we check return type here?

DanCodedThis: To get precision and scale the easy way. Since we already have figured them out and put in the return type.

DanCodedThis:
For anybody reviewing. Forgot to add the Boolean support (probably works out of the box).

DanCodedThis:
Why did insta test fail with new snapshots? I've committed the new ones.

crates/embucket-functions/src/conversion/errors.rs
DanCodedThis: I kinda went the "not enough context" route. I'm sure later will change it to adhere to snowflake errors, anyway.

DanCodedThis:
Does it not auto review the new snapshots. Does it only reviewing them upon the first snapshot creation commit?

crates/embucket-functions/src/conversion/to_decimal.rs
```diff
@@ -0,0 +1,342 @@
...
+                ],
+                Volatility::Immutable,
+            ),
+            aliases,
+            r#try,
+        }
+    }
+    fn get_precision_checked(precision_scalar: &ScalarValue) -> DFResult<u8> {
+        let result = match precision_scalar {
+            ScalarValue::Int64(Some(precision)) => u8::try_from(*precision),
```
osipovartem: looks like a pattern
```rust
    /// Tries to convert a scalar to the target integer type
    fn try_convert_scalar<T>(scalar: &ScalarValue) -> Result<T, ()>
    where
        T: TryFrom<i64> + TryFrom<u64> + Copy,
    {
        match scalar {
            ScalarValue::Int64(Some(v)) => T::try_from(*v).map_err(|_| ()),
            ScalarValue::Int32(Some(v)) => T::try_from(i64::from(*v)).map_err(|_| ()),
            ScalarValue::Int16(Some(v)) => T::try_from(i64::from(*v)).map_err(|_| ()),
            ScalarValue::Int8(Some(v)) => T::try_from(i64::from(*v)).map_err(|_| ()),
            ScalarValue::UInt64(Some(v)) => T::try_from(*v).map_err(|_| ()),
            ScalarValue::UInt32(Some(v)) => T::try_from(u64::from(*v)).map_err(|_| ()),
            ScalarValue::UInt16(Some(v)) => T::try_from(u64::from(*v)).map_err(|_| ()),
            ScalarValue::UInt8(Some(v)) => T::try_from(u64::from(*v)).map_err(|_| ()),
            _ => Err(()),
        }
    }
    .....
    fn get_precision_checked(precision_scalar: &ScalarValue) -> DFResult<u8> {
        let precision = Self::try_convert_scalar::<u8>(precision_scalar).map_err(|_| {
            conv_errors::InvalidPrecisionSnafu {
                precision: precision_scalar.clone(),
            }
            .build()
        })?;
      .....
    fn get_scale_checked(scale_scalar: &ScalarValue, precision: u8) -> DFResult<i8> {
        let scale = Self::try_convert_scalar::<i8>(scale_scalar).map_err(|_| {
            conv_errors::InvalidScaleSnafu {
                precision_minus_one: precision - 1,
                scale: scale_scalar.clone(),
            }
            .build()
        })?;
        ...

Comment (DanCodedThis) [reply crates/embucket-functions/src/conversion/to_decimal.rs]: I'll check it out

File: crates/embucket-functions/src/conversion/to_decimal.rs [hunk 4]
```diff
@@ -0,0 +1,342 @@
...
+
+    fn signature(&self) -> &Signature {
+        &self.signature
+    }
+
+    fn return_type(&self, _arg_types: &[DataType]) -> DFResult<DataType> {
+        conv_errors::ReturnTypeFromArgsShouldBeCalledSnafu.fail()?
+    }
+
+    fn return_type_from_args(&self, args: ReturnTypeArgs) -> DFResult<ReturnInfo> {
```
osipovartem, line 230:
```rust
fn return_type_from_args(&self, args: ReturnTypeArgs) -> DFResult<ReturnInfo> {
        use ScalarValue::*;

        let get_precision = |val: &ScalarValue| Self::get_precision_checked(val);
        let get_scale = |val: &ScalarValue, precision| Self::get_scale_checked(val, precision);

        match args.arg_types.len() {
            0 => conv_errors::TooLittleArgumentsSnafu { got: 0, at_least: 1 }.fail(),
            1 => Ok(ReturnInfo::new(DataType::Decimal128(38, 0), true)),
            2 => match &args.scalar_arguments[1] {
                Some(Utf8(..) | Utf8View(..) | LargeUtf8(..)) => Ok(ReturnInfo::new(DataType::Decimal128(38, 0), true)),
                Some(precision) => {
                    let p = get_precision(precision)?;
                    Ok(ReturnInfo::new(DataType::Decimal128(p, 0), true))
                }
                None => conv_errors::NoInputArgumentOnPositionsSnafu { positions: vec![2] }.fail(),
            },
            3 => {
                let second = &args.scalar_arguments[1];
                let third = &args.scalar_arguments[2];
                match (second, third) {
                    (Some(Utf8(..) | Utf8View(..) | LargeUtf8(..)), Some(precision)) => {
                        let p = get_precision(precision)?;
                        Ok(ReturnInfo::new(DataType::Decimal128(p, 0), true))
                    }
                    (Some(precision), Some(scale)) => {
                        let p = get_precision(precision)?;
                        let s = get_scale(scale, p)?;
                        Ok(ReturnInfo::new(DataType::Decimal128(p, s), true))
                    }
                    _ => conv_errors::NoInputArgumentOnPositionsSnafu { positions: vec![1, 2] }.fail(),
                }
            }
            4 => match (&args.scalar_arguments[2], &args.scalar_arguments[3]) {
                (Some(precision), Some(scale)) => {
                    let p = get_precision(precision)?;
                    let s = get_scale(scale, p)?;
                    Ok(ReturnInfo::new(DataType::Decimal128(p, s), true))
                }
                _ => conv_errors::NoInputArgumentOnPositionsSnafu { positions: vec![3, 4] }.fail(),
            },
            n => conv_errors::TooManyArgumentsSnafu { got: n, at_maximum: 4 }.fail(),
        }
    }
```

crates/embucket-functions/src/conversion/to_decimal.rs
osipovartem: try to move format extraction and applying to a separate funcs like **extract_format_arg** and **apply_formatting_if_needed** to reduce size of code

crates/embucket-functions/src/conversion/to_decimal.rs
DanCodedThis: I was wondering if it was hard to read the `fn return_type_from_args`. You answered my question :)

osipovartem:
> Why did insta test fail with new snapshots? I've committed the new ones.

You didn't commit snap files

DanCodedThis:
> For anybody reviewing. Forgot to add the Boolean support (probably works out of the box).

Boolean support only for the `Variant` type: https://docs.snowflake.com/en/sql-reference/functions/to_decimal#usage-notes

DanCodedThis:
> > Why did insta test fail with new snapshots? I've committed the new ones.
> 
> You didn't commit snap files

Oh right, they are unversioned. Thank you!

crates/embucket-functions/src/conversion/to_decimal.rs
DanCodedThis: Took some time, but I added context erroring here

crates/embucket-functions/src/conversion/errors.rs
DanCodedThis: I got what you're saying. An enum in an enum. But this could break the ease of `.context()`.

crates/embucket-functions/src/conversion/to_decimal.rs
```diff
@@ -0,0 +1,330 @@
...
+
+                        Arc::new(StringArray::from(values))
+                    }
+                    None => array.clone(),
+                };
+
+                cast_with_options(
+                    &array,
+                    &DataType::Decimal128(*precision, *scale),
+                    &CastOptions {
```
osipovartem: CastOptions can be moved to let (as format_options) and reused in both cast_with_options calls

osipovartem (approved):

crates/embucket-functions/src/conversion/to_decimal.rs
```diff
@@ -0,0 +1,335 @@
...
+                    TypeSignature::Any(4),
+                ],
+                Volatility::Immutable,
+            ),
+            aliases,
+            r#try,
+        }
+    }
+    /// Tries to convert a scalar to the target integer type
+    fn try_convert_scalar<T>(scalar: &ScalarValue) -> Result<T, conv_errors::Error>
```
ravlio: have you looked at `ScalarValue::cast_to`? Is it the same case?

` ScalarValue::Int32(Some(1)).cast_to(&DataType::Int64)?;`

DanCodedThis: Here we're actually converting the intrinsic value of the numeric scalar to u8 or i8 (not DF, just rust types) for decimals. Probably, naming is misleading.

ravlio (approved):

DanCodedThis:
Unit tests linker (cc) failed. Restarting.

crates/embucket-functions/src/conversion/errors.rs
```diff
@@ -44,6 +55,66 @@ pub enum Error {
...
+
+    #[snafu(display("Invalid scale only allowed from 1 to {precision_minus_one} got: {scale:?}"))]
+    InvalidScale {
+        precision_minus_one: u8,
+        scale: ScalarValue,
+        #[snafu(implicit)]
+        location: Location,
+    },
+
+    #[snafu(display("Too little arguments, got: {got}, expected at least: {at_least}"))]
```
Vedin: nit: better to use `not enough`. It's more understandable and Snowflake using `not enough arguments for function [{function_call}], expected {expected}, got {actual}`. For now it's fine to keep it as is. We still need to refactor this common errors.

crates/embucket-functions/src/conversion/to_decimal.rs
```diff
@@ -0,0 +1,335 @@
...
+    #[must_use]
+    pub fn new(r#try: bool, aliases: Vec<String>) -> Self {
+        Self {
+            signature: Signature::one_of(
+                vec![
+                    //TO_DECIMAL( <expr> )
+                    TypeSignature::Any(1),
+                    //TO_DECIMAL( <expr> [, '<format>' ] )
+                    //TO_DECIMAL( <expr> [, <precision> ] )
+                    TypeSignature::Any(2),
```
Vedin, line 43: Don't use any for 2-4 options. 

- Format is always string. So, I'd say `logical_string` can fit here. 
- Precision is always  from 1 to 38 even it's checked later it's better not to allow this. Probably you should list all integer types for it but maybe you find more elegant solution. It's even important in case of error handing. Wrong type for precision should show: `argument 2 to function TO_DECIMAL needs to be an integer, found: '1.2'` but now it's Invalid precision.
- scale integer as well.

I think it should be possible to list all available options. In case of any troubles or if it looks messy you can use `Signature::user_defined` and then implement `coerce_types` by you own.

crates/embucket-functions/src/conversion/to_decimal.rs
```diff
@@ -0,0 +1,335 @@
...
+#[derive(Debug)]
+pub struct ToDecimalFunc {
+    signature: Signature,
+    aliases: Vec<String>,
+    r#try: bool,
+}
+
+impl ToDecimalFunc {
+    #[must_use]
+    pub fn new(r#try: bool, aliases: Vec<String>) -> Self {
```
Vedin: nit: my personal preference is not using verbs for variable names. Even in this case if `try` can be used as noun. The second things is why using reserved keywords for this? We already had `try_mode` in `to_binary`. What do you think about renaming it?

crates/embucket-functions/src/conversion/to_decimal.rs
```diff
@@ -0,0 +1,335 @@
...
+    /// Tries to convert a scalar to the target integer type
+    fn try_convert_scalar_to_integer<T>(scalar: &ScalarValue) -> Result<T, conv_errors::Error>
+    where
+        T: TryFrom<i128, Error = TryFromIntError>
+            + TryFrom<i64, Error = TryFromIntError>
+            + TryFrom<u64, Error = TryFromIntError>
+            + Copy,
+    {
+        match scalar {
+            ScalarValue::Int64(Some(v)) => {
```
Vedin, line 65: What do you think about using some macro for this similar lines of code? Will it looks better or worse? In general, I'm fine with this. So, I just commented this out and leave a decision for you.

crates/embucket-functions/src/conversion/to_decimal.rs
```diff
@@ -0,0 +1,335 @@
...
+            return conv_errors::InvalidPrecisionSnafu {
+                precision: precision_scalar.clone(),
+            }
+            .fail()?;
+        }
+        Ok(precision)
+    }
+    #[allow(clippy::as_conversions, clippy::cast_possible_wrap)]
+    fn get_scale_checked(scale_scalar: &ScalarValue, precision: u8) -> DFResult<i8> {
+        let scale: i8 = Self::try_convert_scalar_to_integer(scale_scalar)?;
```
Vedin, line 110: what happens here in case of any overflows? Can we get it situation where this randomly becomes valid?

crates/embucket-functions/src/conversion/to_decimal.rs
```diff
@@ -0,0 +1,335 @@
...
+        };
+
+        let format = Self::extract_format_arg(&args.args)?;
+
+        //TODO: should we have type info as before, good datapoint to think about on other types, functions, etc
+        let cast_options = CastOptions {
+            safe: self.r#try,
+            format_options: FormatOptions::default(),
+        };
+        //TODO: override NULL formatting is not working, expected? Visitor somewhere?
```
Vedin: I don't fully get this TODO.
General comment for TODO's: Please add link to the issue if some `TODO` is planned to be merged.

crates/embucket-functions/src/conversion/mod.rs
```diff
@@ -24,6 +26,14 @@ pub fn register_udfs(registry: &mut dyn FunctionRegistry) -> datafusion_common::
         Arc::new(ScalarUDF::from(ToBooleanFunc::new(true))),
         Arc::new(ScalarUDF::from(ToTimeFunc::new(false))),
         Arc::new(ScalarUDF::from(ToTimeFunc::new(true))),
+        Arc::new(ScalarUDF::from(ToDecimalFunc::new(
```
Vedin: Let's have aliases outside of functions registration. So, I'd try to do smth like:
```rust
 pub fn new(try_mode: bool) -> Self {
        let aliases = if try_mode {
            vec![String::from("try_to_number"), String::From("try_to_numeric")]
        } else {
            vec![String::from("to_number"), String::From("to_numeric")]
        };
        ....
```
I remember you mentioned something about aliases. Doesn't the approach above work for you?

test/sql/bronze_scope/sql-reference-functions/Conversion/try_to_decimal.slt
```diff
@@ -6,7 +6,7 @@ SELECT column1 AS orig_string,
   FROM VALUES ('345.123')
   AS T(column1)
 ----
-100039 (22003): Numeric value '345.123' is out of range
+000200: Arrow error: Invalid argument error: 34512 is too large to store in a Decimal128 of precision 4. Max is 9999
```
Vedin, line 9: I'd put a note somewhere about this error change.  Just in case we have an issue with it.

Vedin (changes requested):
some small comments and questions

crates/embucket-functions/src/conversion/to_decimal.rs
DanCodedThis: Sure. Probably should have the same try_mode, is_try, etc everywhere as a standard in the future.

crates/embucket-functions/src/conversion/to_decimal.rs
DanCodedThis: @Vedin I actually started with Coercion, but quickly found that it's not complete yet. There is no `logical_numeric` yet, and no `logical_decimal`. Meaning we can't use `Coercion`, `explicit` or `implicit` even. Writing every signature would be hell. It would be 20+ signatures, probably. This change was suggested by Artem - to check the types at invoke. And we also check some of the args at return type. Additionally, this allows us to throw our errors instead of Coercion errors of Datafusion.

I thought about `Signature::user_defined`, but I think it adds unnecessary indirection. It's probably more "Datafusion idomatic", however, as their docs say it's better to use `Coercion` (or just other signatures in general). Whenever they finish the `Coercion`, it probably will become the go to, and we can revisit our signatures then.

crates/embucket-functions/src/conversion/errors.rs
DanCodedThis: Makes sense, the naming felt off for me also. But couldn't point the finger as to why.

crates/embucket-functions/src/conversion/to_decimal.rs
DanCodedThis: Probably it can not, since `try_convert_...` would have an error. If somebody puts not `i8` or `u8` range value here - it should error (i.e. `value` > `u8::MAX or `i8::MAX`). 

Interesting thing I didn't check does Snowflake error here on `try_`? It should, since it's not part of the casting logic.

crates/embucket-functions/src/conversion/to_decimal.rs
DanCodedThis: This probably needs to be removed.

crates/embucket-functions/src/conversion/mod.rs
DanCodedThis: Probably should, I'll check it out.

test/sql/bronze_scope/sql-reference-functions/Conversion/try_to_decimal.slt
DanCodedThis: Do slts have comments?

crates/embucket-functions/src/conversion/to_decimal.rs
DanCodedThis: Not really, still would be same lines of code, but more indirect. Though it did start as a macro at first, but sadly wasn't necessary here.

test/sql/bronze_scope/sql-reference-functions/Conversion/try_to_decimal.slt
Vedin: I think they should allow `# style comments`. At least it's worth checking.

crates/embucket-functions/src/conversion/to_decimal.rs
DanCodedThis: Actually, on another note it probably would be 100+ signatures, since we can't mix strings with numeric signatures. Will have to write out almost any signature combination in Exact. There is a way to do coercion, but will have to make each logical type, since coercion to one type wouldn't be possible (implicit), because of size differences. This would only remove the need to list out all the string types. Also, decimal can only go with exact still (and there are three types of strings). For brevity’s sake, it would be very messy, even with coercion explicit.

DanCodedThis: Coercion implicit could work for precision and scale. But at that point it still probably like `30..40` signatures.

Vedin: That's not "unnecessary indirection". `coerce_types` will be called during the analyzer step. While `invoke_with_args` is called on execution itself. Even though we do some checks on `return_type_with_args`, I don't see a reason to check any coercions on `invoke_with_args`. 
I don't want it to be a blocker for now. In the common case, I'd avoid using `TypeSignature::Any`. The only exception is when all arguments can be expressions themselves.

DanCodedThis: With `user_defined` will just make an unnecessary calculation for precision and scale. Where we still would need to get them by type in `return_type` at least (so we would just divide the coercion from there to a different function). Also, we would still need to check the format (second arg) at invoke and expr (first arg) at invoke.

DanCodedThis: Sorry for wall text :)

DanCodedThis: > That's not "unnecessary indirection". `coerce_types` will be called during the analyzer step. While `invoke_with_args` is called on execution itself. Even though we do some checks on `return_type_with_args`, I don't see a reason to check any coercions on `invoke_with_args`. I don't want it to be a blocker for now. In the common case, I'd avoid using `TypeSignature::Any`. The only exception is when all arguments can be expressions themselves.

We'd still need to check the types since, for string, we need to format. And any other type probably would be ok, since we already coerced it and checked it at the analyzer step. And since we use arrow cast it should work like a charm. Hmm, I'll check it out tomorrow. Good catch.

Vedin (approved):

Issue #1330, opened 2025-07-15
Create a functional test on embucket working with multiple writers


YaroslavLitvinov:
### To ensure embucket has a reasonable behavior, on at least 2 layers:
* logical sql layer. Here we check if SQL logic is correct. This can be achieved with InMemory object store as it gives no physical conflicts. I guess.
* physical layer, accessed by multiple writers with intent to mutate the same resource

### Following SQL statements treated like a tests, running in conditions described above ^^
* `CREATE TABLE embucket.public.hello(amount number)`
  Assuming underlying schema exists and transaction should pass.
  - On logical layer we expect a success of single statement and the rest should fail
  - On physical layer ...

PR #1348, opened 2025-07-16
e2e test with multiple writers

YaroslavLitvinov:
Related to issue #1330 

What it does:
* Creates multiple writers for object_store in Memory & File .
* Runs the same query in parallel with multiple writers.
* It supports tests queries array, expect one writer succeed and the rest writers fail.
* Run using: 
  ``` bash
  cargo e2e
  ```


rampage644:
@YaroslavLitvinov I think you better to rebase this one: I tried to restart the check job - it still goes to self-hosted runner

DanCodedThis:
@YaroslavLitvinov Just some quick notes (sorry for text wall):
- Futures `join_all` is not parallel, it's concurrent meaning it will use the same thread and interleave the execution (poll all futures until all are complete). 
- To run in true parallel mode, you need to use `tokio::spawn`. Tokio defaults to a multi-threaded runtime with worker thread pools. However, it's not guaranteed that the join all on the `tokio::spawn` handles will run on different worker thread (may run on just one, but not the main thread if not explicitly configured), it depends on how the tokio scheduler decides to do it. 
- Additionally, `tokio::spawn_blocking` exists that uses a separate worker thread pool to run tasks that "block", for example, blocking I/O. Here, "block" actually means that each task is guaranteed to get its own worker thread.
- If you want true parallelism, you can use `join_all` on each handle of the `tokio::spawn` or `tokio::spawn_blocking` (more tricky with the last one).

DanCodedThis:
> @YaroslavLitvinov Just some quick notes (sorry for text wall):
> 
> * Futures `join_all` is not parallel, it's concurrent meaning it will use the same thread and interleave the execution (poll all futures until all are complete).
> * To run in true parallel mode, you need to use `tokio::spawn`. Tokio defaults to a multi-threaded runtime with worker thread pools. However, it's not guaranteed that the join all on the `tokio::spawn` handles will run on different worker thread (may run on just one, but not the main thread if not explicitly configured), it depends on how the tokio scheduler decides to do it.
> * Additionally, `tokio::spawn_blocking` exists that uses a separate worker thread pool to run tasks that "block", for example, blocking I/O. Here, "block" actually means that each task is guaranteed to get its own worker thread.
> * If you want true parallelism, you can use `join_all` on each handle of the `tokio::spawn` or `tokio::spawn_blocking` (more tricky with the last one).

Probably not the point of this test, but just something to keep in mind.

YaroslavLitvinov:
@DanCodedThis We'll see how the current approach works. I agree that it might not cover all use cases as intended. To make it truly concurrent and parallel, the best approach would be to run two instances of Embucket from the command line — and that's it.

DanCodedThis:
> @DanCodedThis We'll see how the current approach works. I agree that it might not cover all use cases as intended. To make it truly concurrent and parallel, the best approach would be to run two instances of Embucket from the command line — and that's it.

As long as the os scheduler assigns them to separate logical cpus (cores or hyper-threads). I think running them o separate computers will truly be parallel :)

DanCodedThis (approved):
@YaroslavLitvinov LGTM

PR #1353, opened 2025-07-17
Allow boolean vs int comparison

### LINKED ISSUE #1345: [BUG] Allow boolean comparison with Int
```sql
select true::boolean = 0
```
Cannot infer common argument type for comparison operation Boolean = Int64

Snowflake returns boolean, comparison works as 0 vs other values (1,2,100, etc.)
```sql
 > select true::boolean = 0;
+-------------------+
| TRUE::BOOLEAN = 0 |
|-------------------|
| False             |
+-------------------+
 > select true::boolean = 10;
+--------------------+
| TRUE::BOOLEAN = 10 |
|--------------------|
| True               |
+--------------------+

osipovartem:
Closes #1345 #1234

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2246,7 +2246,7 @@ dependencies = [
 [[package]]
 name = "datafusion"
 version = "47.0.0"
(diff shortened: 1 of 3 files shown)


DanCodedThis (approved):
@osipovartem LGTM

PR #1359, opened 2025-07-17
`hex_decode_string` scalar function

### LINKED ISSUE #1354: Implement `hex_decode_string` function
https://docs.snowflake.com/en/sql-reference/functions/hex_decode_string

ravlio:
closes #1354


crates/core-executor/src/tests/snapshots/query_alter_iceberg_table.snap
```diff
@@ -1,11 +1,12 @@
 ---
 source: crates/core-executor/src/tests/query.rs
 description: "\"ALTER ICEBERG TABLE test ADD col INT;\""
-snapshot_kind: text
 ---
 Ok(
     [
-        "++",
-        "++",
+        "+--------+",
```
Vedin: weird that this test changed? Is it an old version?

crates/embucket-functions/src/string-binary/hex_decode_string.rs
```diff
@@ -0,0 +1,294 @@
...
+
+impl Default for HexDecodeStringFunc {
+    fn default() -> Self {
+        Self::new(false)
+    }
+}
+
+impl HexDecodeStringFunc {
+    #[must_use]
+    pub fn new(r#try: bool) -> Self {
```
Vedin: The same comment I left for Danil's PR:
> nit: my personal preference is not using verbs for variable names. Even in this case if try can be used as noun. The second things is why using reserved keywords for this? We already had try_mode in `to_binary`. What do you think about renaming it?

test/slt_errors_stats_embucket.csv
```diff
@@ -0,0 +1,10 @@
+filename,statement,error,error_stack_trace
+sql/bronze_scope/sql-reference-functions/String_&_binary/hex_decode_string.slt,SELECT HEX_DECODE_STRING('536E6F77666C616B65'),000200: SQL compilation error: unsupported feature: Function 'hex_decode_string' is not implemented yet. Details: https://docs.snowflake.com/en/sql-reference/functions/hex_decode_string,"0: DataFusion error: This feature is not implemented: Function 'hex_decode_string' is not implemented yet. Details: https://docs.snowflake.com/en/sql-reference/functions/hex_decode_string, at crates/core-executor/src/query.rs:280:44
```
Vedin: probably accidently pushed,

crates/embucket-functions/src/visitors/unimplemented/helper/implemented_functions.csv
```diff
@@ -164,6 +164,7 @@ get_path
 greatest
 grouping
 hex_encode
+hex_decode_string
```
Vedin, line 166: this list is in lexicographical order. Better use `cargo update-functions`. It'll put everything in the right place and automatically add `try_hex_decode_string` here as well.

crates/embucket-functions/src/visitors/unimplemented/generated_snowflake_functions.rs
```diff
@@ -2433,12 +2433,6 @@ pub const STRING_BINARY_FUNCTIONS: &[(&str, FunctionInfo)] = &[
     )
     .with_docs("https://docs.snowflake.com/en/sql-reference/functions/hex_decode_binary")
     ),
-    ("HEX_DECODE_STRING", FunctionInfo::new(
```
Vedin, line 2400: Same comment as bellow. Better use generators to update this file
```bash
python generate_unimplemented_snowflake_functions.py
cargo fmt
```
It'd remove everything registered like `hex_decode_string` and `try_hex_decode_string`. The only issue is you need to run `cargo fmt` after regeneration.

Vedin (changes requested):
some small request as registring `try_*` for this function and avoid using reference words as variable names.

crates/embucket-functions/src/string-binary/hex_decode_string.rs
ravlio: ok lets use try_mode, but if I remember correctly we discussed r#try somewhere

test/slt_errors_stats_embucket.csv
ravlio: will add to gitignore

crates/embucket-functions/src/string-binary/hex_decode_string.rs
DanCodedThis: I like `r#try` but really it's only usable if you need a struct where you will serialize it, i.e. Serde JSON and you need a field named `try`. Sadly, if you're using it in rust it still doesn't allow you to use `self.try` only `self.r#try` which is inconvenient.

I also moved to `try_mode` in `to_decimal`.

Vedin (approved):

PR #1357, opened 2025-07-17
Add custom object type

osipovartem:
Closes #1352 

https://docs.snowflake.com/en/sql-reference/data-types-structured#label-structured-types-specifying-object

In an [Apache Iceberg™ table](https://docs.snowflake.com/en/user-guide/tables-iceberg), the [Apache Iceberg™ data types](https://docs.snowflake.com/en/user-guide/tables-iceberg-data-types) **list, struct, and map** correspond to the structured ARRAY, **structured OBJECT**, and MAP types in Snowflake.

To specify a structured OBJECT type, use the following syntax:
```sql
OBJECT(
  [
    <key> <value_type> [ NOT NULL ]
    [ , <key> <value_type> [ NOT NULL ] ]
    [ , ... ]
  ]
)
```
```sql
CREATE TABLE customer (
  c_id VARCHAR,
  c_name VARCHAR,
  c_address OBJECT(
    state VARCHAR,
    city VARCHAR,
    street VARCHAR,
    zip_code NUMBER
  )
);
```

PR #1366, opened 2025-07-18
update iceberg-rust

JanKaul:
Update iceberg-rust to its most recent version.


osipovartem:
Already added this change in https://github.com/Embucket/embucket/pull/1357/files :)

PR #1374, opened 2025-07-21
Add created_at, updated_at to databases information schema

osipovartem:
This is required for migration to `CREATE DATABASE` sql since we have to return CreateDatabase response with additional created_at, updated_at timestamps
related to #1372


DanCodedThis (approved):
@osipovartem LGTM

PR #1380, opened 2025-07-22
[DOCS] Remove index page, NCU

nikitastryuk:
- Index page removed (now redirects to /essentials/introduction/)
- Page order configuration
- Internal links fixed (should not use /docs prefix)
- NCU (Planned packages upgrade)


docs/src/content/docs/essentials/introduction.md
```diff
@@ -1,6 +1,8 @@
 ---
 title: Introduction
 description: Meet Embucket – an open-source, Snowflake-compatible OLAP lakehouse that marries the simplicity of a single binary with the openness of Apache Iceberg
+sidebar:
+  order: 0
```
nikitastryuk, line 5: A way to control the page order in an autogenerated group (sorted in alphabetical order by default). Lower numbers are displayed higher, bigger - lower. Set 0 for "Introduction" so it appears first. 

CleanShot 2025-07-22 at 18 28 05@2x

cc @rampage644

rampage644 (approved):

PR #1389, opened 2025-07-24
Support null array for timestamp cast

osipovartem:
To fix 
```sql
select cast(null as timestamp);
```


DanCodedThis (approved):
@osipovartem LGTM

rampage644 (approved):

PR #1409, opened 2025-07-28
Support temporary view

### LINKED ISSUE #1392: Temporary views are not supported

osipovartem:
- allow temporary views
- moved snap tests to correct folder
Closes #1392


crates/core-executor/src/query.rs
```diff
@@ -1293,6 +1296,18 @@ impl UserQuery {
         self.created_entity_response()
     }
 
+    #[instrument(name = "UserQuery::create_schema", level = "trace", skip(self), err)]
+    pub async fn create_view(&self, statement: Statement) -> Result<QueryResult> {
+        let mut plan = self.sql_statement_to_plan(statement).await?;
+        match &mut plan {
+            LogicalPlan::Ddl(DdlStatement::CreateView(cv)) => {
```
DanCodedThis, line 1303: Does the parser already have `materialized` that we need to also set to false?

DanCodedThis (approved):
@osipovartem LGTM

crates/core-executor/src/query.rs
osipovartem: materialized is not supported and out of the scope of PR

PR #1425, opened 2025-07-30
e2e s3tables tests

### LINKED ISSUE #1424: Add s3tables tests into e2e

YaroslavLitvinov:
closes #1424

It includes:
* refactor e2e test suite
* changes/alters .env prefixes used to configure e2e tests
* s3tables volume related e2e tests
* Generalize snowflake error, refactor snowflake-rest error
* Fix error message for Diagnostic error
* Now e2e tests also use aws_sdk for assigning s3tabes bucket policies


rampage644 (approved):

crates/core-executor/Cargo.toml
```diff
@@ -58,6 +59,9 @@ insta = { version = "1.42.0", features = ["yaml", "filters"] }
 paste = "1"
 slatedb = { workspace = true }
 dotenv = "0.15.0"
+aws-config = { version = "1.8.3", features = ["behavior-version-latest"] }
```
osipovartem, line 62: Do we have aws modules in workspace? If yes, let's use workspace

osipovartem (approved):

crates/core-executor/Cargo.toml
YaroslavLitvinov: These are dev-dependencies and are currently used exclusively in core-executor. I think it's better to keep them at the crate level, so they don't bloat our project-wide dependencies and won't impact build time.

PR #1429, opened 2025-07-31
CREATE EXTERNAL VOLUME support

osipovartem:
**📦 Add create_volume logic with support for conditional s3tables catalog creation**
- This PR introduces and documents the logic for creating a new volume in the system, supporting various storage backends (file, memory, s3, s3tables). It also adds conditional registration of query catalogs depending on the provider type and session context.

**✅ What’s included:**
- Accepts a logical volume name and a list of CloudProviderParams.
- Supports file, memory, s3, and s3tables as valid provider types.
- Validates presence of required credentials based on provider.
- Persists the volume in the metastore.
- Special handling for s3tables:
  - If the volume is of type s3tables, we optionally create a corresponding catalog.
  - This behavior is controlled by the session variable DISABLE_S3TABLES_CATALOG_CREATION.
  - If this variable is set to "true" (case-insensitive), catalog creation is skipped — useful for tests or sandboxed environments

**🧪 Example (skip catalog creation in test):**
```sql
SET DISABLE_S3TABLES_CATALOG_CREATION = 'true';
```

**📋 Error handling:**
The function returns clear, user-facing errors in the following cases:
- Unknown provider type.
- Missing required credentials for the provider.
- Failure to write volume metadata to the metastore.
- Failure to register catalog (unless intentionally skipped via session var).

**Next steps**
- Migrate creete volume endpoint to SQL
- **disable catalog creation when s3tables volumes added since we need to call create database sql for that**


crates/core-executor/src/tests/sql/ddl/volume.rs
```diff
@@ -0,0 +1,45 @@
...
+        ))"],
+    snapshot_path = "volume"
+);
+
+test_query!(
+    s3tables,
+    "SELECT volume_name, volume_type FROM slatedb.meta.volumes",
+    setup_queries = [
+        // disable s3tables catalog creation
+        "SET DISABLE_S3TABLES_CATALOG_CREATION = true",
```
YaroslavLitvinov, line 36: Since the DISABLE_S3TABLES_CATALOG_CREATION option is coupled with volume creation, it makes sence to make it part of CloudProviderParams along with other options.

YaroslavLitvinov (approved):
In overall LGTM.
Providing credentials over SQL when creating s3 volume would also require somehow masking credentials from being saved into query history.

DanCodedThis (approved):
@osipovartem LGTM

crates/core-executor/src/tests/sql/ddl/volume.rs
osipovartem: It will be removed in the next PR since we will not create catalog for s3tables 
the logic will be
- create external volume (only create volume)
- create database NAME external volume ident (create database as usual, same for s3tables: register catalog)

DanCodedThis (approved):

PR #1435, opened 2025-07-31
Migrate to external volume sql

osipovartem:
- create external volume without catalog for s3 tables(only create volume)
- create database NAME external volume ident (create database as usual, same for s3tables: register catalog)
Next steps
- migrate all scripts
- cleanup code


crates/api-ui/src/volumes/handlers.rs
```diff
@@ -88,14 +94,69 @@ pub async fn create_volume(
...
+        MetastoreVolumeType::S3(vol) => {
+            let credentials_str = match &vol.credentials {
+                Some(AwsCredentials::AccessKey(creds)) => format!(
+                    "CREDENTIALS=(AWS_KEY_ID='{}' AWS_SECRET_KEY='{}')",
+                    creds.aws_access_key_id, creds.aws_secret_access_key
+                ),
+                _ => return VolumeMissingCredentialsSnafu.fail().context(CreateSnafu)?,
+            };
+            format!(
+                "STORAGE_PROVIDER = 'S3' STORAGE_BASE_URL = '{:?}' STORAGE_ENDPOINT = '{:?}' {}",
```
rampage644: I think we won't need STORAGE_ENDPOINT unless it was specified

crates/api-ui/src/volumes/handlers.rs
```diff
@@ -88,14 +94,69 @@ pub async fn create_volume(
...
+        .execution_svc
+        .query(
+            &session_id,
+            &format!(
+                "CREATE EXTERNAL VOLUME '{ident}' STORAGE_LOCATIONS = ((NAME = '{ident}' {vol_params}))",
+            ),
+            QueryContext::default(),
+        )
+        .await
+        .context(QuerySnafu)?;
```
rampage644: I think here and below we should keep CreateSnafu since this is the context: we are trying to create a volume and this is the context we want to preserve

crates/core-metastore/src/models/volumes.rs
```diff
@@ -144,8 +144,6 @@ pub struct S3TablesVolume {
     pub endpoint: Option<String>,
     #[validate(nested)]
     pub credentials: AwsCredentials,
-    #[validate(length(min = 1), custom(function = "validate_bucket_name"))]
-    pub database: String,
```
rampage644, line 148: Could you explain what is the change? Previously, when created s3 table volume we specified database it mapped to. Is it that user now has to call create database explicitly after volume created?

osipovartem: Yes, since volume is just a volume, not catalog/database

crates/api-ui/src/volumes/handlers.rs
osipovartem: Previous error was from metastore, new one from core-executor, I'll rename

rampage644: @YaroslavLitvinov Do you have a better idea what should we keep here?

crates/core-metastore/src/models/volumes.rs
rampage644: What happens now when two databases going to be created with a single s3 table volume?

osipovartem: They both will look into single s3 table bucket as a links to schemas and tables

crates/api-ui/src/volumes/handlers.rs
```diff
@@ -88,14 +94,69 @@ pub async fn create_volume(
...
+                    "CREDENTIALS=(AWS_KEY_ID='{}' AWS_SECRET_KEY='{}')",
+                    creds.aws_access_key_id, creds.aws_secret_access_key
+                ),
+                AwsCredentials::Token(_) => {
+                    return VolumeMissingCredentialsSnafu.fail().context(CreateSnafu)?;
+                }
+            };
+            format!(
+                "STORAGE_PROVIDER = 'S3TABLES' STORAGE_ENDPOINT = '{:?}' \
+                STORAGE_AWS_ACCESS_POINT_ARN = '{}' {}",
```
rampage644: Is it s3 table bucket arn? Not sure aws access point arn applicable here

crates/df-catalog/src/catalog_list.rs
```diff
@@ -112,41 +109,35 @@ impl EmbucketCatalogList {
             .metastore
             .get_volume(&volume_ident.to_string())
             .await
+            .context(MetastoreSnafu)?
+            .map_or(
```
rampage644, line 113: What this pattern (map_or) does?

rampage644: I am confused since we have contextualized an error previously here already - I feel `.context` and `.map_or` should merge here

rampage644 (approved):
Overall, it seems this PR is one in the series: i was not good at review `df-catalog` part of changes. 

I feel errors should be redone - I'll defer to @YaroslavLitvinov to weight in here, happy to further discuss (in this PR or following) about how to approach them here.

overall lgtm, just to unblock

crates/api-ui/src/volumes/handlers.rs
osipovartem: We can use only STORAGE_AWS_ROLE_ARN or STORAGE_AWS_ACCESS_POINT_ARN 
```cloudProviderParams (for Amazon S3) ::=
  STORAGE_PROVIDER = '{ S3 | S3GOV }'
  STORAGE_AWS_ROLE_ARN = '<iam_role>'
  STORAGE_BASE_URL = '<protocol>://<bucket>[/<path>/]'
  [ STORAGE_AWS_ACCESS_POINT_ARN = '<string>' ]
  [ STORAGE_AWS_EXTERNAL_ID = '<external_id>' ]
  [ ENCRYPTION = ( [ TYPE = 'AWS_SSE_S3' ] |
              [ TYPE = 'AWS_SSE_KMS' [ KMS_KEY_ID = '<string>' ] ] |
              [ TYPE = 'NONE' ] ) ]
  [ USE_PRIVATELINK_ENDPOINT = { TRUE | FALSE } ]```
  
  Seems STORAGE_AWS_ACCESS_POINT_ARN is more correct here

File: crates/df-catalog/src/catalog_list.rs [hunk 6]
```diff
@@ -191,14 +180,34 @@ impl EmbucketCatalogList {
...
-            .collect()
+            .context(df_catalog_error::CoreSnafu)?;
+        for db in databases {
+            let volume = self
+                .metastore
+                .get_volume(&db.volume)
+                .await
+                .context(MetastoreSnafu)?
+                .map_or(
+                    MissingVolumeSnafu {
```
YaroslavLitvinov, line 197: Because metastore returns Option, we have no other choices than creating error here (

crates/df-catalog/src/catalog_list.rs
YaroslavLitvinov: Yeah, because of Metastore. Operation returned Option instead of error.
Old Metastore's issue

YaroslavLitvinov (approved):
LGTM

crates/api-ui/src/databases/error.rs
```diff
@@ -8,6 +8,12 @@ use snafu::prelude::*;
...
 pub enum Error {
+    #[snafu(display("Create database query error: {source}"))]
+    CreateQuery {
+        source: core_executor::Error,
+        #[snafu(implicit)]
+        location: Location,
+    },
     #[snafu(display("Create database error: {source}"))]
     Create {
         source: core_metastore::Error,
```
DanCodedThis, line 19: No need for this error anymore

crates/api-ui/src/volumes/error.rs
```diff
@@ -7,6 +7,12 @@ use snafu::prelude::*;
...
 pub enum Error {
+    #[snafu(display("Create volume query error: {source}"))]
+    CreateQuery {
+        source: core_executor::Error,
+        #[snafu(implicit)]
+        location: Location,
+    },
     #[snafu(display("Create volume error: {source}"))]
     Create {
         source: core_metastore::Error,
```
DanCodedThis, line 18: No need for this error anymore also

crates/api-ui/src/volumes/handlers.rs
```diff
@@ -88,14 +94,77 @@ pub async fn create_volume(
...
+        MetastoreVolumeType::S3(vol) => {
+            let region = vol.region.clone().unwrap_or_default();
+            let credentials_str = match &vol.credentials {
+                Some(AwsCredentials::AccessKey(creds)) => format!(
+                    "CREDENTIALS=(AWS_KEY_ID='{}' AWS_SECRET_KEY='{}' REGION='{region}')",
+                    creds.aws_access_key_id, creds.aws_secret_access_key,
+                ),
+                _ => return VolumeMissingCredentialsSnafu.fail().context(CreateSnafu)?,
+            };
+            let base_url = vol.bucket.clone().unwrap_or_default();
```
DanCodedThis, line 114: What is the default here?

crates/api-ui/src/volumes/handlers.rs
```diff
@@ -88,14 +94,77 @@ pub async fn create_volume(
...
+            let endpoint_str = vol
+                .endpoint
+                .as_ref()
+                .map(|e| format!(" STORAGE_ENDPOINT = '{e}'"))
+                .unwrap_or_default();
+            format!(
+                "STORAGE_PROVIDER = 'S3' STORAGE_BASE_URL = '{base_url}'{endpoint_str}{credentials_str}",
+            )
```
DanCodedThis, line 122: Nitpick: Is there no blank space between `endpoint_str` and `credential str`?

crates/api-ui/src/volumes/handlers.rs
```diff
@@ -88,14 +94,77 @@ pub async fn create_volume(
...
+        .get_volume(&ident)
         .await
-        .map(Volume::from)
-        .context(CreateSnafu)?;
+        .context(GetSnafu)?
+        .map_or(
+            metastore_error::VolumeNotFoundSnafu { volume: ident }.fail(),
+            Ok,
+        )
+        .context(GetSnafu)?;
```
DanCodedThis: Do you need to map here? And double context. Maybe better just to add an error here? Shouldn't it already work with this type of error out of the box. Perhaps use `#[snafu(source)] for automatic `?`, no ?

crates/df-catalog/src/catalog_list.rs
DanCodedThis: **Disregard** @osipovartem Just using context on an option works. Snafu is ok with this. The problem probably is with error mappings. But this is solvable, providing a special error with `#[snafu(source)]` pointing to that, so that you could use the `.context(new_error)?`. **Disregard**

crates/df-catalog/src/catalog_list.rs
DanCodedThis: @osipovartem Here adding double context makes sense (without using map_or). Since first, you have a result of trying to get the thing. And second is if the thing exists at all. You can add without the `#[snafu(source)]` an error something like this here after, I think this would be totally valid hence of `Result<Option<...>>`:

```rust
...
pub enum Error {
...
    #[snafu(display("Not found"))]
    VolumeNotFound {
       #[snafu(implicit)]
       location: Location,
    }
...
} 
...
```

Something like this:
```rust
use snafu::{OptionExt, ResultExt};

let test = Ok(Some(1));

        let yesy = test.context(errors::FormatMustBeNonNullScalarValueSnafu)?.context(errors::FormatMustBeNonNullScalarValueSnafu)?;
```        

Works and is totally valid. I would just change the second error to NotFoundSomethingSnafu though.

crates/api-ui/src/volumes/handlers.rs
DanCodedThis: @osipovartem I wrote a comment down later about `Result<Option<...>>`.

DanCodedThis (approved):
@osipovartem Other than erroring, looks good to me

crates/api-ui/src/volumes/handlers.rs
DanCodedThis: There is another default lower here also

osipovartem: unwrap_or_default is empty string for Option<String>

crates/df-catalog/src/catalog_list.rs
DanCodedThis:
```rust
use snafu::{OptionExt, ResultExt};

let test = Ok(Some(1));

        let yesy = test.context(errors::FormatMustBeNonNullScalarValueSnafu)?.context(errors::FormatMustBeNonNullScalarValueSnafu)?;
```        

Works and is totally valid. I would just change the second error to NotFoundSomethingSnafu though.

DanCodedThis (approved):

PR #1441, opened 2025-08-01
Feat/copy into external

JanKaul:
This is a proposal on how to implement the Snowflake COPY INTO statement for external locations. The main idea is to use the [STORAGE_INTEGRATION](https://docs.snowflake.com/en/sql-reference/sql/copy-into-table#additional-cloud-provider-parameters) parameter as a Reference for an Embucket Volume. [STORAGE_INTEGRATION](https://docs.snowflake.com/en/sql-reference/sql/create-storage-integration)s are very similar to Snowflake External Volumes and we could tell users that we automatically create STORAGE_INTEGRATIONS for all Embucket Volumes. Behind the scenes STORAGE_INTEGRATIONS and EXTERNAL VOLUMES both refer to Embucket Volumes.

```
COPY INTO mytable
  FROM 's3://mybucket/data/files'
  STORAGE_INTEGRATION = test_volume
  FILE_FORMAT = (TYPE = CSV);
```

Once we know which volume the statement refers to, we use that to load the `ObjectStore`, register the `ObjectStore` with the current context and create a Listing Table for the TableScan.

This is currently just a sketch of the solution and needs further testing. Let me know what you think.

### DIFF
--- a/crates/core-executor/src/query.rs
+++ b/crates/core-executor/src/query.rs
@@ -107,7 +107,7 @@ use sqlparser::ast::{
 use std::collections::hash_map::Entry;
 use std::collections::{HashMap, HashSet};
 use std::fmt::Write;
-use std::ops::ControlFlow;
+use std::ops::{ControlFlow, Deref};
 use std::result::Result as StdResult;
 use std::sync::Arc;
 use tracing_attributes::instrument;
@@ -1156,11 +1156,17 @@ impl UserQuery {
                 .register_object_store(url.object_store().as_ref(), object_store);
 
             let config = ListingTableConfig::new(url.clone());
-            let config = if let Some(format) = create_file_format(&file_format)? {
+            let config = if let Some((format, infer_schema)) = create_file_format(&file_format)? {
                 let options = ListingOptions::new(format);
-                config
-                    .with_listing_options(options)
-                    .with_schema(into_provider.schema())
+                let schema = if infer_schema {
+                    options
+                        .infer_schema(&self.session.ctx.state(), &url)
+                        .await
+                        .context(ex_error::DataFusionSnafu)?
+                } else {
+                    into_provider.schema()
+                };
+                config.with_listing_options(options).with_schema(schema)
             } else {
                 config
             };
@@ -1184,6 +1190,12 @@ impl UserQuery {
             };
 
             let input = builder.build().context(ex_error::DataFusionSnafu)?;
+
+            let input = if input.schema().as_arrow() == into_provider.schema().deref() {
+                input
+            } else {
+                cast_input_to_target_schema(Arc::new(input), &into_provider.schema())?
+            };
 
             let plan = LogicalPlanBuilder::insert_into(
                 input,
@@ -3056,14 +3068,23 @@ pub fn get_kv_option<'a>(options: &'a KeyValueOptions, key: &str) -> Option<&'a 
         .map(|opt| opt.value.as_str())
 }
 
-fn create_file_format(file_format: &KeyValueOptions) -> Result<Option<Arc<dyn FileFormat>>> {
+fn create_file_format(
+    file_format: &KeyValueOptions,
+) -> Result<Option<(Arc<dyn FileFormat>, bool)>> {
     if let Some(format_type) = get_kv_option(file_format, "type") {
         if format_type.eq_ignore_ascii_case("parquet") {
-            Ok(Some(Arc::new(ParquetFormat::default())))
+            Ok(Some((Arc::new(ParquetFormat::default()), true)))
         } else if format_type.eq_ignore_ascii_case("csv") {
-            Ok(Some(Arc::new(CsvFormat::default())))
+            let infer_schema = get_kv_option(file_format, "parse_header")
+                .is_some_and(|x| x.to_lowercase() == "true");
+            let has_header =
+                get_kv_option(file_format, "skip_header").is_some_and(|x| x.to_lowercase() == "1");
+            Ok(Some((
+                Arc::new(CsvFormat::default().with_has_header(has_header)),
+                infer_schema,
+            )))
         } else if format_type.eq_ignore_ascii_case("json") {
-            Ok(Some(Arc::new(JsonFormat::default())))
+            Ok(Some((Arc::new(JsonFormat::default()), true)))
         } else {
             Err(ex_error::UnsupportedFileFormatSnafu {
                 format: format_type,
--- a/crates/core-executor/src/query.rs
+++ b/crates/core-executor/src/query.rs
@@ -30,6 +30,8 @@ use datafusion::catalog::MemoryCatalogProvider;
 use datafusion::catalog::{CatalogProvider, SchemaProvider};
 use datafusion::datasource::DefaultTableSource;
 use datafusion::datasource::default_table_source::provider_as_source;
+use datafusion::datasource::listing::{ListingTable, ListingTableConfig, ListingTableUrl};
+use datafusion::execution::object_store::ObjectStoreUrl;
 use datafusion::execution::session_state::{SessionContextProvider, SessionState};
 use datafusion::logical_expr::{self, col};
 use datafusion::logical_expr::{LogicalPlan, TableSource};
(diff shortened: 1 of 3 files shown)


DanCodedThis (approved):
@JanKaul I support this approach. This would allow us to do something like this (we had before): https://github.com/Embucket/compatibility-test-suite/blob/main/tpc-h/snowflake/setup.sql#L87-L94

P.S. It would be cool, to just write copy into something from the whole s3 bucket (automatic table listing/recognition). Not sure Snowflake itself supports even :)

crates/core-executor/src/query.rs
```diff
@@ -1050,22 +1052,86 @@ impl UserQuery {
...
+                .ctx
+                .table_provider(insert_reference.clone())
+                .await
+                .context(ex_error::DataFusionSnafu)?;
+
+            let plan = LogicalPlanBuilder::insert_into(
+                input,
+                insert_reference,
+                Arc::new(DefaultTableSource::new(into_provider)),
+                InsertOp::Append,
```
rampage644, line 1180: Is it possible to get out from the statement?

crates/core-executor/src/query.rs
```diff
@@ -1050,22 +1052,86 @@ impl UserQuery {
...
+            into,
+            from_obj,
+            stage_params,
+            ..
+        } = statement
+        else {
             return ex_error::OnlyCopyIntoStatementsSnafu.fail();
         };
-        if let Some(from_obj) = from_obj {
+        let Some(from_obj) = from_obj else {
```
rampage644, line 1120: I am also in favor of dropping previous functionality and in case we don't get external location but stage (that we don't support atm), error out.

crates/core-executor/src/query.rs
```diff
@@ -1050,22 +1052,86 @@ impl UserQuery {
...
+                }
+                (None, _) => {
+                    todo!()
+                }
+            };
+
+            let url = ObjectStoreUrl::parse(&location.value).unwrap();
+            self.session
+                .ctx
+                .register_object_store(url.as_ref(), object_store)
```
rampage644: Curios if we can deregister once statement is done

rampage644 (commented):
lgtm 👍 , ready to approve once convert from draft (finished + tests added)

crates/core-executor/src/query.rs
osipovartem: ObjectStoreRegistry has only register_store and get_store
So seems we cannot deregister

JanKaul: Yeah looks that way

crates/core-executor/src/query.rs
JanKaul: What do you mean by out of the statement?

crates/core-executor/src/query.rs
JanKaul: I didn't want to be too invasive and just add the additional functionality. I don't know if the old functionality is used somewhere. But we can also do that.

JanKaul:
There are a couple of things (proper error handling, alias handling, file-type) still missing. I will add those and then convert this into a real PR.

crates/core-executor/src/query.rs
rampage644: I've meant something like `LOAD_MODE = { FULL_INGEST | ADD_FILES_COPY }`, to specify whether the operation should overwrite existing data or append new files. I couldn’t find a direct equivalent in the SQL syntax and was looking for a setting that explicitly defines this behavior.

crates/core-executor/src/query.rs
rampage644: Understood, I am pretty sure old functionality is not used anywhere.

crates/core-executor/src/query.rs
JanKaul: I would suggest to only support FULL_INGEST, even though it might be less preformant. ADD_FILES_COPY is a bit more involved and has the issue that the Parquet files might not be in the Iceberg compatible form.

JanKaul:
@osipovartem @rampage644 @DanCodedThis This PR is ready for review. It works with csv files.

However, it currently has issues with Parquet files. I think Parquet files store their field names internally and they seem to be case sensitive. This leads to problems when trying to COPY INTO an Embucket Iceberg table because we always store the field identifiers in lowercase. I've found no way to cast the Parquet schema to the Iceberg schema.
- When trying to explicitly cast the TableScan output it throws an Error telling me that the lowercase column doesn't exist
- When I try to enforce the Schema for the TableScan, it just return NULLs because the RecordBatches contain no Fields with the lowercase names

But I think these issues can be handled in future PRs. Just the csv functionality simplifies certain setups.

crates/core-executor/src/query.rs
```diff
@@ -1053,22 +1060,111 @@ impl UserQuery {
...
+
+            let url = ListingTableUrl::parse(&location.value).context(ex_error::DataFusionSnafu)?;
+            self.session
+                .ctx
+                .register_object_store(url.object_store().as_ref(), object_store);
+
+            let config = ListingTableConfig::new(url.clone());
+            let config = if let Some((format, infer_schema)) = create_file_format(&file_format)? {
+                let options = ListingOptions::new(format);
+                let schema = if infer_schema {
```
DanCodedThis: @JanKaul Can we build the schema by hand in lowercase, just using the inferred names from it? (Perhaps I'm misunderstanding the problem)

DanCodedThis (approved):
@JanKaul LGTM

crates/core-executor/src/query.rs
JanKaul: If I use a schema with lowercase fields I just get NULL values. I think datafusion tries to find the lowercase field in the parquet file. It doesn't exist(only the uppercase version), and hence datafusion inserts a NULL value.

rampage644 (approved):

JanKaul:
It should now also work with Parquet files

crates/core-executor/src/query.rs
```diff
@@ -2866,27 +2985,107 @@ fn apply_show_filters(sql: String, filters: &[String]) -> String {
...
+                }
+            }
+            None
+        })
+        .ok_or_else(|| {
+            ex_error::FieldNotFoundInInputSchemaSnafu {
+                field_name: name.to_string(),
+            }
+            .build()
+        })?;
```
DanCodedThis: Using `use snafu::OptExt` you can use `.context()` with Options. This is a suggesting instead of  `.ok_or_else()`. I think you can also pass down field there with `.build()` or `.fail()` on the error itself. But context is more useful with automatic field recognition of the error, passing down the error with `#[snafu(source)]`.

crates/core-executor/src/query.rs
```diff
@@ -2445,6 +2517,53 @@ impl UserQuery {
...
+        stage_params: StageParamsObject,
+    ) -> Result<Arc<dyn ObjectStore + 'static>> {
+        let object_store = match (stage_params.storage_integration, stage_params.credentials) {
+            (Some(volume), _) => {
+                let volume = self
+                    .metastore
+                    .get_volume(&volume)
+                    .await
+                    .context(ex_error::MetastoreSnafu)?
+                    .ok_or_else(|| ex_error::VolumeNotFoundSnafu { volume }.build())?;
```
DanCodedThis: Using `use snafu::OptExt` you can use `.context()` with Options. This is a suggesting instead of  `.ok_or_else()`. Double context is better for now. Sometime in the future, we should just return a result here without an option.

DanCodedThis: Or you can go the: 
```rust
let Ok(Some(volume) = self.metastore.get_volume(&volume) else {
    return just a singular error
}
```
(not if let)

crates/core-executor/src/query.rs
```diff
@@ -2911,3 +3110,52 @@ pub fn get_volume_kv_option(
...
+            };
+
+            Ok(Some((Arc::new(csv_format), infer_schema)))
+        } else if format_type.eq_ignore_ascii_case("json") {
+            Ok(Some((Arc::new(JsonFormat::default()), true)))
+        } else {
+            Err(ex_error::UnsupportedFileFormatSnafu {
+                format: format_type,
+            }
+            .build())
```
DanCodedThis: nitpick: No need to wrap it in `Err(...)` just using `.fail()` instead of `.build()` should work here

DanCodedThis (approved):
@JanKaul LGTM, only concern is erroring

crates/core-executor/src/query.rs
DanCodedThis: I think it's already irrelevant at this point. However, what I was trying to say is to build the schema by hand with the names in the format it needs to be in to work (upper or lower). But I digress :)

crates/core-executor/src/query.rs
```diff
@@ -1091,29 +1099,93 @@ impl UserQuery {
         self.status_response()
     }
 
-    #[instrument(
-        name = "UserQuery::copy_into_snowflake_query",
-        level = "trace",
-        skip(self),
-        err
-    )]
+    // #[instrument(
```
osipovartem: removed for debugging? :)

JanKaul: Good catch! No weirdly my `rust-analyzer` has more options when I disable the tracing macros.

osipovartem (approved):

PR #1443, opened 2025-08-01
test: Add Spark write and Embucket read verification test

YevheniiNiestierov:
test_read_spark.py: ensures the database/schema exist, uses Spark to create a table and execute insert/update/delete operations, then reads back via Embucket to verify those changes persisted correctly

### DIFF
--- a/test/integration_tests/conftest.py
+++ b/test/integration_tests/conftest.py
@@ -1,0 +1,90 @@
+import pytest
+import os
+from pyspark.sql import SparkSession
+
+CATALOG_URL = "http://localhost:8080"
+WAREHOUSE_ID = "test_db"
+AWS_REGION = os.getenv("AWS_REGION")
+AWS_ACCESS_KEY_ID = os.getenv("AWS_ACCESS_KEY_ID")
+AWS_SECRET_ACCESS_KEY = os.getenv("AWS_SECRET_ACCESS_KEY")
+
+@pytest.fixture()
+def rest_spark_session():
+    spark_session = spark_session_factory(config_type='file_catalog', app_name='FileCatalogTest')
+    yield spark_session
+    spark_session.stop()
+
+
+@pytest.fixture()
+def s3_spark_session():
+    spark_session = spark_session_factory(config_type='s3_catalog', app_name='S3CatalogTest')
+    yield spark_session
+    spark_session.stop()
+
+
+def spark_session_factory(
+        config_type: str,
+        app_name: str,
+        config_overrides: dict = None
+    ):
+    """
+    The actual function that builds the SparkSession.
+
+    Args:
+        config_type (str): The type of configuration to use.
+                           Expected values: 'file_catalog' or 's3_catalog'.
+        app_name (str): The name for the Spark application.
+        config_overrides (dict): A dictionary of Spark configs to add or
+                                 override the base configuration.
+
+    Returns:
+        SparkSession: An initialized SparkSession object.
+    """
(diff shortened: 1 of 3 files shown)


test/integration_tests/test_read_spark.py
```diff
@@ -0,0 +1,85 @@
...
+    ensure_db_and_schema(emb)
+
+    spark_client = PySparkClient()
+    spark_create_insert_update_delete(spark_client)
+
+    embucket_read_verify(emb)
+    logger.info("Test completed")
+
+
+if __name__ == "__main__":
```
rampage644: Let's get rid of custom main entrypoints and instead rely on [`pytest`](https://docs.pytest.org/en/stable/) directly.

This also has a consequences of:
 * test cases to be renamed with `test_****`
 * embucket_client and spark_client becoming a fixtures to be reused across all other tests

rampage644: * custom if else statements becoming assserts, thus simplifying test cases

YevheniiNiestierov: sure, will rewrite with pytest

test/integration_tests/conftest.py
```diff
@@ -0,0 +1,90 @@
...
+               .config("spark.sql.defaultCatalog", "rest")
+
+    # --- Configuration Type 2: Direct S3 Access with Explicit Keys ---
+    elif config_type == 's3_catalog':
+        builder.config("spark.driver.memory", "15g") \
+               .config("spark.jars.packages", "org.apache.iceberg:iceberg-spark-runtime-3.5_2.12:1.9.1,org.apache.hadoop:hadoop-aws:3.3.4") \
+               .config("spark.hadoop.fs.s3a.impl", "org.apache.hadoop.fs.s3a.S3AFileSystem") \
+               .config("spark.hadoop.fs.s3a.access.key", AWS_ACCESS_KEY_ID or "") \
+               .config("spark.hadoop.fs.s3a.secret.key", AWS_SECRET_ACCESS_KEY or "") \
+               .config("spark.hadoop.fs.s3a.endpoint", f"s3.{AWS_REGION}.amazonaws.com" if AWS_REGION else "") \
```
rampage644: Let's drop this endpoint - aws sdk is smart enough to figure out itself

test/integration_tests/conftest.py
```diff
@@ -0,0 +1,90 @@
+import pytest
+import os
+from pyspark.sql import SparkSession
+
+CATALOG_URL = "http://localhost:8080"
+WAREHOUSE_ID = "test_db"
```
rampage644, line 6: I now realized why embucket fixture might also could be helpful: `warehouse_id` in this context means a database on the embucket side. This is a quirk coming from Iceberg terminology. So, it could be coming from the embucket fixture.

It's totally ok to leave it as is for now, but this means we will have to create a database in embucket client that is `test_db`.

test/integration_tests/test_read_spark.py
```diff
@@ -0,0 +1,97 @@
...
+    spark.sql(
+        f"""CREATE TABLE IF NOT EXISTS {SCHEMA}.{TABLE} (
+            id INT,
+            page_name STRING,
+            category STRING
+        )"""
+    )
+
+    # clear
+    logger.info("Deleting all rows from table")
```
rampage644: let's drop logging everywhere

YevheniiNiestierov: why?

rampage644 (approved):
overall lgtm

rampage644 (approved):

Issue #1438, opened 2025-07-31
Add sane default settings for memory pools and disk manager


rampage644:
At the moment, we do not provide custom (or specifically designed) `MemoryPool` and `DiskManagerConfig` settings to the `RuntimeEnvBuilder` and datafusion uses its own default which are
 * UnboundedMemoryPool (no limit)
 * 100GB disk space

The goal is to let user specify what memory pool to use and how much temp disk space should be used for spilling. Potentially using shared mempry pool across different sessions/queries.

DanCodedThis:
@osipovartem Should it be closed?

PR #1448, opened 2025-08-04
Add configurable memory and disk shared pools management 

osipovartem:
Related to #1438

This PR introduces new configuration options that allow fine-grained control over memory and disk resource management in the **runtime_env** setup. The goal is to make memory usage more predictable and debuggable across workloads by supporting multiple memory pool strategies and optional per-consumer tracking.

### ✅ Summary of changes:
- Added support for selecting memory pool type via --mem-pool-type (**greedy or fair**)
- Introduced configurable memory pool size (--mem-pool-size-mb)
- Introduced disk pool size limit (--disk-pool-size-mb) for spillable execution
- Added optional TrackConsumersPool wrapper (--mem-enable-track-consumers-pool) to monitor memory usage per consumer
- All options can be controlled via CLI flags or environment variables
---
### 🧠 Memory pool types

| Pool Type | Description | When to Use |
|------------|-------------|-------------|
| `Greedy`  | Default. All operators attempt to acquire as much memory as possible and hold it until explicitly released. This provides maximum throughput but risks starving other operators under high load. | Use for simple, predictable pipelines or when performance is the top priority. |
| `Fair`    | Memory is divided equally across tasks, providing more stable and predictable memory usage patterns across concurrent queries. May result in slower execution but better resource isolation. | Use in multi-tenant or concurrent query environments where fairness is critical. |

### ⚙️ Defaults (if no configuration is provided)

If no explicit configuration is provided via CLI or environment variables, the following **default settings** are applied:

| Resource          | Default Implementation              | Limit         | Notes |
|------------------|--------------------------------------|---------------|-------|
| Memory Pool   | `UnboundedMemoryPool`               | No limit      | No memory usage enforcement; suitable only for testing or controlled environments. |
| Disk Spilling | `DiskManagerConfig::NewOs` → `DiskManager` | 100 GB        | Temp files are created in a temporary OS directory. Limit enforced via `max_temp_directory_size`. |

The `DiskManagerConfig::NewOs` variant tells DataFusion to create temporary files in an OS-managed temp directory (e.g., `/tmp` on Unix), and applies the default maximum directory size of **100 GB** unless overridden.

**> ✅ We strongly recommend configuring memory and disk usage explicitly in production deployments.**
---
### 🧪 Memory tracking
If --mem-enable-track-consumers-pool is set to true, the selected memory pool will be wrapped in TrackConsumersPool, which tracks memory consumption per logical consumer (up to 5 by default). This is useful for:
- Monitoring which operators or stages consume the most memory
- Diagnosing leaks or inefficiencies
- Building internal memory usage dashboards or logs

This feature introduces slight overhead and is optional.

### 💾 Disk management
The --disk-pool-size-mb option sets an upper bound on how much temporary spill data (e.g., for sort or hash joins) can be stored on disk. If the limit is exceeded, queries may fail or be forced to back off. This helps enforce soft limits in low-disk environments.


JanKaul (approved):

crates/core-executor/src/utils.rs
```diff
@@ -30,27 +31,29 @@ use std::collections::HashMap;
 use std::sync::Arc;
 use strum::{Display, EnumString};
 
-#[derive(Clone)]
+#[derive(Clone, Debug)]
```
DanCodedThis, line 34: nitpick: `#[derive(..., Default)]` with `#[default]` on `MemPoolType::Greedy` in the enum

DanCodedThis (approved):
@osipovartem LGTM

crates/core-executor/src/service.rs
```diff
@@ -88,6 +97,71 @@ impl CoreExecutionService {
...
+
+    #[allow(clippy::unwrap_used, clippy::as_conversions)]
+    pub fn runtime_env(
+        config: &Config,
+        catalog_list: Arc<EmbucketCatalogList>,
+    ) -> Result<Arc<RuntimeEnv>> {
+        let mut rt_builder = RuntimeEnvBuilder::new().with_object_store_registry(catalog_list);
+
+        if let Some(memory_limit_mb) = config.mem_pool_size_mb {
+            const NUM_TRACKED_CONSUMERS: usize = 5;
```
rampage644, line 121: Why 5?

crates/core-executor/src/service.rs
```diff
@@ -88,6 +97,71 @@ impl CoreExecutionService {
...
+
+        // set disk limit
+        if let Some(disk_limit) = config.disk_pool_size_mb {
+            let disk_limit_bytes = (disk_limit as u64) * 1024 * 1024;
+
+            let disk_manager = DiskManager::try_new(DiskManagerConfig::NewOs)
+                .context(ex_error::DataFusionSnafu)?;
+
+            let disk_manager = Arc::try_unwrap(disk_manager)
+                .map_err(|_| {
```
rampage644: Is it possible to avoid creating Datafusion error here?

rampage644: Going straight to `.context()` semantics?

crates/core-executor/src/utils.rs
osipovartem: MemPoolType::Greedy is already default in Cli opts

crates/core-executor/src/service.rs
osipovartem: This number is default for datafusion cli and also used for only debagging porpose 
The bigger value will slow down the engine

crates/core-executor/src/service.rs
osipovartem: I can add new try_unwrap error to use with .context

crates/core-executor/src/utils.rs
osipovartem: we have embucket_version which should be set in separate Default
so we cannot have both derive Default and impl Default for Config

DanCodedThis (approved):

PR #1450, opened 2025-08-04
Fix stale metadata in cache

### LINKED ISSUE #1439: Embucket doesn't see snapshots created by Spark
For tests use https://github.com/Embucket/embucket/blob/main/test/intergration_tests/clients.py

**Spark is able to see data inserted by Embucket
Embucket is not able to see data inserted by Spark**

```
embucket_client = EmbucketClient()
embucket_client.volume()
embucket_client.sql("CREATE DATABASE IF NOT EXISTS test_db EXTERNAL_VOLUME = 'test'")
embucket_client.sql("CREATE SCHEMA IF NOT EXISTS test_db.public")
embucket_client.sql("DROP TABLE IF EXISTS test_db.public.test")
embucket_client.sql("CREATE TABLE IF NOT EXISTS test_db.public.test (id int, name string)")
embucket_client.sql(f"INSERT INTO test_db.public.test VALUES (1, 'test_name')")

spark_client = PySparkClient()
spark_client.sql(f"INSERT INTO public.test VALUES (2, 'test_name2')")
spark_client.sql(f"SELECT * FROM public.test")
spark_client.sql(f"SELECT count(*) FROM public.test")

print(embucket_client.sql(f"SELECT count(*) FROM test_db.public.test")['result']['rows'])

```

the output is
```
+---+----------+
| id|      name|
+---+----------+
|  2|test_name2|
|  1| test_name|
+---+----------+

+--------+
|count(1)|
+--------+
|       2|
+--------+

[[1]]
```

osipovartem:
Closes #1439 

We should always rely on the original schema provider instead of the cache, because the underlying Iceberg catalog may have updated the table metadata outside of SQL (e.g., **via direct iceberg catalog API calls**). In such cases, our cache could contain
stale metadata and ignore the latest snapshot updates.

Therefore, we remove the cache lookup logic and always fetch from the original schema.

### DIFF
--- a/crates/df-catalog/src/schema.rs
+++ b/crates/df-catalog/src/schema.rs
@@ -45,22 +45,22 @@ impl SchemaProvider for CachingSchema {
     #[allow(clippy::as_conversions)]
     #[tracing::instrument(name = "CachingSchema::table", level = "debug", skip(self), err)]
     async fn table(&self, name: &str) -> Result<Option<Arc<dyn TableProvider>>, DataFusionError> {
-        if let Some(table) = self.tables_cache.get(name) {
-            Ok(Some(Arc::clone(table.value()) as Arc<dyn TableProvider>))
+        // NOTE: We should always rely on the original schema provider instead of the cache,
+        // because the underlying Iceberg catalog may have updated the table metadata outside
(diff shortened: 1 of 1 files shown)


DanCodedThis (approved):
@osipovartem LGTM

PR #1452, opened 2025-08-04
Limit maximum number of concurrent queries

### LINKED ISSUE #1137: Implement concurrent query limits and workload control
- [ ] Add configuration parameter `MAX_CONCURRENCY_LEVEL`
 - [ ] Refuse incoming queries once there are `MAX_CONCURRENCY_LEVEL` amount already processing

osipovartem:
Closes #1137 
This pull request introduces a concurrency limit mechanism to restrict the number of queries that can be executed simultaneously. The goal is to prevent system overload by limiting parallel query execution and protecting shared resources such as CPU, memory, or I/O.

### ✅ Summary of Changes
- Added a new CLI/config argument
  - This controls the maximum number of queries that can be processed simultaneously. Default: 100.
- Integrated semaphore (tokio::sync::Semaphore) logic into the query execution path

### 📌 Motivation
This addition ensures that the system is protected from overload caused by too many concurrent queries running at once. Rather than relying on downstream timeouts or resource exhaustion, we fail fast with a clear error when the concurrency limit is exceeded.
This is particularly useful for:
- Deployments under constrained environments (e.g., local development, CI, or small cloud instances).
- Preventing resource saturation due to accidental or malicious query floods.

### 🔄 Behavior
- If the number of concurrent queries is below the configured limit, the query is allowed to run.
- If the limit is reached:
  - The query fails immediately with a ConcurrencyLimitExceeded error (via try_acquire()).
  - **If blocking behavior is preferred, this can be replaced with .acquire().await.**

### 🧪 Testing
This logic was tested by submitting multiple concurrent long-running queries (e.g., SELECT sleep(5)). The semaphore behaves correctly by allowing only the configured number of simultaneous executions. Additional queries fail fast when the limit is exceeded.

### DIFF
--- a/crates/core-executor/src/tests/mod.rs
+++ b/crates/core-executor/src/tests/mod.rs
@@ -3,3 +3,57 @@ pub mod query;
 pub mod service;
 pub mod snowflake_errors;
 pub mod sql;
+
+use datafusion::arrow::array::Int64Array;
+use datafusion::arrow::datatypes::DataType;
+use datafusion::logical_expr::{ColumnarValue, ScalarUDF, Volatility};
+use datafusion_common::ScalarValue;
+use datafusion_expr::{ScalarFunctionImplementation, create_udf};
+use std::sync::Arc;
+use std::thread;
+use std::time::Duration;
+
+/// Returns a custom scalar UDF named `sleep`, which simulates a delay by blocking the current thread.
+///
+/// # Purpose
+/// This UDF is intended primarily for testing purposes, to simulate long-running queries or
+/// evaluate the behavior of concurrency limits and timeout mechanisms within the query execution engine.
+///
+/// # SQL Usage
+/// ```sql
+/// SELECT sleep(5);
+/// ```
+///
+/// This will block the executing thread for 5 seconds before returning the value `5`.
+///
+/// # Notes
+/// - This UDF performs blocking (`thread::sleep`), and should **not** be used in production settings,
+///   as it will block the execution thread and may reduce system throughput.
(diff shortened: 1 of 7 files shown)


crates/core-executor/src/tests/mod.rs
```diff
@@ -3,3 +3,57 @@ pub mod query;
...
+            ColumnarValue::Array(array) => {
+                let array = array
+                    .as_any()
+                    .downcast_ref::<Int64Array>()
+                    .expect("Expected Int64Array");
+                array.value(0)
+            }
+            _ => 0,
+        };
+        thread::sleep(Duration::from_secs(seconds.try_into().unwrap()));
```
DanCodedThis, line 48: nitpick: Why not tokio sleep?

DanCodedThis (approved):
@osipovartem LGTM

rampage644 (approved):

JanKaul (approved):

crates/core-executor/src/tests/mod.rs
osipovartem: tokio sleep is async
we cannot use async in ScalarFunctions since they are sync

crates/embucketd/src/cli.rs
```diff
@@ -155,6 +155,14 @@ pub struct CliOpts {
     )]
     pub sql_parser_dialect: Option<String>,
 
+    #[arg(
+        long,
+        env = "MAX_CONCURRENCY_LEVEL",
+        default_value = "100",
```
rampage644: Let's decrease to 8

YaroslavLitvinov (approved):
LGTM

PR #1463, opened 2025-08-05
`convert_timezone` scalar function

ravlio:


osipovartem (approved):

DanCodedThis (approved):
@ravlio LGTM

crates/embucket-functions/src/datetime/convert_timezone.rs
```diff
@@ -121,292 +107,173 @@ impl ScalarUDFImpl for ConvertTimezoneFunc {
...
+            };
+
+            let arr = match &args[2] {
+                ColumnarValue::Array(arr) => arr.to_owned(),
+                ColumnarValue::Scalar(v) => v.to_array()?,
+            };
+
+            (source_tz, target_tz, arr)
+        } else {
+            return internal_err!("Invalid number of arguments");
```
DanCodedThis, line 188: Needs correct erroring

crates/embucket-functions/src/datetime/convert_timezone.rs
```diff
@@ -121,292 +107,173 @@ impl ScalarUDFImpl for ConvertTimezoneFunc {
...
+        };
+
+        match (source_tz, target_tz, arr) {
+            (Some(source_tz), Some(target_tz), arr) => {
+                let source_tz = source_tz
+                    .parse::<Tz>()
+                    .map_err(|_| CantParseTimezoneSnafu.build())?;
+                let target_tz = target_tz
+                    .parse::<Tz>()
+                    .map_err(|_| CantParseTimezoneSnafu.build())?;
```
DanCodedThis: you can make an error with snfaue source and context here withou map err

crates/embucket-functions/src/datetime/convert_timezone.rs
```diff
@@ -121,292 +107,173 @@ impl ScalarUDFImpl for ConvertTimezoneFunc {
...
+    // Create a naive datetime from the timestamp (treating it as epoch time but interpreting as naive)
+    let naive_dt = DateTime::from_timestamp(secs, nanos)
+        .ok_or("Invalid timestamp")?
+        .naive_utc(); // Get the naive representation
+
+    // Interpret this naive datetime as being in the source timezone
+    let source_dt = source_timezone
+        .from_local_datetime(&naive_dt)
+        .single()
+        .ok_or("Failed to convert source time to source timezone")?;
```
DanCodedThis: for options there is a snafu `OptExt` to make them context with errors too

just use snafu::OptExt and you can `.context(error)?` any error on the option type

DanCodedThis (approved):

crates/embucket-functions/src/datetime/convert_timezone.rs
ravlio: I believe it doesn't need to be handled properly, it is edge case that potentially shouldn't be called because we have strict signature that allows only 2 or 3 arguments. There errors such as `internal_err!("Invalid source_timestamp_tz type")` `internal_err!("Invalid source_timestamp_ntz type")` etc that are impossible to reach

DanCodedThis: > I believe it doesn't need to be handled properly, it is edge case that potentially shouldn't be called because we have strict signature that allows only 2 or 3 arguments. There errors such as `internal_err!("Invalid source_timestamp_tz type")` `internal_err!("Invalid source_timestamp_ntz type")` etc that are impossible to reach

unreachable!() macro call?

crates/embucket-functions/src/datetime/convert_timezone.rs
```diff
@@ -1,116 +1,139 @@
...
         match args.arg_types.len() {
             2 => {
                 let tz = match &args.scalar_arguments[0] {
-                    Some(ScalarValue::Utf8(Some(part))) => part.clone(),
+                    Some(
+                        ScalarValue::Utf8(Some(part))
+                        | ScalarValue::Utf8View(Some(part))
+                        | ScalarValue::LargeUtf8(Some(part)),
+                    ) => part.clone(),
                     _ => return internal_err!("Invalid target_tz type"),
```
DanCodedThis, line 129: This function errors should be reachable right? You can check out the **regexp** directory. I use `TooManyArgumanets` and `NotEnoughArguments`, you too can make those errors here also. I also used such an approach in the conversion functions. These errors are defined multiple times for each module for now. But they are really neat.

DanCodedThis (approved):
@ravlio LGTM, up to you with errors

rampage644:
@ravlio is it ready for review? You will likely need to rebase it to merge

ravlio:
@rampage644 yes please review and approve 🙏

rampage644 (approved):

PR #1467, opened 2025-08-06
Add query execution timeout

### LINKED ISSUE #1465: When query runs more than predefined timeout, it should be cancelled

osipovartem:
Closes #1465 
This PR introduces a timeout mechanism for query execution to prevent runaway or hanging queries from consuming resources indefinitely. The timeout duration is configurable via query_timeout_secs.

### Key changes:
- Wrapped query execution in tokio::time::timeout. If the timeout is exceeded, the system returns a structured QueryTimeout error.
- Ensured that all query executions — whether successful, failed, or timed out — are recorded in the history_store with appropriate status and metadata.
- This improves observability, traceability, and operational stability by guaranteeing that every query attempt is logged, even in edge cases.

### Why it matters:
- Prevents long-running queries from monopolizing system resources.
- Ensures complete auditability of all queries, even failed or aborted ones.
- Improves user feedback in cases where queries take too long.

### DIFF
--- a/crates/core-executor/src/utils.rs
+++ b/crates/core-executor/src/utils.rs
@@ -35,6 +35,7 @@ use strum::{Display, EnumString};
 pub struct Config {
     pub embucket_version: String,
     pub sql_parser_dialect: Option<String>,
+    pub query_timeout_secs: u64,
     pub max_concurrency_level: usize,
     pub mem_pool_type: MemPoolType,
     pub mem_pool_size_mb: Option<usize>,
@@ -47,6 +48,7 @@ impl Default for Config {
         Self {
             embucket_version: "0.1.0".to_string(),
             sql_parser_dialect: None,
+            query_timeout_secs: 1200, // 20 minutes
             max_concurrency_level: 100,
             mem_pool_type: MemPoolType::default(),
             mem_pool_size_mb: None,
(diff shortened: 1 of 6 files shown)


crates/core-executor/src/utils.rs
```diff
@@ -47,6 +48,7 @@ impl Default for Config {
         Self {
             embucket_version: "0.1.0".to_string(),
             sql_parser_dialect: None,
+            query_timeout_secs: 1200, // 20 minutes
```
DanCodedThis, line 51: Is this the same value as snowflake?

DanCodedThis (approved):
@osipovartem LGTM

crates/core-executor/src/utils.rs
osipovartem: `STATEMENT_TIMEOUT_IN_SECONDS` is 3600 by [default](https://community.snowflake.com/s/article/Solution-Statement-reached-its-statement-or-warehouse-timeout-of-xxx-second-s-and-was-canceled), but our snowflake account has different value
Снимок экрана 2025-08-06 160648

rampage644 (approved):

YaroslavLitvinov (approved):

PR #1470, opened 2025-08-06
Refactor sessions

### LINKED ISSUE #1602: Improve sessions management, instrument sessions related code
* Currently global write lock is acquired when deleting expired sessions, where more appropriate mechanism is possible.
* Current session management is not mirrored in exported telemetry spans, which also can be improved

YaroslavLitvinov:
Closes #1602

### Refactor sessions management interface
* Update sessions interface
* Replace session expiry Mutex by AtomicI64
* Improve sessions locks
* Add tracing instrument for sessions related functions


crates/api-sessions/src/session.rs
```diff
@@ -46,7 +47,7 @@ impl SessionStore {
 
             for session_id in session_ids {
                 tracing::trace!("Deleting expired: {}", session_id);
-                sessions.remove(&session_id);
+                let _ = self.execution_svc.delete_session(session_id).await;
```
DanCodedThis: This is what was written originally. There is a deadlock here, you acquire the read lock, which is only dropped in the end, but you're trying to acquire a write lock on each delete. If you fix the deadlock (drop it before asking for delete), this means that we can delete a session in use. Since there is time between the original loop of getting expired session, and the deletion. A session will update its expiry, but is already marked for deletion. Also, each delete will need to acquire its own lock, instead of deleting together (slower execution).


P.S. Can't point out all the lines, since GH decided to make the file smaller.

DanCodedThis:
@YaroslavLitvinov The tests won't complete since there is deadlock, as I described. They have been running for over 40 minutes.

osipovartem (approved):

crates/api-sessions/src/session.rs
```diff
@@ -29,24 +29,26 @@ impl SessionStore {
...
+                tracing::trace!("Starting to delete expired for: {}", now);
+                //Sadly can't use `sessions.retain(|_, session| { ... }`, since the `OffsetDatetime` is in a `Mutex`
+
+                for (session_id, session) in sessions.iter() {
+                    let expiry = session.expiry.lock().await;
+                    if *expiry <= now {
+                        session_ids.push(session_id.clone());
+                    }
                 }
             }
```
DanCodedThis: @YaroslavLitvinov When the drop happens, we don't go to execute deletion. Instead, we go to the next queue item of this lock. For example, you marked the session for deletion (added it to the list), but then the next await queue item is to use this session to do a query. What happens? We update the expiry of this session, but the session is marked for deletion, so when the query finishes, even though the expiry is updated we still delete this session when we get to the deletion item await in queue. But a person still tries t, by this id, get session to do the next query, since the session shouldn't be deleted (we didn't give them then new session id). This is what _originally_ was written, and this is a **data race**. You can test this yourself just lower both expiration values to something like 15 seconds. And try to query and query, etc. At some point you'll start getting errors or generating infinite amount of sessions.

DanCodedThis (changes requested):
This will introduce a data race

crates/api-sessions/src/session.rs
DanCodedThis: Deadlock fixed, but data race acquired

crates/api-sessions/src/session.rs
YaroslavLitvinov: @DanCodedThis I wouldn't neither consider original solution as safe, as it was acquiring a global lock so it could be resulted in congestion at least. Instead I wanted to minimize impact of global lock acquiring, because sessions was monopolized until every `expiry.lock().await` completes.

DanCodedThis: @YaroslavLitvinov If there is lock on sessions, for example for deletion, the expiry is not touched or locked for sure. It can be in a state of closing the lock (update the expiry before giving away control). But if sessions are in a lock by the deletion job, expiry shouldn't make the deletion job wait any extra time. 

I would disconnect the `DedicatedExecutor` from the sessions before introducing this. Otherwise, we are going back to the code that had a data race. And we are reintroducing it.

For snowflake cli it shouldn't matter (the data race, at least, I imagine). For UI, not so sure.

DanCodedThis (approved):
@YaroslavLitvinov Very nice, LGTM

PR #1478, opened 2025-08-07
fix unwrap in merge_query

JanKaul:
## Summary
- Replace unwrap() with proper error handling in merge_query
- Add DataFusionIcebergError import and proper error chain conversion

## Test plan
- Verify build passes
- Ensure error handling flows correctly through the system

### DIFF
--- a/crates/core-executor/src/query.rs
+++ b/crates/core-executor/src/query.rs
@@ -67,7 +67,6 @@ use datafusion_iceberg::DataFusionTable;
 use datafusion_iceberg::catalog::catalog::IcebergCatalog;
 use datafusion_iceberg::catalog::mirror::Mirror;
 use datafusion_iceberg::catalog::schema::IcebergSchema;
-use datafusion_iceberg::error::Error as DataFusionIcebergError;
 use datafusion_iceberg::table::DataFusionTableConfigBuilder;
 use df_catalog::catalog::CachingCatalog;
 use df_catalog::catalog_list::CachedEntity;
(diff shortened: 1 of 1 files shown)


crates/core-executor/src/query.rs
```diff
@@ -1314,13 +1315,13 @@ impl UserQuery {
...
-                    //TODO Return proper Error
-                    #[allow(clippy::unwrap_used)]
                     DataFusionTableConfigBuilder::default()
                         .enable_data_file_path_column(true)
                         .enable_manifest_file_path_column(true)
                         .build()
-                        .unwrap(),
+                        .map_err(DataFusionIcebergError::from)
+                        .map_err(IcebergError::from)
+                        .context(ex_error::IcebergSnafu)?,
```
DanCodedThis: This probably can be done using just one `.context()`. You can convert errors with the `From` trait (this how we've been able to use `.context()` with `Result` from Datafusion in **embucket-functions**). And if you need to pass down an error to `.context()` use `#[snafu(source)]` on the field of your error (struct) to pass down the concrete type of that error you want to pass down.

rampage644: Second this, should be doable without `map_err` and single context

DanCodedThis (approved):
@JanKaul LGTM, I was just writing a comment about error conversion and `.map_err()` :) But you were faster

JanKaul:
The the error wasn't handled properly in `iceberg-rust` that's why the issue surfaced here. Once that was fixed, it's a simple fix here.

PR #1483, opened 2025-08-07
Respond with 503 on connection failures related to Object Store

### LINKED ISSUE #1496: Error saved in query history not the same as showed to user
### :memo: Issue
Instead of saving query error as `SnowflakeError`, which is tend to be displayable error for queries in Emucket, we save it in query history as `core_executor::Error`. Also saved error differs from error displayed to user.

### :gear: Implementation
Error needs to be somehow converted to the `SnowlakeError`, but it can't be done with `From::from trait` which consumes self upon conversion, when result is not yet even returned. As it saves query result at the latest stage of query execution, error should be converted to displayable error without consuming error itself, so something like `SnowflakeError::from_execution_error(&self)` should be used.

YaroslavLitvinov:
closes #1496 

* Temporary lint fix for [utoipa issue](https://github.com/juhaku/utoipa/issues/1420), directive macro added directly into multiple files
* Refactor `SnowflakeError` replacing `SnowflakeError::from(self)` by `SnowflakeError::from_executor_error(&self)` so error is not consumed upon conversion, and error can be saved to the query history when query fails (Before it was saving a `core_executor::Error`)
* Introduced StatusCode for SnowflakeError
* "Downgrade" `SnowflakeError`, so it creates shorten custom error text for context rich errors
* Respond with 503 status code in `snowflake-rest-api` when having s3 connection issues
* status codes doesn't change in `api-ui`
* e2e tests with injected s3 connection falures, and with 503 status_code checks
* Tests connection failures in Objects store accessed through Metastore, for:
  * Tests create table
  * create database
  * create schema 
* Tests connection failures in Objects store accessed from s3 Volume, for:
  * create table
  * insert into table


crates/core-executor/src/snowflake_error.rs
```diff
@@ -2,31 +2,35 @@
...
 
-// How SLT tests are used in Snowflake error conversion?
-// Database engine has a variety of errors, and you need to have error's structure to be able to match it
-// and then properly return appropriate Snowflake error. Currently error_stack_trace available in logs and
-// provides extended error context. It is helpful for cathing occasional errors. But with SLT tests
-// you can get all the errors sources returned by Embucket, since slt runner produces slt_errors_stats_embucket.csv
-// file having all the errors occured during a test run, including error_stack_trace column.
+#[derive(Debug, Eq, PartialEq)]
+#[repr(u16)]
+pub enum StatusCode {
```
rampage644: should it be in `status_code.rs`?

rampage644: Also, makes sense to take a look at https://github.com/GreptimeTeam/greptimedb/blob/main/src/common/error/src/status_code.rs#L22 - originally stacked error was insipired by their approach, but status code was left out scope.

Perhaps not a bad idea to align errors/status codes as they do.

crates/api-snowflake-rest/src/error.rs
```diff
@@ -117,23 +117,15 @@ impl IntoResponse for Error {
...
-                | core_executor::Error::DataFusionLogicalPlanMergeSource { .. }
-                | core_executor::Error::DataFusionLogicalPlanMergeJoin { .. }
-                | core_executor::Error::LogicalExtensionChildCount { .. }
-                | core_executor::Error::MergeFilterStreamNotMatching { .. }
-                | core_executor::Error::MatchingFilesAlreadyConsumed { .. }
-                | core_executor::Error::MissingFilterPredicates { .. }
-                | core_executor::Error::RegisterCatalog { .. } => {
+            Self::Execution { source } => {
+                // use status code defined in core executor
+                http::StatusCode::from_u16(source.status_code_u16()).unwrap_or_else(|e| {
```
rampage644: So this basically moves status code into core-executor, since here we try to return it as is?

crates/core-executor/src/snowflake_error.rs
YaroslavLitvinov: > Also, makes sense to take a look at https://github.com/GreptimeTeam/greptimedb/blob/main/src/common/error/src/status_code.rs#L22 - originally stacked error was insipired by their approach, but status code was left out scope.
> 
> Perhaps not a bad idea to align errors/status codes as they do.

I didn't copied their StatusCode, as they have custom codes.
I won't added a dependency on http for our executor, that is why I added own enum.
As status codes logically related to any transport, I added it in place where it can be accessible by both snowflake, api-ui. 

Though in our case transports have own status codes and it is not ready to be applied on both of them.

crates/core-executor/src/status_code.rs
```diff
@@ -0,0 +1,97 @@
...
+    if let MetastoreError::UtilSlateDB { source, .. } = error {
+        let error = source.as_ref();
+        match error {
+            DbError::Database { error, .. }
+            | DbError::KeyGet { error, .. }
+            | DbError::KeyDelete { error, .. }
+            | DbError::KeyPut { error, .. }
+            | DbError::ScanFailed { error, .. } => {
+                if let SlateDBError::ObjectStoreError(_obj_store_error) = error {
+                    StatusCode::ServiceUnavailable
```
rampage644: Perhaps this is not the best place to define status code for the service, but I don't think I can think about the better. However, one suggestion from my is to decouple if from http codes if possible (or makes sense).

On the other hand I feel going status code route is basically duplicating having each error have its own type. For example, we might have status code `TableNotFound = 4001`, but we already have an error `MetastoreError::TableNotFound` (we actually don't, but let's imagine it is so).

rampage644 (approved):
I am ready to approve test related changes, however I don't fully get the problem with errors. Likely will need to reiterate on what is our state with errors and we're going the direction we want to go.

crates/core-executor/src/snowflake_error.rs
rampage644: Yeah, perhaps it was too idealistic to try to decouple transport level codes in the very first place

crates/api-ui/src/router.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(clippy::needless_for_each)]
```
DanCodedThis, line 1: I wouldn't add those, since they are global on file. And we're getting this warning from a crate that's not part of our codebase, so it shouldn't affect our clippy ci.

crates/api-snowflake-rest/src/error.rs
DanCodedThis: No way to add `.context()` with `#[snafu(source)]` here?

crates/core-executor/src/error.rs
```diff
@@ -513,6 +514,13 @@ pub enum Error {
     },
 }
 
+impl Error {
+    #[must_use]
+    pub fn to_snowflake_error(&self) -> SnowflakeError {
+        SnowflakeError::from_executor_error(self)
+    }
+}
+
```
DanCodedThis, line 523: Why not a `From` trait?

crates/core-executor/src/snowflake_error.rs
```diff
@@ -110,54 +113,96 @@ impl From<Error> for SnowflakeError {
...
+        .filter(|s| !s.is_empty())
+        .copied()
+        .collect::<Vec<_>>()
+        .join(" ");
+    if subtext.is_empty() {
+        error
+    } else {
+        format!("{subtext}: {error}")
+    }
+}
```
DanCodedThis, line 155: This could be done as an iterator with `for_each` (or better even pass a static dispatch iterator instead of an array of slices) to build a `String` without the need to collect to `Vec`, that would still be turned in to a `String` in the end (one less alloc and faster execution also). Take a look at the `regexp_substr_all` for a reference.

crates/core-executor/src/tests/e2e/tests_e2e.rs
```diff
@@ -733,24 +922,134 @@ async fn test_e2e_file_store_single_executor_bad_aws_creds_s3_volume_insert_shou
...
+    unsafe {
+        std::env::set_var(
+            format!("{E2E_BAD_METASTORE_AWS_CREDS_PREFIX}AWS_ACCESS_KEY_ID"),
+            "1",
+        );
+        std::env::set_var(
+            format!("{E2E_BAD_METASTORE_AWS_CREDS_PREFIX}AWS_SECRET_ACCESS_KEY"),
+            "1",
+        );
+    }
```
DanCodedThis, line 954: You sure it needs to be unsafe?

crates/embucketd/src/main.rs
```diff
@@ -1,3 +1,6 @@
+// Set this clippy directive to suppress clippy::needless_for_each warnings
+// until following issue will be fixed https://github.com/juhaku/utoipa/issues/1420
+#![allow(clippy::needless_for_each)]
```
DanCodedThis, line 3: Here also.

crates/core-executor/src/snowflake_error.rs
```diff
@@ -246,108 +279,76 @@ fn datafusion_error(df_error: DataFusionError) -> SnowflakeError {
...
                 }
-            } else if err.is::<iceberg_rust::error::Error>() {
-                if let Ok(e) = err.downcast::<iceberg_rust::error::Error>() {
-                    iceberg_error(*e)
-                } else {
-                    unreachable!()
+                .build()
+            } else if let Some(e) = err.downcast_ref::<object_store::Error>() {
+                CustomSnafu {
+                    message: e.to_string(),
```
DanCodedThis: Why are we downcasting here?

DanCodedThis (approved):
@YaroslavLitvinov LGTM, some nitpicks

crates/core-executor/src/error.rs
YaroslavLitvinov: As From::from is consuming self in query history. So I replaced From::from by this and kept only single implementation to be used for query history and here.

crates/core-executor/src/tests/e2e/tests_e2e.rs
YaroslavLitvinov: this is how it should be used, according to this:
https://doc.rust-lang.org/std/env/fn.set_var.html

crates/embucketd/src/main.rs
YaroslavLitvinov: I think green builds more important than code purity, it is temporary

crates/core-executor/src/snowflake_error.rs
YaroslavLitvinov: as it is `Box<dyn>` and we need a typed Error

rampage644 (approved):

PR #1487, opened 2025-08-07
engine: `regexp_like`

### LINKED ISSUE #1485: engine: Missing `regexp_like` function
Impl regexp like using utils regexp iterator approach.

DanCodedThis:
Closes #1485

- [x] `regexp_like` impl
- [x] Uses an iterator approach for performance and reusability
- [x] Docs
- [x] Insta tests
- [x] SLT tests

P.S. Small clippy error fixes

Reference: https://docs.snowflake.com/en/sql-reference/functions/regexp_like

### DIFF
--- a/crates/core-executor/src/tests/mod.rs
+++ b/crates/core-executor/src/tests/mod.rs
@@ -43,7 +43,7 @@ fn sleep_udf() -> ScalarUDF {
                     .expect("Expected Int64Array");
                 array.value(0)
             }
-            _ => 0,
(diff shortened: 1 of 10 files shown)


DanCodedThis:
Checking slt test revealed something I think we don't support: https://docs.snowflake.com/en/sql-reference/functions-regexp#example-of-using-metacharacters-in-a-single-quoted-string-constant

osipovartem (approved):

PR #1489, opened 2025-08-07
engine: `regexp_replace`

### LINKED ISSUE #1488: engine: Missing `regexp_replace` function
Impl `regexp_replace` function using the utils iterator approach

DanCodedThis:
Closes #1488

- [x] `regexp_replace` impl
- [x] Uses an iterator approach for performance and reusability
- [x] Uses a performative string slicing approach, instead of an iterative per-match replacement
- [x] Docs
- [x] Insta tests
- [x] SLT tests

P.S. Small clippy error fixes. Also, this does not include the last reference example test since our insta tests do not register our rewriters in context

Reference: https://docs.snowflake.com/en/sql-reference/functions/regexp_substr_all

### DIFF
--- a/crates/embucket-functions/src/visitors/functions_rewriter.rs
+++ b/crates/embucket-functions/src/visitors/functions_rewriter.rs
@@ -50,6 +50,32 @@ impl VisitorMut for FunctionsRewriter {
                             span: *span,
                         });
                     }
+                    if let FunctionArguments::List(FunctionArgumentList { args, .. }) = args
+                        && let Some(FunctionArg::Unnamed(FunctionArgExpr::Expr(replacement))) =
+                        args.get_mut(2)
+                        //third arg may be a replacement
+                        && let Expr::Value(ValueWithSpan {
+                                               value: SingleQuotedString(value),
+                                               span,
+                                           }) = replacement
+                    {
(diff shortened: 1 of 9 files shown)


crates/embucket-functions/src/visitors/functions_rewriter.rs
```diff
@@ -50,6 +50,32 @@ impl VisitorMut for FunctionsRewriter {
...
+                        && let Some(FunctionArg::Unnamed(FunctionArgExpr::Expr(replacement))) =
+                        args.get_mut(2)
+                        //third arg may be a replacement
+                        && let Expr::Value(ValueWithSpan {
+                                               value: SingleQuotedString(value),
+                                               span,
+                                           }) = replacement
+                    {
+                        let value = if let Ok(regex) = regex::Regex::new(r"\\(\d)") {
+                            tracing::error!("Made a regex: {:?}", regex);
```
ravlio: do we need it?

DanCodedThis: the regex - yes, but I imagine this is about the tracing, so no - no :)

osipovartem (approved):

osipovartem (approved):

PR #1491, opened 2025-08-08
Fix unquote storage-integration identifier

JanKaul:
## Summary
- Removes quotes from storage-integration identifiers in CopyIntoSnowflake statements
- Sanitizes storage_integration parameter by trimming leading and trailing quotes


DanCodedThis (approved):
@JanKaul LGTM

PR #1494, opened 2025-08-08
Allow snowflake formats in TO_CHAR func

### LINKED ISSUE #1493: Unsupported DATE format in to_char func
Snowflake [formats ](https://docs.snowflake.com/en/sql-reference/functions-conversion#label-date-time-format-conversion)
DF supported [formats](https://docs.rs/chrono/latest/chrono/format/strftime/index.html)

Add convertor between different types of formats

osipovartem:
Closes #1493
```sql
SELECT to_char(col::DATE, 'YYYYMMDD)
```
Snowflake [formats](https://docs.snowflake.com/en/sql-reference/functions-conversion#label-date-time-format-conversion)
DF supported [formats](https://docs.rs/chrono/latest/chrono/format/strftime/index.html)

Added convertor between different types of date formats.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -66,6 +66,7 @@ jobs:
         with:
           toolchain: ${{ matrix.toolchain }}
           components: clippy
+          rust-version: "1.88.0"
       - name: Run clippy
         run: cargo clippy --all-targets --workspace
   # semver:
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -66,7 +66,6 @@ jobs:
         with:
           toolchain: ${{ matrix.toolchain }}
           components: clippy
-          rust-version: "1.88.0"
       - name: Run clippy
         run: cargo clippy --all-targets --workspace
   # semver:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -119,5 +119,3 @@ missing_panics_doc = { level = "allow", priority = 2 }
 significant_drop_tightening = { level = "allow", priority = 2 }
 module_name_repetitions = { level = "allow", priority = 2 }
 option_if_let_else = { level = "allow", priority = 2 }
-collapsible_if = "allow"
-cloned_ref_to_slice_refs = "allow"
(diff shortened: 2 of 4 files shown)


DanCodedThis:
@osipovartem Can you call the `to_date` function here? The logic is already there for this. Or probably better here is the timestamp formatting.

DanCodedThis (approved):
@osipovartem LGTM, up to you with what I wrote before

osipovartem:
> @osipovartem Can you call the `to_date` function here? The logic is already there for this. Or probably better here is the timestamp formatting.

to_char returns Utf8

DanCodedThis:
> > @osipovartem Can you call the `to_date` function here? The logic is already there for this. Or probably better here is the timestamp formatting.
> 
> to_char returns Utf8

@osipovartem I meant like inside the `to_char(col::Date, "format") -> to_char(to_date(col, "format"))`. We already have a rule for cast to call the `::DATE -> to_date()`, it's just that you can't really pass down the format. You might be able to if you rewrite the rule to take the next argument also and pass it down to the `to_date()`. And then whatever date you have will be passed down from the `to_date()` as `yyyy-mm-dd` to the `to_char`.

It might be overkill, up to you.

DanCodedThis:
> > > @osipovartem Can you call the `to_date` function here? The logic is already there for this. Or probably better here is the timestamp formatting.
> > 
> > 
> > to_char returns Utf8
> 
> @osipovartem I meant like inside the `to_char(col::Date, "format") -> to_char(to_date(col, "format"))`. We already have a rule for cast to call the `::DATE -> to_date()`, it's just that you can't really pass down the format. You might be able to if you rewrite the rule to take the next argument also and pass it down to the `to_date()`. And then whatever date you have will be passed down from the `to_date()` as `yyyy-mm-dd` to the `to_char`.
> 
> It might be overkill, up to you.

Also, it might need to include the formatting for timestamp for `HH24` since dates don't have hours, but if I remember correctly, `DateTime` (which is what `to_date()` internally calls) eats the string up to the format and discards the rest. Not sure.

DanCodedThis:
@osipovartem Disregard what I said, it supposed to be the output format, not the input one.

PR #1515, opened 2025-08-12
Change test with `COLLATE`

Vedin:
We don't plan to support collations and charsets soon. So, there's no point in keeping this in the test suite for now. 
This pr changes such tests. (Except for collate and collation themselves, obviously). 
Additionally, updating the docs to highlight this.


DanCodedThis (approved):
@Vedin LGTM

DanCodedThis (approved):

PR #1526, opened 2025-08-14
Disable json get visitor to use semi-structured functionality

osipovartem:
This PR disables visitor to convert json data access to json_get function call. 
Since we already implemented basic semi-strucured fucntions we should use them.

NOTE: This increased 
dbt-gitlab from 65% to 78.2%


Vedin (approved):
Nice!

PR #1538, opened 2025-08-15
Add support for COPY INTO without volume

### LINKED ISSUE #1522: Allow COPY INTO without STORAGE_INTEGRATION for public endpoints/buckets

JanKaul:
## Summary
- Enable COPY INTO functionality without requiring a volume/storage integration
- Add automatic object store creation from URL scheme (S3 and local file system)
- Support for endpoint parameter in stage configuration
- Add comprehensive test coverage for the new functionality

## Changes
- Modified `get_object_store` method to handle cases without volume/credentials
- Added support for S3 URLs with automatic region resolution
- Added support for local file system URLs
- Enhanced error handling with new `UnsupportedUrlScheme` error
- Added test case `copy_into_without_volume` with snapshot verification

## Test plan
- [x] Added new test case that verifies COPY INTO works without volume
- [x] Test loads data from S3 bucket and verifies correct sum calculation
- [x] Snapshot test ensures output consistency
- [x] Existing tests continue to pass

Closes #1522

### DIFF
--- a/crates/core-executor/src/query.rs
+++ b/crates/core-executor/src/query.rs
@@ -2502,78 +2502,65 @@ impl UserQuery {
         stage_params: StageParamsObject,
         url: &ListingTableUrl,
     ) -> Result<Arc<dyn ObjectStore + 'static>> {
-        // First, try to get object store from stage parameters
-        let object_store = match (stage_params.storage_integration, stage_params.credentials) {
+        match (&stage_params.storage_integration, &stage_params.credentials) {
             (Some(volume), _) => {
                 let volume = self
                     .metastore
-                    .get_volume(&volume)
+                    .get_volume(volume)
                     .await
                     .context(ex_error::MetastoreSnafu)?
                     .context(ex_error::VolumeNotFoundSnafu { volume })?;
-                Some(
-                    volume
-                        .get_object_store()
-                        .context(ex_error::MetastoreSnafu)?,
-                )
+                Ok(volume
+                    .get_object_store()
+                    .context(ex_error::MetastoreSnafu)?)
             }
-            (None, credentials) => {
-                if credentials.options.is_empty() {
-                    None
-                } else {
-                    // Create object store from credentials
-                    let access_key_id = get_kv_option(&credentials, "AWS_KEY_ID");
-                    let secret_access_key = get_kv_option(&credentials, "AWS_SECRET_KEY");
-                    let session_token = get_kv_option(&credentials, "AWS_SESSION_TOKEN");
+            (None, credentials) if !credentials.options.is_empty() => {
+                // Create object store from credentials
+                let access_key_id = get_kv_option(credentials, "AWS_KEY_ID");
+                let secret_access_key = get_kv_option(credentials, "AWS_SECRET_KEY");
+                let session_token = get_kv_option(credentials, "AWS_SESSION_TOKEN");
 
-                    if let (Some(access_key), Some(secret_key)) = (access_key_id, secret_access_key)
-                    {
-                        let object_store_url = url.object_store();
-                        let bucket = object_store_url
-                            .as_str()
-                            .trim_start_matches("s3://")
-                            .trim_end_matches('/');
-
-                        let store = create_s3_object_store(
-                            bucket,
-                            stage_params.endpoint.clone(),
-                            Some(access_key),
-                            Some(secret_key),
-                            session_token,
-                        )
-                        .await?;
-                        Some(store)
-                    } else {
-                        None
-                    }
-                }
-            }
-        };
-
-        // If no object store from stage params, create one from URL
-        if let Some(store) = object_store {
-            Ok(store)
-        } else {
-            match url.scheme() {
-                "s3" => {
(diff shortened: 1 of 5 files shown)


crates/core-executor/src/query.rs
```diff
@@ -2455,15 +2457,65 @@ impl UserQuery {
...
+                        .context(ex_error::MetastoreSnafu)?,
+                )
             }
-            (None, _) => {
-                todo!()
+            (None, credentials) => {
+                if credentials.options.is_empty() {
+                    None
+                } else {
+                    todo!()
```
DanCodedThis: what is this `todo()` case for?

crates/core-executor/src/query.rs
```diff
@@ -2455,15 +2457,65 @@ impl UserQuery {
...
+                    } else {
+                        builder
+                    };
+
+                    let s3 = builder.build().map_err(|_| {
+                        ex_error::InvalidBucketIdentifierSnafu {
+                            ident: bucket.to_string(),
+                        }
+                        .build()
+                    })?;
```
DanCodedThis: We are avoiding `map_err()` even without using the actual value. You could do something like:
```rust
let Ok(s3) = builder.build() else {
                        return ex_error::InvalidBucketIdentifierSnafu {
                            ident: bucket.to_string(),
                        }
                        .fail()?;
                    });
```

DanCodedThis (approved):
@JanKaul LGTM

crates/core-executor/src/query.rs
JanKaul: You can hardcode your credentials in the COPY INTO statement. [Link](https://docs.snowflake.com/en/sql-reference/sql/copy-into-table#additional-cloud-provider-parameters)

rampage644: Let's replace with an actual error - otherwise the thread panics and is stuck in endless loop

rampage644 (approved):

crates/core-executor/src/query.rs
JanKaul: Shall I just go ahead and implement it?

crates/core-executor/src/query.rs
DanCodedThis: @JanKaul or even this:
```rust
let s3 = builder.build().context(
                        ex_error::InvalidBucketIdentifierSnafu {
                            ident: bucket.to_string(),
                        }
                 )?;
```

crates/core-executor/src/query.rs
```diff
@@ -2454,15 +2456,65 @@ impl UserQuery {
...
+                    } else {
+                        builder
+                    };
+
+                    let Ok(s3) = builder.build() else {
+                        return ex_error::InvalidBucketIdentifierSnafu {
+                            ident: bucket.to_string(),
+                        }
+                        .fail();
+                    };
```
DanCodedThis: @JanKaul My mistake, not seeing it the first time, up to you if you want to change (since both are adequate):
```rust
let s3 = builder.build().context(
                        ex_error::InvalidBucketIdentifierSnafu {
                            ident: bucket.to_string(),
                        }
                    )?;
```

DanCodedThis: I think the first thing I wrote could be reserved for `Ok(Some(...))` if `Result<Option<...>>`.

Or just a `match` if you care about `Ok(None)`.

crates/core-executor/src/query.rs
```diff
@@ -3053,6 +3110,48 @@ pub fn get_kv_option<'a>(options: &'a KeyValueOptions, key: &str) -> Option<&'a
         .map(|opt| opt.value.as_str())
 }
 
+async fn create_s3_object_store(
```
osipovartem: Does it make sence to replace this logic by
S3Volume{....}.get_s3_builder()?
I also think this logic should be outside of query.rs somewhere.

crates/core-executor/src/query.rs
```diff
@@ -3053,6 +3110,48 @@ pub fn get_kv_option<'a>(options: &'a KeyValueOptions, key: &str) -> Option<&'a
...
+    if let Some(endpoint) = endpoint {
+        builder = builder.with_endpoint(endpoint);
+    }
+
+    let Ok(s3) = builder.build() else {
+        return ex_error::InvalidBucketIdentifierSnafu {
+            ident: bucket.to_string(),
+        }
+        .fail();
+    };
```
DanCodedThis: super nitpick (my bad on giving the wrong example):
```rust
let s3 = builder.build().context(ex_error::InvalidBucketIdentifierSnafu {
            ident: bucket.to_string(),
        })?;
```

DanCodedThis (approved):
@JanKaul LGTM

crates/core-executor/src/query.rs
JanKaul: Could you have, another look. I moved it to metastore volumes.

DanCodedThis (approved):
@JanKaul LGTM

osipovartem (approved):

PR #1540, opened 2025-08-15
Add query id to error, fix errors for create, drop, alter table

### LINKED ISSUE #1322: SQL: Bad errors in table ddl sql statements
When issuing ddl statements on a table in non-existing schema, using snowsql

## In Embucket snowsql returns following for statements:

`CREATE TABLE foo.hello(amount number)`
000200: 200: Iceberg error: Schema foo not found in database embucket

`ALTER TABLE foo.hello add column c5 VARCHAR`
No data

`DROP TABLE foo.hello`
No data

## In Snowlake snowsql return kind of the same error for that statements:
All three queries return the same:

002003 (02000): 01bdae95-0003-2f29-0004-d66e02ccc126: SQL compilation error:
Schema 'TESTS.FOO' does not exist or not authorized.

## What is expected on Embucket
* Error message should be the same or at least be reasonable
* For further debugging "id of failed query should be included as well"
* Query id should be in logs too

YaroslavLitvinov:
closes #1322

### :memo: What is included in this PR:

* Add `query_id` to the execution error, logs, tracing, etc;
* Snowflake like errors for `create table`, `drop table` statements, in case if schema doesn't exist.
* Unsupported feature error for `alter table`


rampage644 (approved):

crates/api-ui/src/error.rs
```diff
@@ -136,19 +137,48 @@ impl IntoResponse for Error {
...
+            Self::QueriesError { source, .. } => match &**source {
+                crate::queries::Error::Query {
+                    source: crate::queries::error::QueryError::Execution { source, .. },
+                    ..
+                } => source.query_id(),
+                _ => String::new(),
+            },
+            _ => String::new(),
+        }
+    }
```
DanCodedThis, line 151: You should be able to (if the &** is not the issue):
```rust
 match self {
            Self::QueriesError { source: crate::queries::Error::Query {
                    source: crate::queries::error::QueryError::Execution { source, .. },
                    ..
                }} => source.query_id(),
            _ => String::new(),
        }
```

crates/api-ui/src/error.rs
```diff
@@ -136,19 +137,48 @@ impl IntoResponse for Error {
...
+                } => format!(
+                    "{}: {}",
+                    source.query_id(),
+                    source.to_snowflake_error().display_error_message()
+                ),
+                _ => error_str,
+            },
+            _ => error_str,
+        }
+    }
```
DanCodedThis, line 170: Also here:
```rust
match self {
            Self::QueriesError { source: crate::queries::Error::Query {
                    source: crate::queries::error::QueryError::Execution { source, .. },
                    ..
                } => format!(
                    "{}: {}",
                    source.query_id(),
                    source.to_snowflake_error().display_error_message()
                ),   
            }
             _ => error_str,
        }
```

crates/core-executor/src/query.rs
```diff
@@ -381,8 +383,14 @@ impl UserQuery {
...
                     return Box::pin(self.copy_into_snowflake_query(*s)).await;
                 }
-                Statement::AlterTable { .. }
-                | Statement::StartTransaction { .. }
+                Statement::AlterTable { .. } => {
+                    NotSupportedStatementSnafu {
+                        statement: "ALTER TABLE".to_string(),
+                    }
+                    .fail()?;
+                    return self.status_response();
```
DanCodedThis: Why return the `self.status_responce`, if the error is guaranteed to bubble up from there? You could probably return the error if the compiler is asking for a return there.

crates/core-executor/src/query.rs
```diff
@@ -606,6 +616,17 @@ impl UserQuery {
...
                     .await?;
+                } else {
+                    // Iceberg error doesn't containt enough information, raise schema error
+                    let _ = catalog.schema(&schema_name).ok_or_else(|| {
+                        ex_error::SchemaNotFoundInDatabaseSnafu {
+                            schema: schema_name,
+                            db: catalog_name.to_string(),
+                        }
+                        .build()
+                    })?;
```
DanCodedThis: Why not `.context()`? Is this is an option `use snafu::OptExt`:
```rust
let _ = catalog.schema(&schema_name).context(
                        ex_error::SchemaNotFoundInDatabaseSnafu {
                            schema: schema_name,
                            db: catalog_name.to_string(),
                        }
                    )?;
```

crates/core-executor/src/query.rs
```diff
@@ -710,8 +732,9 @@ impl UserQuery {
             let target_table = catalog
                 .schema(schema_name)
                 .ok_or_else(|| {
-                    ex_error::SchemaNotFoundSnafu {
+                    ex_error::SchemaNotFoundInDatabaseSnafu {
                         schema: schema_name.to_string(),
+                        db: &catalog_name,
                     }
                     .build()
                 })?
```
DanCodedThis: Here also `.context()` (and lower also you can change it). If option `use snafu::OptExt`:
```rust
let Ok(Some(target_table)) = catalog
            .schema(schema_name)
            .context(
                ex_error::SchemaNotFoundSnafu {
                    ex_error::SchemaNotFoundInDatabaseSnafu {
                    schema: schema_name.to_string(),
                    db: &catalog_name,
                }
            )?
            .table(name.table())
            .await
            else {
                return ex_error::TableProviderNotFoundSnafu {
                    table_name: name.table().to_string(),
                }.fail()?; //could be `build`, not `fail`
            };
```

crates/df-catalog/src/catalogs/embucket/catalog.rs
```diff
@@ -75,7 +75,7 @@ impl CatalogProvider for EmbucketCatalog {
                 .get_schema(&SchemaIdent::new(database.clone(), schema_name.clone()))
                 .await
             {
-                Ok(_) => {
+                Ok(Some(_)) => {
```
DanCodedThis: I know you didn't write this, but there is no need for `.unwrap_or(None)` since we already return an option. This probably can be:
 ```rust
 metastore
                .get_schema(&SchemaIdent::new(database.clone(), schema_name.clone()))
                .await.map(|Some(_)| {
let schema = EmbucketSchema {
                        database,
                        schema: schema_name,
                        metastore,
                        iceberg_catalog,
                    };
                    Arc::new(schema) as Arc<dyn SchemaProvider>
})
```

**Review by DanCodedThis — APPROVED** [review]
@YaroslavLitvinov LGTM, some nitpicks with errors

File: crates/api-ui/src/error.rs (hunk 1 above)
Comment (YaroslavLitvinov) [reply crates/api-ui/src/error.rs:151]: @DanCodedThis Example you provided just won't work with Boxes
FYI
```
    QueriesError {
        source: Box<crate::queries::Error>,
    },
```

File: crates/api-ui/src/error.rs (hunk 2 above)
Comment (YaroslavLitvinov) [reply crates/api-ui/src/error.rs:170]: box also used here

**Review by rampage644 — APPROVED** [review]

File: crates/core-executor/src/query.rs [hunk 7]
```diff
@@ -380,8 +382,11 @@ impl UserQuery {
                 Statement::CopyIntoSnowflake { .. } => {
                     return Box::pin(self.copy_into_snowflake_query(*s)).await;
                 }
-                Statement::AlterTable { .. }
-                | Statement::StartTransaction { .. }
+                Statement::AlterTable { .. } => NotSupportedStatementSnafu {
+                    statement: "ALTER TABLE".to_string(),
```
osipovartem, line 386: This is a breaking change

crates/core-executor/src/query.rs
```diff
@@ -605,6 +612,16 @@ impl UserQuery {
                         table: ident.name().to_string(),
                     }))
                     .await?;
+                } else {
```
osipovartem, line 615: This is a breaking change

crates/core-executor/src/query.rs
```diff
@@ -605,6 +612,16 @@ impl UserQuery {
...
+                    // Iceberg error doesn't containt enough information, raise schema error
+                    catalog.schema(&schema_name).context(
```
osipovartem, line 617: Table drop not related to schema

PR #1541, opened 2025-08-15
Allow timestamps with timezone for snowflake connector

### LINKED ISSUE #1539: Fix snowflake connector timestamp with timezone parsing
```cpp
ThreeFieldTimeStampTZConverter::ThreeFieldTimeStampTZConverter(
    ArrowArrayView* array, ArrowSchemaView* schema, int32_t scale,
    PyObject* context)
    : TimeStampBaseConverter(context, scale), m_array(array) {
  if (schema->schema->n_children != 3) {
    std::string errorInfo = Logger::formatString(
        "[Snowflake Exception] arrow schema field number does not match, "
        "expected 3 but got %d instead",
        schema->schema->n_children);
    logger->error(__FILE__, __func__, __LINE__, errorInfo.c_str());
    PyErr_SetString(PyExc_Exception, errorInfo.c_str());
    return;
  }
  for (int i = 0; i < schema->schema->n_children; i += 1) {
    ArrowSchema* c_schema = schema->schema->children[i];
    if (std::strcmp(c_schema->name, internal::FIELD_NAME_EPOCH.c_str()) == 0) {
      m_epoch = m_array->children[i];
    } else if (std::strcmp(c_schema->name,
                           internal::FIELD_NAME_TIME_ZONE.c_str()) == 0) {
      m_timezone = m_array->children[i];
    } else if (std::strcmp(c_schema->name,
                           internal::FIELD_NAME_FRACTION.c_str()) == 0) {
      m_fraction = m_array->children[i];
    } else {
      // do nothing
    }
  }
}```
To support correct snowflake display for timestamps with timezone we need to change returning type for ARROW

### PR DESCRIPTION
**Pull request by osipovartem**
Closes #1539 

- Added snowflake timestamp with representation as a struct with epoch, fraction and timezone
- Added default timestamp formatter for UI for the structure above
- Fixed to_timestamp with decimals args

### CODE REVIEW

File: crates/core-executor/src/utils.rs [hunk 1]
```diff
@@ -370,8 +358,170 @@ fn convert_timestamp_to_struct(
...
         }
+
+        let epoch = epochs.value(i);
+        let fraction = i64::from(fractions.value(i));
+        let dt = DateTime::from_timestamp_nanos(epoch * 1_000_000_000 + fraction);
+        let dt_str = if let Some(tz) = tz_col {
+            let offset_minutes = tz.value(i);
+            let offset = FixedOffset::east_opt((offset_minutes - 1440) * 60)
+                .or_else(|| FixedOffset::east_opt(0)) // 0 fallback for invalid offsets
+                .ok_or_else(|| CantCastToSnafu { v: "fixed_offset" }.build())?;
```
rampage644: @DanCodedThis Is it a good place to switch to `map_err` here?

osipovartem: We have option there, not  error

crates/core-executor/src/utils.rs
```diff
@@ -370,8 +358,170 @@ fn convert_timestamp_to_struct(
...
+
+            let (field, column) = if let DataType::Struct(fields_in_struct) = field.data_type() {
+                let has_epoch = fields_in_struct.iter().any(|f| f.name() == "epoch");
+                let has_fraction = fields_in_struct.iter().any(|f| f.name() == "fraction");
+
+                if has_epoch && has_fraction {
+                    let struct_array = column
+                        .as_any()
+                        .downcast_ref::<StructArray>()
+                        .ok_or_else(|| CantCastToSnafu { v: "struct_array" }.build())?;
```
rampage644: Can you share how error stack trace looks like in case we hit this error? I have a suspicion we can do better to provide more context about what happened.

crates/core-executor/src/utils.rs
DanCodedThis: @osipovartem why not just `.context()` with `use snafu::OptExt`?
```rust 
let offset = FixedOffset::east_opt((offset_minutes - 1440) * 60)
                .or_else(|| FixedOffset::east_opt(0)) // 0 fallback for invalid offsets
                .context(CantCastToSnafu { v: "fixed_offset" })?;
```
 You could also:
```rust  
let Some(offset) = FixedOffset::east_opt((offset_minutes - 1440) * 60)
                .or_else(|| FixedOffset::east_opt(0)) else {
               return  CantCastToSnafu { v: "fixed_offset" }.fail()?; //maybe `build`, not `fail`
                };
```

@rampage644 If you meant `.map_or_else()` it's probably not a good fit here, since the output is sequential, if not the real value -> then 0 -> but both of those are options so we need to get them out. You could use it, but it would look uglier than `.context()` and would only work with the last option. You probably could write it such that the both the 0 and the real value and the `err()` variant could be produced, but it would be two `map_or_else()`.

DanCodedThis: @osipovartem @rampage644 Not two, but probably uglier, it's harder to read I think:
```rust
let offset = FixedOffset::east_opt((offset_minutes - 1440) * 60)
                .map_or_else(
                    || FixedOffset::east_opt(0).context(CantCastToSnafu { v: "fixed_offset" }),
                    Ok,
                )?;
```

crates/core-executor/src/utils.rs
osipovartem: This code is only used when we showing result on UI
Also seems downcasting error is not possible here since we check if the Datatype is Struct above

crates/core-executor/src/utils.rs
```diff
@@ -375,6 +363,174 @@ fn convert_timestamp_to_struct(
...
+            TimeUnit::Millisecond => {
+                downcast_and_iter!(column, TimestampMillisecondArray).collect()
+            }
+            TimeUnit::Microsecond => {
+                downcast_and_iter!(column, TimestampMicrosecondArray).collect()
+            }
+            TimeUnit::Nanosecond => downcast_and_iter!(column, TimestampNanosecondArray).collect(),
+        };
+        return Arc::new(Int64Array::from(timestamps)) as ArrayRef;
+    };
```
DanCodedThis: super nitpick:
```rust
let DataType::Timestamp(_, Some(tz)) = column.data_type() else {
        let timestamps: Vec<_> = match unit {
            TimeUnit::Second => downcast_and_iter!(column, TimestampSecondArray).collect(),
            TimeUnit::Millisecond => {
                downcast_and_iter!(column, TimestampMillisecondArray).collect()
            }
            TimeUnit::Microsecond => {
                downcast_and_iter!(column, TimestampMicrosecondArray).collect()
            }
            TimeUnit::Nanosecond => downcast_and_iter!(column, TimestampNanosecondArray).collect(),
        };
        return Arc::new(Int64Array::from(timestamps)) as ArrayRef;
    };
```

crates/core-executor/src/utils.rs
```diff
@@ -375,6 +363,174 @@ fn convert_timestamp_to_struct(
...
+    tz_str
+        .parse::<Tz>()
+        .map(|tz| {
+            tz.offset_from_utc_datetime(&dt.naive_utc())
+                .fix()
+                .local_minus_utc()
+                / 60
+                + 1440
+        })
+        .unwrap_or(1440)
```
DanCodedThis: not sure it would be better here, up to you: `map().unwrap_or()` could be `.map_or_else()`

crates/embucket-functions/src/conversion/to_timestamp.rs
```diff
@@ -522,7 +523,7 @@ fn parse_decimal(
     timezone: Option<Arc<str>>,
     s: i8,
 ) -> DFResult<ColumnarValue> {
-    let s = i128::from(s).pow(10);
+    let s = 10_i128.pow(s as u32);
```
DanCodedThis, line 526: Isn't it inverted now? (s as i128 )^10 and now it's (10 as i128)^s?

DanCodedThis (approved):
@osipovartem LGTM, some not important nitpicks and a question

crates/embucket-functions/src/conversion/to_timestamp.rs
osipovartem: yes, it was a bug
scale means 10^scale 
 not scale^10

DanCodedThis (approved):
@osipovartem LGTM

PR #1542, opened 2025-08-15
`RANDSTR` Impl

### LINKED ISSUE #1528: Implement `randstr` function
https://docs.snowflake.com/en/sql-reference/functions/randstr

Vedin:
Closes #1528


osipovartem (approved):

crates/embucket-functions/src/string-binary/randstr.rs
```diff
@@ -0,0 +1,217 @@
...
+            let len_val = lengths.value(i);
+            if len_val < 0 {
+                InvalidParameterValueSnafu {
+                    value: len_val,
+                    reason: "length must not be negative".to_string(),
+                }
+                .fail()?;
+            }
+
+            let target_len: usize = len_val.try_into().unwrap();
```
DanCodedThis: Do we know for sure that `.unwrap()` will never lead to a panic? If not, use `.context()`, if it is an option `use snafu::OptExt` to allow for `.context()` on options.

crates/embucket-functions/src/string-binary/randstr.rs
```diff
@@ -0,0 +1,217 @@
...
+    match cast(arr, &DataType::Int64) {
+        Ok(v) => Ok(v),
+        Err(_) => match arr.data_type() {
+            DataType::Utf8 => {
+                let s = as_string_array(arr)?;
+                if s.len() > 0 && !s.is_null(0) {
+                    NumericValueNotRecognizedSnafu { value: s.value(0).to_string() }.fail()?;
+                }
+                NumericValueNotRecognizedSnafu { value: "".to_string() }.fail()?;
+                unreachable!()
```
DanCodedThis:
```rust
DataType::Utf8 => {
                let s = as_string_array(arr)?;
                NumericValueNotRecognizedSnafu { value: if s.len() > 0 && !s.is_null(0) { s.value(0).to_string() } else { "".to_string() } }.fail()?
            } 
```

Others by the same example. Also you can make the `if ... else ...` outside the error.

crates/embucket-functions/src/string-binary/randstr.rs
```diff
@@ -0,0 +1,217 @@
...
+        let ScalarFunctionArgs { args, .. } = args;
+
+        if args.len() != 2 {
+            InvalidArgumentCountSnafu {
+                function_name: "RANDSTR".to_string(),
+                expected: "2".to_string(),
+                actual: args.len(),
+            }
+            .fail()?;
+        }
```
DanCodedThis: Since you already checked it at the `coerce_types` and it ran before this, you don't need to check it again. 

Also, I would add a `return` before the error in the `coerce_types`.

crates/embucket-functions/src/string-binary/randstr.rs
```diff
@@ -0,0 +1,217 @@
...
+            }
+
+            let target_len: usize = len_val.try_into().unwrap();
+            let mut rng = StdRng::seed_from_u64(seeds.value(i) as u64);
+            let s = Alphanumeric.sample_string(&mut rng, target_len);
+            builder.append_value(&s);
+        }
+
+        let res: ArrayRef = Arc::new(builder.finish());
+        Ok(ColumnarValue::Array(res))
```
DanCodedThis: super nitpick:
```rust
Ok(ColumnarValue::Array(builder.finish()))
```

DanCodedThis (approved):
@Vedin LGTM

crates/embucket-functions/src/string-binary/randstr.rs
Vedin: done

crates/embucket-functions/src/string-binary/randstr.rs
Vedin: sure, missed this unwrap

rampage644 (approved):

PR #1548, opened 2025-08-17
DBT github action job timeout

osmazur:
Added job timeout to prevent running for 6 hours straight.


rampage644:
Duplicate of #1546?

PR #1554, opened 2025-08-18
Fix drop table if exists

### LINKED ISSUE #1553: Fix drop table if exists
https://github.com/Embucket/embucket/pull/1540/files intoduced issue whenwith table exists flag.
cc @YaroslavLitvinov

osipovartem:
https://github.com/Embucket/embucket/pull/1540/files intoduced issue when drop table query contains **if_exists** flag

We don't need to return error if drop table query contains **if_exists** flag

- If the flag and table are missing
  - check if schema exists, if not return error (e.g., **Iceberg: Namespace gitlab_dotcom22 not found.**)
  - return original Iceberg error (e.g., **Iceberg: Table gitlab_dotcom_project_repositories_dedupe_source__dbt_tmp not found.**)

Also enable dbt gitlab volume, db, schema creation.

Closes #1553

### DIFF
--- a/crates/core-executor/src/query.rs
+++ b/crates/core-executor/src/query.rs
@@ -548,6 +548,7 @@ impl UserQuery {
             object_type,
             names,
             cascade,
+            if_exists,
             ..
         } = statement.clone()
         else {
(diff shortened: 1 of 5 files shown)


Vedin (approved):

crates/core-executor/src/query.rs
```diff
@@ -612,16 +613,14 @@ impl UserQuery {
...
-                            schema: schema_name,
-                            db: catalog_name.to_string(),
-                        },
-                    )?;
-                    // return original error, since schema is exists
-                    table.context(ex_error::IcebergSnafu)?;
+                } else if !if_exists {
+                    // Check if the schema exists first
+                    iceberg_catalog
+                        .load_namespace(ident.namespace())
```
YaroslavLitvinov, line 619: Actually, downgrading context is not good idea, as lack of context makes impossible returning meaningful errors.

osipovartem: But we don't downgrading here, just pass as is

PR #1555, opened 2025-08-18
`to_object` scalar function

ravlio:


crates/embucket-functions/src/conversion/to_object.rs
```diff
@@ -0,0 +1,150 @@
...
+        &self.signature
+    }
+
+    fn return_type(&self, _arg_types: &[DataType]) -> DFResult<DataType> {
+        Ok(DataType::Utf8)
+    }
+
+    fn invoke_with_args(&self, args: ScalarFunctionArgs) -> DFResult<ColumnarValue> {
+        let ScalarFunctionArgs { args, .. } = args;
+        let arr = match args[0].clone() {
```
osipovartem:
```rust
let ScalarFunctionArgs { args, number_rows, .. } = args;
let arr =  args[0].into_array(number_rows)
```

crates/embucket-functions/src/conversion/to_object.rs
```diff
@@ -0,0 +1,150 @@
...
+                        }
+                    } else {
+                        b.append_null();
+                    }
+                }
+            }
+            _ => return InvalidTypeSnafu.fail()?,
+        }
+
+        Ok(if arr.len() == 1 {
```
osipovartem: We don't need to return slacar function anywhere since datafusion doing this automaticaly for arrays with len = 1
We should always return just arrays

crates/embucket-functions/src/conversion/to_object.rs
```diff
@@ -0,0 +1,150 @@
...
+            DataType::Null => {
+                for _ in 0..arr.len() {
+                    b.append_null();
+                }
+            }
+            DataType::Utf8 | DataType::LargeUtf8 | DataType::Utf8View => {
+                let arr: &StringArray = as_generic_string_array(&arr)?;
+
+                for v in arr {
+                    if let Some(v) = v {
```
osipovartem, line 77: Maybe json::encode_array maybe usefull here

ravlio: I check validity of the value (if value is object, otherwise throw an error)

crates/embucket-functions/src/conversion/to_object.rs
```diff
@@ -0,0 +1,139 @@
...
+                            return InvalidTypeSnafu.fail()?;
+                        }
+                    } else {
+                        b.append_null();
+                    }
+                }
+
+                Arc::new(b.finish())
+            }
+            _ => return InvalidTypeSnafu.fail()?,
```
DanCodedThis, line 92: Perhaps, one of those errors should be different, or maybe both. Take a look at Snowflakes error text, if you can match it, that's great, if not, that's fine also.

DanCodedThis (approved):
@ravlio LGTM, minor erroring nitpick

crates/embucket-functions/src/conversion/to_object.rs
ravlio: `select TO_OBJECT( 1 );`

`Invalid type [TO_OBJECT(1)] for parameter 'TO_OBJECT'`
I don't know how to form `TO_OBJECT(1)`

DanCodedThis: It could be an expected or text or reason which is just a string. But again up to you.

DanCodedThis (approved):
reapprove

DanCodedThis:
Merged in #1587 as #1579

PR #1564, opened 2025-08-19
Add basic checks to alter table

### LINKED ISSUE #1563: AlterTable add basic checks for schema and table existence

osipovartem:
Closes #1563

Return a successful status response for `ALTER TABLE` when basic validation checks pass, and return an error response if any of the checks fail

Also fixed embucket catalog schema check. Metastore returns Ok(None) for missing schemas.

### DIFF
--- a/crates/core-executor/src/query.rs
+++ b/crates/core-executor/src/query.rs
@@ -6,7 +6,8 @@ use super::catalog::{
 };
 use super::datafusion::planner::ExtendedSqlToRel;
 use super::error::{
-    self as ex_error, Error, ObjectType as ExistingObjectType, RefreshCatalogListSnafu, Result,
+    self as ex_error, Error, InvalidColumnIdentifierSnafu, MergeSourceNotSupportedSnafu,
+    ObjectType as ExistingObjectType, RefreshCatalogListSnafu, Result,
 };
 use super::session::UserSession;
 use super::utils::{NormalizedIdent, is_logical_plan_effectively_empty};
@@ -15,9 +16,6 @@ use crate::datafusion::physical_plan::merge::{
     DATA_FILE_PATH_COLUMN, MANIFEST_FILE_PATH_COLUMN, SOURCE_EXISTS_COLUMN, TARGET_EXISTS_COLUMN,
 };
 use crate::datafusion::rewriters::session_context::SessionContextExprRewriter;
-use crate::error::{
-    InvalidColumnIdentifierSnafu, MergeSourceNotSupportedSnafu, NotSupportedStatementSnafu,
-};
 use crate::models::{QueryContext, QueryResult};
 use arrow_schema::SchemaBuilder;
 use core_history::HistoryStore;
@@ -99,10 +97,10 @@ use snafu::{OptionExt, ResultExt};
 use sqlparser::ast::helpers::key_value_options::KeyValueOptions;
 use sqlparser::ast::helpers::stmt_data_loading::StageParamsObject;
 use sqlparser::ast::{
-    AssignmentTarget, CloudProviderParams, MergeAction, MergeClause, MergeClauseKind,
-    MergeInsertKind, ObjectNamePart, ObjectType, OneOrManyWithParens, PivotValueSource,
-    ShowObjects, ShowStatementFilter, ShowStatementIn, TruncateTableTarget, Use, Value,
-    visit_relations_mut,
+    AlterTableOperation, AssignmentTarget, CloudProviderParams, MergeAction, MergeClause,
+    MergeClauseKind, MergeInsertKind, ObjectNamePart, ObjectType, OneOrManyWithParens,
+    PivotValueSource, ShowObjects, ShowStatementFilter, ShowStatementIn, TruncateTableTarget, Use,
+    Value, visit_relations_mut,
 };
 use std::collections::hash_map::Entry;
 use std::collections::{HashMap, HashSet};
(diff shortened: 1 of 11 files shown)


crates/core-executor/src/query.rs
```diff
@@ -541,6 +543,38 @@ impl UserQuery {
...
+        let resolved = self.resolve_table_ref(ident);
+        let catalog = self.get_catalog(&resolved.catalog)?;
+        let schema =
+            catalog
+                .schema(&resolved.schema)
+                .context(ex_error::SchemaNotFoundInDatabaseSnafu {
+                    schema: &resolved.schema.to_string(),
+                    db: &resolved.catalog.to_string(),
+                })?;
+        println!("schema: {schema:?}");
```
Vedin: remove this `println!`

crates/df-catalog/src/catalogs/embucket/catalog.rs
```diff
@@ -71,19 +71,22 @@ impl CatalogProvider for EmbucketCatalog {
...
+                .is_some()
+            {
+                let schema: Arc<dyn SchemaProvider> = Arc::new(EmbucketSchema {
+                    database,
+                    schema: schema_name,
+                    metastore,
+                    iceberg_catalog,
+                });
+                Some(schema)
+            } else {
```
Vedin: Why we have `None` in this else clause if we'd have it anyway on `unwrap_or(None)`?

osipovartem: block_in_new_runtime returns `Result<R>` so we need to check the result here
Also by some historical reason metastore.get_schema returns  Result<Option<RwObject<Schema>>>
And for missing schemas it returns **Ok(None)**

DanCodedThis: I saw this thing also, but in Yaroslav's pr (he didn't write the `.unwapr_or()`, etc), so what do you think about: 
```rust
metastore
               .get_schema(&SchemaIdent::new(database.clone(), schema_name.clone()))
               .await.and_then(|schema| {
                      schema.map(|_| {
                      let schema = EmbucketSchema {
                           database,
                           schema: schema_name,
                           metastore,
                           iceberg_catalog,
                       };
                      Arc::new(schema) as Arc<dyn SchemaProvider>
                      })             
              })

```
If you can write it similar to this, There is no need for `.unwrap_or(None)` since we already return an option (so any `.map()` should be good, but not `.and_then()`, although it can be. This probably can be `.map_or_else`, perhaps, because although it allows for pattern binding in capture, the compiler will be furious, since what about none variants of an option? Also, `let Ok(Some(...)) else { ... }` can work.

P.S. rewrote the code a bit

crates/core-executor/src/query.rs
```diff
@@ -541,6 +543,37 @@ impl UserQuery {
...
+                })?;
+        if !if_exists {
+            schema
+                .table(&resolved.table)
+                .await
+                .context(ex_error::DataFusionSnafu)?
+                .context(ex_error::TableNotFoundSnafu {
+                    table: &resolved.table.to_string(),
+                    schema: &resolved.schema.to_string(),
+                })?;
```
DanCodedThis, line 572: This should be better, like this (I thought about double `.context()`, it's good, but not great):
```rust
let Ok(Some(_) = schema.table(&resolved.table).await else {
    return ex_error::TableNotFoundSnafu {
                    table: &resolved.table.to_string(),
                    schema: &resolved.schema.to_string(),
                }).fail()?;
}
```

crates/df-catalog/src/catalogs/embucket/catalog.rs
osipovartem: block_in_new_runtime returns `Result<R>`, not Option
but for **fn schema(&self, name: &str) -> Option<Arc<dyn SchemaProvider>> we expect Option**

DanCodedThis (approved):
@osipovartem LGTM, nitpicks on code erroring structure

crates/core-executor/src/query.rs
osipovartem: @YaroslavLitvinov WDYT

crates/core-executor/src/query.rs
```diff
@@ -541,6 +543,37 @@ impl UserQuery {
...
+        let resolved = self.resolve_table_ref(ident);
+        let catalog = self.get_catalog(&resolved.catalog)?;
+        let schema =
+            catalog
+                .schema(&resolved.schema)
+                .context(ex_error::SchemaNotFoundInDatabaseSnafu {
+                    schema: &resolved.schema.to_string(),
+                    db: &resolved.catalog.to_string(),
+                })?;
+        if !if_exists {
```
YaroslavLitvinov, line 564: so that `if_exists` check means if there are no IF EXISTS in statement, it will check table existance, but if exists it will not check. that is expected behaviour?

crates/core-executor/src/query.rs
YaroslavLitvinov: From my perspective, it has conflicting changes among used errors I introduced in PR#1565, will require some changes anyway.

YaroslavLitvinov (approved):

crates/core-executor/src/query.rs
osipovartem: yes

DanCodedThis (approved):

PR #1565, opened 2025-08-19
Snowflake error regression tests for drop table, alter table

### LINKED ISSUE #1557: Snowflake error regression in `drop table`
### Regression error of recently added snowflake error for `drop table`

With error fix regression test should be added as well

YaroslavLitvinov:
closes #1557

Add complete set of snowflake error tests checking existence errors texts for drop table, alter table


crates/core-executor/src/error.rs
```diff
@@ -186,6 +186,15 @@ pub enum Error {
         location: Location,
     },
 
+    #[snafu(display("Table {table} not found in schema {schema} in database {db}"))]
+    TableNotFoundInSchemaInDatabase {
```
osipovartem, line 191: The error above is not used (TableNotFound)
Let's reuse/update/delete

crates/core-executor/src/query.rs
```diff
@@ -541,6 +551,60 @@ impl UserQuery {
...
     }
 
+    #[instrument(
+        name = "UserQuery::alter_table_query",
+        level = "trace",
+        skip(self),
+        err
+    )]
+    #[allow(clippy::too_many_lines)]
+    pub async fn alter_table_query(&self, statement: Statement) -> Result<QueryResult> {
```
osipovartem: check this PR https://github.com/Embucket/embucket/pull/1564
we don't need iceberg catalog here

crates/core-executor/src/error.rs
YaroslavLitvinov: This actually was used in drop table, now I added alter table as well

DanCodedThis (approved):
@YaroslavLitvinov LGTM, but do check iceberg catalog as @osipovartem said

crates/core-executor/src/query.rs
YaroslavLitvinov: @osipovartem  I wonder why iceberg catalog needed in drop table, but in alter table it is not needed

crates/core-executor/src/error.rs
```diff
@@ -187,6 +187,15 @@ pub enum Error {
         location: Location,
     },
 
+    #[snafu(display("Table {table} not found in schema {schema} in database {db}"))]
```
rampage644: `Table {table} not found in {db}.{schema}`

rampage644 (approved):

crates/core-executor/src/query.rs
```diff
@@ -566,9 +566,10 @@ impl UserQuery {
                 .table(&resolved.table)
                 .await
                 .context(ex_error::DataFusionSnafu)?
-                .context(ex_error::TableNotFoundSnafu {
+                .context(ex_error::TableNotFoundInSchemaInDatabaseSnafu {
                     table: &resolved.table.to_string(),
                     schema: &resolved.schema.to_string(),
+                    db: &resolved.catalog.to_string(),
                 })?;
         }
```
DanCodedThis, line 574: This here can be (since we don't really have a specific first `.context()` error):
```rust
let Ok(Some(_)) = schema
                .table(&resolved.table)
                .await else {
                    return ex_error::TableNotFoundInSchemaInDatabaseSnafu {
                        table: &resolved.table.to_string(),
                        schema: &resolved.schema.to_string(),
                        db: &resolved.catalog.to_string(),
                }.fail()?;
            };
```

Did I already suggest this (seems like a déjà vu)?

DanCodedThis (approved):
@YaroslavLitvinov LGTM

PR #1568, opened 2025-08-19
Apply analyzer rules when creating table

### LINKED ISSUE #1566: Enable plan optimization when creating a table
When we create a table we don't use optimizer rules
But in case we have a CTE, during insert we run optimizer rules that can change initial schema
```sql
create or replace transient table EMBUCKET.common_prep.prep_gitlab_dotcom_plan2 as
(WITH source AS ( SELECT  CURRENT_TIMESTAMP() AS dbt_updated_at, 1 as tt)
SELECT * FROM source);
```
That's why we have to use the same schema on each step to avoid schema error
```Error during planning: Inserting query must have the same schema with the table.```

osipovartem:
Closes #1566 
This will fix `Error during planning: Inserting query must have the same schema with the table.` when we trying to create and insert data within CTE

### DIFF
--- a/crates/core-executor/src/query.rs
+++ b/crates/core-executor/src/query.rs
@@ -683,9 +683,18 @@ impl UserQuery {
         create_table_statement.storage_serialization_policy = None;
         create_table_statement.cluster_by = None;
 
-        let plan = self
+        let mut plan = self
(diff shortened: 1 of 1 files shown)


DanCodedThis (approved):
@osipovartem LGTM

rampage644 (approved):

Issue #1504, opened 2025-08-11
engine: `'\\'` turn in to `''` when converting from AST to Logical plan


DanCodedThis:
- This was found as part of the `regexp_*` function family
- However, this affects all strings
- Parser is fine with `'\\'` but Logical plans seem to remove the `\\` to turn it to an empty string
- This also happens like this `\\b` -> `b`, etc (there are many other similar cases)
- The cure seems to be (a hacky way solution) is to replace `\\` with `\\\\` then when we get to the execution from the logical plan, we actually get `\\` which is what we want
- Perhaps there is a setting in DF ctx config?
- This can be fixed probably in the expr planner, etc

PR #1569, opened 2025-08-19
engine: Backslash docs

DanCodedThis:
Related to #1504 

- The problem is explained in docs and in the issue.
- There is a way to turn off the backslash escaping (inside the dialect of the parser) and then `\b` -> `\b`, closer to snowflake's `\\b` -> `\b`

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,19 +1,19 @@
-# Embucket Docs
+# Embucket docs
 
 ## About
 
-Astro Starlight frontend application for the Embucket docs.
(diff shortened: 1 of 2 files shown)


rampage644 (approved):
@DanCodedThis Rewrote the section with same prompts as for other documentation + fixed `vale` messages

DanCodedThis:
> @DanCodedThis Rewrote the section with same prompts as for other documentation + fixed `vale` messages

Great! Can't merge however: 
image

osipovartem (approved):

PR #1573, opened 2025-08-20
fix merging into empty table

### LINKED ISSUE #1571: dbt snowplow on-run-end hook fails
It fails with `External(Iceberg(InvalidFormat(\"Snapshot to overwrite\")))` error which seems to happen at https://github.com/JanKaul/iceberg-rust/blob/49302b6478006bcf3bb64ae1a18c101609fb6f89/iceberg-rust/src/table/transaction/operation.rs#L404

Error stack trace
```
8244371014595: Query execution error: DataFusion error: External error: Snapshot to overwrite doesn't have the right format

  0: <transparent>

  1: 8244371014595: Query execution error: DataFusion error: External error: Snapshot to overwrite doesn't have the right format, at crates/core-executor/src/service.rs:298:25

  2: DataFusion error: External error: Snapshot to overwrite doesn't have the right format, at crates/core-executor/src/query.rs:1976:22

  3: External(Iceberg(InvalidFormat("Snapshot to overwrite")))
```

JanKaul:
## Summary
- Fixed MERGE INTO operation when target table is empty
- Changed from using overwrite to append_data when no matching files exist
- Fixed stream polling to return data instead of pending when no matches found
- Added proper handling for empty strings in unique value collection

## Closes
Closes #1571

### DIFF
--- a/crates/core-executor/src/datafusion/physical_plan/merge.rs
+++ b/crates/core-executor/src/datafusion/physical_plan/merge.rs
@@ -181,14 +181,23 @@ impl ExecutionPlan for MergeIntoCOWSinkExec {
                     })?
                 };
 
-                if !datafiles.is_empty() && !matching_files.is_empty() {
+                if !datafiles.is_empty() {
                     // Commit transaction on Iceberg table
-                    table
-                        .new_transaction(branch.as_deref())
-                        .overwrite(datafiles, matching_files)
-                        .commit()
-                        .await
-                        .map_err(DataFusionIcebergError::from)?;
+                    if matching_files.is_empty() {
+                        table
+                            .new_transaction(branch.as_deref())
+                            .append_data(datafiles)
+                            .commit()
+                            .await
+                            .map_err(DataFusionIcebergError::from)?;
+                    } else {
+                        table
+                            .new_transaction(branch.as_deref())
+                            .overwrite(datafiles, matching_files)
+                            .commit()
+                            .await
+                            .map_err(DataFusionIcebergError::from)?;
+                    }
                 }
 
                 Ok(RecordBatch::new_empty(schema))
@@ -414,7 +423,7 @@ impl Stream for MergeCOWFilterStream {
                 }
 
                 if matching_data_and_manifest_files.is_empty() {
-                    Poll::Pending
+                    Poll::Ready(Some(Ok(batch)))
                 } else {
                     let file_predicate = matching_data_files
                         .iter()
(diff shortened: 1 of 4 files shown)


osipovartem (approved):

crates/core-executor/src/datafusion/physical_plan/merge.rs
```diff
@@ -181,14 +181,23 @@ impl ExecutionPlan for MergeIntoCOWSinkExec {
...
+                            .commit()
+                            .await
+                            .map_err(DataFusionIcebergError::from)?;
+                    } else {
+                        table
+                            .new_transaction(branch.as_deref())
+                            .overwrite(datafiles, matching_files)
+                            .commit()
+                            .await
+                            .map_err(DataFusionIcebergError::from)?;
```
DanCodedThis: I would suggest adding errors, using `.context()` if that's not too much hassle.

DanCodedThis (approved):
@JanKaul LGTM, erroring nitpick

crates/core-executor/src/datafusion/physical_plan/merge.rs
JanKaul: The function has to return a DatafusionError, hence I need something that implements `Into` DatafusionError.

rampage644 (approved):

rampage644 (approved):

crates/core-executor/src/datafusion/physical_plan/merge.rs
DanCodedThis: @JanKaul Something like this? Not sure if it's worth the hassle, but potential a good long-term decision:
```rust
impl From<Error> for datafusion_common::DataFusionError {
    fn from(value: Error) -> Self {
        Self::External(Box::new(crate::df_error::DFExternalError::Conversion {
            source: value,
        }))
    }
}
```

DanCodedThis: We've been doing it in **embucket-functions** crate for a while for our snafu errors. So when you call `.context(...)?` (notice the `?`) - it would turn itself into the `DatafusionError`.

DanCodedThis: However, there is also additional wrapping in to the type of the error type so Datafusion::Internal -> Error::Conversion type -> Specific Snafu Error -> Potential source error. You probably could avoid `Error::Conversion type` (e.g. for regexp, it's another variant struct with a snafu enum inside, so you probably don't need such modularity here).

JanKaul: Could you have another look?

DanCodedThis (approved):
@JanKaul LGTM

PR #1578, opened 2025-08-20
engine: `array_to_string` signature fix

### LINKED ISSUE #1577: engine: `array_to_string` signature
Dbt snowplow run:
```
13:04:38    Database Error in model snowplow_web_consent_events_this_run (models/optional_modules/consent/scratch/snowflake/snowplow_web_consent_events_this_run.sql)
  8244304927220: Error during planning: Function 'array_to_string' expects NativeType::String but received NativeType::List(LogicalField { name: "item", logical_type: LogicalType(Native(String), String), nullable: true }) No function matches the given name and argument types 'array_to_string(List(Field { name: "item", data_type: Utf8, nullable: true, dict_id: 0, dict_is_ordered: false, metadata: {} }), Utf8)'. You might need to add explicit type casts.
        Candidate functions:
        array_to_string(String(2))
```

DanCodedThis:
Closes #1577

Couldn't add a specail unit-test case for ` SELECT ARRAY_TO_STRING(['f', 'ff']::array, ',')`, since the parser is failing to work with `[]`, either we need to register the visitor or/and add the snowflake dialect

### DIFF
--- a/crates/core-executor/src/tests/query.rs
+++ b/crates/core-executor/src/tests/query.rs
@@ -873,44 +873,44 @@ test_query!(
 // Basic timezone conversion tests
 test_query!(
     convert_timezone_basic_utc_to_est,
-    r#"SELECT
+    r"SELECT
         '2024-01-15T12:00:00Z'::TIMESTAMP AS utc_time,
-        CONVERT_TIMEZONE('America/New_York', '2024-01-15T12:00:00Z'::TIMESTAMP) AS est_time"#,
+        CONVERT_TIMEZONE('America/New_York', '2024-01-15T12:00:00Z'::TIMESTAMP) AS est_time",
     snapshot_path = "convert_timezone"
 );
 
 test_query!(
     convert_timezone_basic_utc_to_pst,
-    r#"SELECT
+    r"SELECT
         '2024-01-15T12:00:00Z'::TIMESTAMP AS utc_time,
-        CONVERT_TIMEZONE('America/Los_Angeles', '2024-01-15T12:00:00Z'::TIMESTAMP) AS pst_time"#,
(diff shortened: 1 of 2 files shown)


crates/core-executor/src/tests/query.rs
```diff
@@ -873,44 +873,44 @@ test_query!(
 // Basic timezone conversion tests
 test_query!(
     convert_timezone_basic_utc_to_est,
-    r#"SELECT
+    r"SELECT
```
osipovartem, line 876:
```sql
SELECT ARRAY_TO_STRING(null::array, ',')
```
Is it possible to add a test like that?

DanCodedThis: I'll try, both in insta and in unit

DanCodedThis: in unit test i get this error: `Error: SQL(ParserError("Expected: <, found: , at Line: 1, Column: 35"), None)` for this code:
```rust
    #[tokio::test]
    async fn test_array_cast() -> DFResult<()> {
        let ctx = SessionContext::new();
        ctx.register_udf(ScalarUDF::from(ArrayToStringFunc::new()));
        let q = r"SELECT ARRAY_TO_STRING(null::array, ',') AS casted";
        let result = ctx.sql(q).await?.collect().await?;

        assert_batches_eq!(
            &[
                "+---------------------------------+",
                "| casted                          |",
                "+---------------------------------+",
                r"|                                |",
                "+---------------------------------+",
            ],
            &result
        );

        Ok(())
    }
```

rampage644 (approved):

crates/core-executor/src/tests/query.rs
DanCodedThis: Insta also get a similar error :(

PR #1592, opened 2025-08-22
Return correct error code on NotFound

rampage644:


crates/api-snowflake-rest/src/error.rs
```diff
@@ -121,35 +122,57 @@ impl IntoResponse for Error {
...
-            | Self::NotImplemented { .. } => http::StatusCode::OK,
+            | Self::NotImplemented { .. } => (http::StatusCode::OK, "02000", StatusCode::Other),
         };
 
         let display_error = self.display_error_message();
         // Give more context to user, not just "Internal server error"
         // if status_code == http::StatusCode::INTERNAL_SERVER_ERROR {
         //     display_error = "Internal server error".to_string();
-        // }
+        // }sno
```
osipovartem, line 171: Typo?

osipovartem (approved):

crates/api-snowflake-rest/src/error.rs
rampage644: Yeah

PR #1593, opened 2025-08-23
Fix aggregate functions on empty tables by preserving AggregateExec in optimizer

rampage644:
## Summary

Fixes a critical regression where aggregate functions like COUNT(*) on empty tables were returning empty result sets instead of the correct SQL-compliant results.

## Problem

The `RemoveExecAboveEmpty` physical optimizer rule introduced in commit 671d82ce was incorrectly replacing **all** execution plan nodes above empty tables with `EmptyExec`, including aggregate operations. This violated SQL standards where:

- `COUNT(*)` on empty tables should return `0`
- `SUM/AVG/MIN/MAX` on empty tables should return `NULL`

Instead, queries like `SELECT COUNT(*) FROM empty_table` were returning completely empty result sets.

## Root Cause

The original implementation in #756 included comprehensive tests for various execution plan types (Sort, Filter, Projection, Join, Limit) but **completely missed testing aggregate operations**. This oversight allowed the bug to slip through.

## Solution

Modified the `RemoveExecAboveEmpty` optimizer rule to exclude `AggregateExec` operations from being replaced with `EmptyExec`. Aggregate functions have special semantics and must be preserved even when their inputs are empty.

### Changes Made

1. **Fixed the optimizer rule** (`remove_exec_above_empty.rs`):
   - Added import for `AggregateExec`
   - Added logic to preserve aggregate operations when inputs are empty
   - Updated documentation to reflect the new behavior

2. **Added missing test coverage** (`count.rs`):
   - Added `COUNT(*)` test case that was missing from the original implementation
   - Includes proper snapshot testing for the expected output format

## Test Results

**Before fix:**
```
"++"
"++"
```

**After fix:**
```
"+----------+"
"| count(*) |"
"+----------+"
"| 0        |"
"+----------+"
```

## Impact

- Fixes COUNT(*) and other aggregate functions on empty tables
- Maintains performance optimizations for non-aggregate operations
- Adds test coverage to prevent future regressions
- No breaking changes to existing functionality

### DIFF
--- a/crates/core-executor/src/datafusion/physical_optimizer/remove_exec_above_empty.rs
+++ b/crates/core-executor/src/datafusion/physical_optimizer/remove_exec_above_empty.rs
@@ -52,8 +52,9 @@ impl PhysicalOptimizerRule for RemoveExecAboveEmpty {
         _config: &ConfigOptions,
     ) -> DFResult<Arc<dyn ExecutionPlan>> {
         plan.transform_up(|plan| {
-            // Skip EmptyExec itself
-            if plan.as_any().is::<EmptyExec>() {
+            // Skip EmptyExec itself and do not replace aggregate operations with EmptyExec because they should
+            // still return results even on empty tables (e.g., COUNT(*) should return 0)
+            if plan.as_any().is::<EmptyExec>() || contains_aggregates(&plan) {
                 return Ok(Transformed::no(plan));
             }
 
@@ -64,14 +65,7 @@ impl PhysicalOptimizerRule for RemoveExecAboveEmpty {
 
             // Replace the current node with EmptyExec if all inputs are EmptyExec
             let all_empty = inputs.iter().all(|input| input.as_any().is::<EmptyExec>());
-
             if all_empty {
-                // Do not replace aggregate operations with EmptyExec because they should
-                // still return results even on empty tables (e.g., COUNT(*) should return 0)
-                if plan.as_any().is::<AggregateExec>() {
-                    return Ok(Transformed::no(plan));
-                }
-
                 return Ok(Transformed::yes(Arc::new(EmptyExec::new(plan.schema()))));
             }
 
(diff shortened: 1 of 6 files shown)


crates/core-executor/src/datafusion/physical_optimizer/remove_exec_above_empty.rs
```diff
@@ -61,6 +66,12 @@ impl PhysicalOptimizerRule for RemoveExecAboveEmpty {
             let all_empty = inputs.iter().all(|input| input.as_any().is::<EmptyExec>());
 
             if all_empty {
+                // Do not replace aggregate operations with EmptyExec because they should
```
osipovartem: Move this part to the begining
```rust
// Skip EmptyExec itself and do not replace aggregate operations with EmptyExec because they should
// still return results even on empty tables (e.g., COUNT(*) should return 0)
if plan.as_any().is::<EmptyExec>() || plan.as_any().is::<AggregateExec>() {
    return Ok(Transformed::no(plan));
}
```

crates/core-executor/src/tests/sql/functions/aggregate/count.rs
```diff
@@ -0,0 +1,8 @@
+use crate::test_query;
+
+test_query!(
+    count,
+    "SELECT COUNT(*) FROM empty_table",
+    setup_queries = ["CREATE TABLE empty_table (id INT)"],
+    snapshot_path = "aggregate"
+);
```
osipovartem, line 8: Add also this test, to check other parent types (like ProjectionExec, etc.)
```rust
test_query!(
    count_cte,
    "WITH c as (SELECT COUNT(*) FROM empty_table) SELECT * FROM c",
    setup_queries = ["CREATE TABLE empty_table (id INT)"],
    snapshot_path = "aggregate"
);
```

DanCodedThis (approved):
@rampage644 @osipovartem LGTM

PR #1605, opened 2025-08-25
Fix timestamp out of range

### LINKED ISSUE #1597: Arrow error: Cast error: Overflow converting 9999-12-31 00:00:00 to Nanosecond

osipovartem:
- added support for timestamps with out of range values like '9999-12-31 00:00:02'' with donwcasting to Microseconds (same as Snowflake) 
- replaced string formats by regex with timezone support
- moved to_timestamp fucntions tests from core-executor to embucket-fucntions
Closes #1597


DanCodedThis:
@osipovartem I had similar error running snowplow web, but there were a `cast(9999-... as TIMESTAMP)`. Do we have a cast in the AST rewriters? If so or not, you can move/add this to the Logical Cast Analyzer just as date by calling the `ToTimestampFunc::new()` directly.

crates/embucket-functions/src/conversion/to_timestamp.rs
```diff
@@ -118,6 +103,16 @@ impl ToTimestampFunc {
...
+
+    fn unit_for_string_input(&self, v: &str) -> TimeUnit {
+        if try_parse_with_auto(v, self.try_mode, TimeUnit::Nanosecond).is_ok() {
+            TimeUnit::Nanosecond
+        } else if try_parse_with_auto(v, self.try_mode, TimeUnit::Microsecond).is_ok() {
+            TimeUnit::Microsecond
+        } else {
+            TimeUnit::Nanosecond
+        }
+    }
```
DanCodedThis, line 116: This could be:
```rust 
fn unit_for_string_input(&self, v: &str) -> TimeUnit {
        if try_parse_with_auto(v, self.try_mode, TimeUnit::Microsecond).is_ok() {
            TimeUnit::Microsecond
        } else {
            TimeUnit::Nanosecond
        }
    }
```

crates/embucket-functions/src/conversion/to_timestamp.rs
```diff
@@ -351,75 +349,141 @@ impl ScalarUDFImpl for ToTimestampFunc {
...
+                    arr,
+                    &format,
+                    self.timezone(),
+                    self.try_mode,
+                )?,
+            };
+            Ok(ColumnarValue::Array(Arc::new(array)))
+        } else {
+            InvalidDataTypeSnafu.fail()?
+        }
```
DanCodedThis, line 375: If that's not too much hassle this could be rewritten as a match, which would allow to use the `other` binding to pass the data type to either the error `UnsupportedInputTypeWithPositionSnafu`, or `UnsupportedInputTypeSnafu` instead of a less understandable `InvalidDataTypeSnafu`.

crates/embucket-functions/src/conversion/to_timestamp.rs
```diff
@@ -351,75 +349,141 @@ impl ScalarUDFImpl for ToTimestampFunc {
...
+    try_mode: bool,
+) -> DFResult<ArrayRef> {
+    let mut builder = T::new(arr.len(), tz.clone());
+    for v in arr {
+        match v {
+            None => T::append_null(&mut builder),
+            Some(s) if contains_only_digits(s) => {
+                let i = s
+                    .parse::<i64>()
+                    .map_err(|_| CantParseTimestampSnafu.build())?;
```
DanCodedThis: `.context()` can be used.

crates/embucket-functions/src/conversion/to_timestamp.rs
```diff
@@ -351,75 +349,141 @@ impl ScalarUDFImpl for ToTimestampFunc {
...
         } else {
-            InvalidDataTypeSnafu.fail()?
+            return None;
+        };
+        let (mut hour, mut minute, mut second, mut nano) = (0, 0, 0, 0);
+
+        if let Some(hour_match) = caps.get(4) {
+            hour = hour_match.as_str().parse().ok()?;
+            minute = caps[5].parse().ok()?;
+            second = caps.get(6).map_or(0, |m| m.as_str().parse().unwrap_or(0));
```
DanCodedThis, line 452: You can return a result if you want to use errors (both `map_or` or better `map_or_else`):
```rust
caps.get(6).map_or(Ok(0), |m| m.as_str().parse().context(...error...))?;
```

crates/embucket-functions/src/conversion/to_timestamp.rs
```diff
@@ -436,9 +500,14 @@ fn try_parse_with_format(s: &str, format: &str, try_mode: bool) -> DFResult<Opti
     }
 }
 
-fn apply_timezone(ts: i64, tz_str: &str, try_mode: bool) -> DFResult<i64> {
+fn apply_timezone(ts: i64, tz_str: &str, try_mode: bool, unit: TimeUnit) -> DFResult<i64> {
     let tz: Tz = tz_str.parse().map_err(|_| CantParseTimezoneSnafu.build())?;
-    let dt = DateTime::from_timestamp_nanos(ts);
+    let dt = match unit {
+        TimeUnit::Microsecond => {
+            DateTime::from_timestamp_micros(ts).ok_or_else(|| CantGetTimestampSnafu.build())?
```
DanCodedThis: `.context()`, if this is an option `use snafu::OptExt;`.

DanCodedThis (approved):
@osipovartem LGTM, comment about cast above & some minor nitpicks

osipovartem:
> @osipovartem I had similar error running snowplow web, but there were a `cast(9999-... as TIMESTAMP)`. Do we have a cast in the AST rewriters? If so or not, you can move/add this to the Logical Cast Analyzer just as date by calling the `ToTimestampFunc::new()` directly.

we already have such rewriter there

crates/embucket-functions/src/conversion/to_timestamp.rs
osipovartem: we don't need to return error in this case, we have only values captured by regex here
so only digits possible here

crates/embucket-functions/src/conversion/to_timestamp.rs
osipovartem: No, since we have to check wided type first of all, Nanoseconds have priority under Microsecond for better precision.
But try_parse_with_auto can also return error for Microseconds

crates/embucket-functions/src/conversion/to_timestamp.rs
osipovartem: I would keep this for the next refactoring to not add additional complication to the PR

DanCodedThis (approved):
@osipovartem LGTM

PR #1620, opened 2025-08-27
Support both 'epoch_second' and 'epoch_seconds' in date_part function rewriter

rampage644:
## Summary
Extended the DataFusion compatibility transformation to handle both singular and plural forms of epoch_seconds parameter in date_part function calls. This improves compatibility with different SQL dialects that may use either variant.


DanCodedThis (approved):
@rampage644 LGTM

PR #1621, opened 2025-08-27
Union schema analyzer rule with timestamp casting

### LINKED ISSUE #1608: Arrow error: Cast error: Overflow converting 9999-12-31 00:00:00 to Nanosecond for  UNION
Related to #1597 
```sql
    SELECT contract_effective_date FROM EMBUCKET.zuora_api_sandbox.zuora_api_sandbox_amendment_source
    UNION ALL
    SELECT '9999-12-31 00:00:00.000 +0000' AS contract_effective_date
```
Doesn't work since we don't call to_timestamp func here
contract_effective_date from source is Timestamp(Nanosecond), and arrow cannot cast '9999-12-31 00:00:00.000 +0000' to Timestamp(Nanosecond by himself

osipovartem:
Closes #1608

- Added to_timestamp call instead of timestamt Cast (this logic appears after type coercion) with specific check of return type to be sure that the schema is the same
- Added analyzer rule to check UNION ALL schema since casting logic can change the initial schema during physical optimization
- Fixed to_timestamp regex to correctly parse timezone in different formats
```rust
// This test ensures that when performing a UNION ALL with a constant timestamp literal
// that has out-of-range nanoseconds (e.g. `'9999-12-31 00:00:00.000 +0000'`),
// the analyzer forces the union schema to use `Timestamp(Microseconds, None)`.
// Without this normalization, schema reconciliation would fail because one side
// might use nanosecond precision while the other provides only microseconds and cannot be cast.
test_query!(
    union_with_timestamp_microseconds,
    "SELECT '2024-12-31 10:00:00.000'::TIMESTAMP as t
        UNION ALL
    SELECT '9999-12-31 00:00:00.000 +0000' AS t;",
    snapshot_path = "union_schema_cast_analyzer"
);
```


DanCodedThis:
@osipovartem Flaky snapshot?

osipovartem:
> @osipovartem Flaky snapshot?

Yeap, union all in random order, let me fix

DanCodedThis (approved):
@osipovartem LGTM

rampage644 (approved):

JanKaul (approved):

PR #1622, opened 2025-08-27
fix timestamps optimize_projections

### LINKED ISSUE #1604: Invalid comparison operation: Timestamp(Nanosecond, None) >= Timestamp(Nanosecond, Some("UTC"))
DBT snowplow web fail to compare dates with TZ and without.

 ```
 Database Error in model snowplow_web_vital_measurements (models/optional_modules/core_web_vitals/snowflake/snowplow_web_vital_measurements.sql)
  010001 (02000): Arrow error: Invalid argument error: Invalid comparison operation: Timestamp(Nanosecond, None) >= Timestamp(Nanosecond, Some("UTC"))
  compiled code at target/run/snowplow_web/models/optional_modules/core_web_vitals/snowflake/snowplow_web_vital_measurements.sql
```

In Embucket:
Image
In Snowflake:
Image

# Steps to recreate:
```
-- Create a test table with sample data
CREATE OR REPLACE TABLE test_timestamp_comparison (
    id INTEGER,
    derived_tstamp TIMESTAMP,
    description STRING
);

-- Insert test data with various timestamps
INSERT INTO test_timestamp_comparison VALUES
    (1, '2025-01-01 10:00:00', 'test'),
    (2, '2025-08-01 10:00:00', 'test'),
    (3, '2025-08-15 10:00:00', 'test'),
    (4, '2025-08-20 10:00:00', 'test'),
    (5, '2025-08-01 10:00:00', 'test');

-- Test 1: Original problematic syntax
SELECT 
    'Test 1 - Original syntax' as test_name,
    COUNT(*) as record_count,
    MIN(derived_tstamp) as min_timestamp,
    MAX(derived_tstamp) as max_timestamp
FROM test_timestamp_comparison
WHERE cast(derived_tstamp as date) >= 
    dateadd(
        day,
        -28,
        date_trunc('day', 
    convert_timezone('UTC', convert_timezone('UTC', current_timestamp()))::timestamp
)
    );
```

### LINKED ISSUE #1612: Optimizer rule 'optimize_projections' failed caused by / Internal error: Failed due to a difference in schemas: original schema: DFSchema
Another issue in dbt -snowplwow-web related to difference in schema, where one field with timezone and another without but the error code is a bit different from the previous issue so I decided to create a new ticket.

Embucket:
Image

Snowflake:

Image

Query:
`select convert_timezone('UTC', convert_timezone('UTC', current_timestamp()))::timestamp as model_tstamp`

Might be related to:
https://github.com/Embucket/embucket/issues/1604

### PR DESCRIPTION
**Pull request by osipovartem**
Closes #1612
Closes #1604 

If the input in to_timestamp func is already a timestamp, we don't need to change it type since during execution it will be used as is with initial type. 
By this we avoid optimize_projections error when cheking return type and actual result type for the function call.

### DIFF
--- a/crates/core-executor/src/tests/query.rs
+++ b/crates/core-executor/src/tests/query.rs
@@ -610,6 +610,11 @@ test_query!(
 );
 
 test_query!(
+    timestamp_with_timezone_to_timestamp,
+    "SELECT TO_TIMESTAMP(CONVERT_TIMEZONE('UTC', '2024-12-31 10:00:00.000'::TIMESTAMP)) as model_tstamp;"
+);
+
+test_query!(
     timestamp_timezone,
     "SELECT TO_TIMESTAMP(1000000000)",
     setup_queries = ["ALTER SESSION SET timestamp_input_mapping = 'timestamp_tz'"]
--- a/crates/core-executor/src/tests/snapshots/query_timestamp_with_timezone_to_timestamp.snap
+++ b/crates/core-executor/src/tests/snapshots/query_timestamp_with_timezone_to_timestamp.snap
@@ -1,0 +1,13 @@
+---
+source: crates/core-executor/src/tests/query.rs
+description: "\"SELECT TO_TIMESTAMP(CONVERT_TIMEZONE('UTC', '2024-12-31 10:00:00.000'::TIMESTAMP)) as model_tstamp;\""
+---
+Ok(
+    [
+        "+----------------------+",
+        "| model_tstamp         |",
+        "+----------------------+",
+        "| 2024-12-31T10:00:00Z |",
+        "+----------------------+",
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: crates/core-executor/src/tests/query.rs [hunk 1]
```diff
@@ -609,6 +609,11 @@ test_query!(
     setup_queries = ["SET timestamp_input_format = 'mm/dd/yyyy hh24:mi:ss'"]
 );
 
+test_query!(
```
rampage644, line 612: `cargo insta review` forgotten?

rampage644 (dismissed):

rampage644:
Is it based on another PR/branch - seems it is not against `main
`

Vedin (approved):

PR #1623, opened 2025-08-28
Update iceberg rust

JanKaul:
Updates iceberg-rust to its most recent version. Introduces a new parquet writer that uses the datafusion [ParquetSink](https://docs.rs/datafusion/latest/datafusion/datasource/file_format/parquet/struct.ParquetSink.html).

Writing data currently still has the issue that the query restarts after approx. 1 min. However, this issue is not introduced by this PR as it already existed before (I tested the main branch).

### DIFF
--- a/crates/core-metastore/src/metastore.rs
+++ b/crates/core-metastore/src/metastore.rs
@@ -556,11 +556,11 @@ impl Metastore for SlateDBMetastore {
                     .build()
                 })?;
 
+                let schema = url_encode(&ident.schema);
+                let table = url_encode(&ident.table);
+
                 let prefix = volume.prefix();
-                format!(
-                    "{prefix}/{}/{}/{}",
-                    ident.database, ident.schema, ident.table
-                )
+                format!("{prefix}/{}/{}/{}", ident.database, schema, table)
             };
 
             let metadata_part = format!("metadata/{}", Self::generate_metadata_filename());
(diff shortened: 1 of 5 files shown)


osipovartem (approved):

DanCodedThis:
@JanKaul very strange error in a ddl test

JanKaul:
It's an issue with url_encoding the schema and table identifiers in the REST catalog in combination with the ObjectStoreRegistry. The new ParquetSink also uses the ObjectStoreRegistry for writing files (before it didn't). I think there is a mismatch with the encoding somewhere. I just can't find it.

crates/core-metastore/src/metastore.rs
```diff
@@ -556,11 +556,11 @@ impl Metastore for SlateDBMetastore {
                     .build()
                 })?;
 
+                let schema = url_encode(&ident.schema);
+                let table = url_encode(&ident.table);
+
```
JanKaul, line 561: I'm not sure if this is the intended behavior but I think this previously happened behind the scenes.

DanCodedThis: @osipovartem cc

DanCodedThis (approved):
@JanKaul LGTM

crates/core-metastore/src/metastore.rs
DanCodedThis: cc @osipovartem

DanCodedThis (approved):
@JanKaul LGTM

rampage644 (approved):

PR #1625, opened 2025-08-28
Update view table source when scaning

### LINKED ISSUE #1601: View cache not invalidated when source table data changes
## Problem

Views are not returning updated results when data is inserted into their source tables. The view appears to be cached and not invalidated when the underlying table data changes.

## Reproduction

The issue can be reproduced with the following test case in `crates/core-executor/src/tests/sql/ddl/view.rs`:

```rust
test_query!(
    view_from_table,
    "SELECT * FROM view ORDER BY id",
    setup_queries = [
        "CREATE TABLE embucket.public.view_source (ID INTEGER, name VARCHAR)",
        "CREATE OR REPLACE VIEW embucket.public.view AS SELECT * FROM embucket.public.view_source",
        "INSERT INTO embucket.public.view_source VALUES (1, 'Alice'), (2, 'Bob')",
    ],
    snapshot_path = "view"
);
```

## Expected Result
The query should return:
- Row 1: id=1, name='Alice'  
- Row 2: id=2, name='Bob'

## Actual Result
The view returns empty results ("+", "+") instead of the inserted data.

## Root Cause
The view caching mechanism is not properly invalidated when data is inserted/updated/deleted in the source tables that the view depends on.

## Impact
- Views do not reflect current data state
- Applications relying on views will see stale data
- Data consistency issues between direct table queries and view queries

## Solution Required
The view cache invalidation logic needs to be updated to:
1. Track dependencies between views and their source tables
2. Invalidate view cache when any dependent table is modified
3. Ensure proper cache invalidation across transactions

osipovartem:
Closes #1601 

If this table is a View, we need to ensure it reflects the latest state of its underlying tables.

ViewTable contains a logical plan that references the snapshot of the source tables it the time the view was created. Without updating TableScan nodes, the view would continue to reference stale data. Here we reconstruct the logical plan with updated TableScan nodes so that any query on the view sees the latest snapshots of the source tables.

### DIFF
--- a/crates/df-catalog/src/table.rs
+++ b/crates/df-catalog/src/table.rs
@@ -1,11 +1,17 @@
+use crate::df_error;
 use async_trait::async_trait;
 use datafusion::arrow::datatypes::SchemaRef;
 use datafusion::catalog::{Session, TableProvider};
+use datafusion::datasource::{ViewTable, provider_as_source};
+use datafusion::execution::SessionState;
+use datafusion_common::tree_node::{Transformed, TreeNode};
 use datafusion_expr::dml::InsertOp;
-use datafusion_expr::{Expr, TableType};
+use datafusion_expr::{Expr, LogicalPlan, TableScan, TableType};
 use datafusion_physical_plan::ExecutionPlan;
 use once_cell::sync::OnceCell;
+use snafu::OptionExt;
 use std::any::Any;
+use std::collections::HashMap;
 use std::sync::Arc;
 
 pub struct CachingTable {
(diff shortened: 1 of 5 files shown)


crates/df-catalog/src/table.rs
```diff
@@ -74,3 +92,77 @@ impl TableProvider for CachingTable {
...
+        .transform_up(|ref plan| {
+            if let LogicalPlan::TableScan(scan) = plan {
+                if let Some(new_scan) = replacements.get(&scan.table_name.to_string()) {
+                    Ok(Transformed::yes(LogicalPlan::TableScan(new_scan.clone())))
+                } else {
+                    Ok(Transformed::no(plan.clone()))
+                }
+            } else {
+                Ok(Transformed::no(plan.clone()))
+            }
```
DanCodedThis: nitpick: This can be (since our rust version update):
```rust 
if let LogicalPlan::TableScan(scan) = plan
&& let Some(new_scan) = replacements.get(&scan.table_name.to_string()) {
                    Ok(Transformed::yes(LogicalPlan::TableScan(new_scan.clone())))
                } else {
                    Ok(Transformed::no(plan.clone()))
                }
```

DanCodedThis (approved):
@osipovartem LGTM

DanCodedThis:
@osipovartem flaky test?

rampage644 (approved):

DanCodedThis (approved):
@osipovartem LGTM

PR #1637, opened 2025-08-29
Asynchronous query execution

### LINKED ISSUE #1609: Asynchronous query execution
### 🛠️  Proposed changes to support async queries in Embucket
 
* QueryHistory
  - Add query status: Canceled, TimedOut  
* Extend Execution service interface, by adding methods:
  - `query` - preserve existing method for synchronous execution
     re-implement using async query execution under the hood
  - `cancel_query` - cancel running query either sync or async
  - `submit_query` - submt query asynchronously
  - `wait_async_query_completion` - wait for query completion synchronously (blocking mode)
  - `get_query_result` - get result of finished query
* Support async execution from snowflake CLI, when added following **query prefix: `;>`** so it is expected to run in async mode
  - Support async queries like: `snow sql -q "create table foo;>" `
  - POST query-request?requestId=eb496531-7555-4db0-a4c3-600e985f0132
    support param **asyncExec=true**
    save requestId as it will be used by abort
  - POST abort-request?requestId=
    disregard "sqlText" query string param
* exec from UI - add rest api support
  - post /ui/queries?mode=async
  - delete /ui/queries/{queryRecordId}

### LINKED ISSUE #1636: Snowflake CLI error: Invalid query ID
### 📝 Running query against Embucket
Currently query_id returned to snowflake cli is not uuid based, this results in following error:
```
Statement compilation error: Invalid query ID passed to 'result' command: 8243530241068
```

Code of python connector for context:
``` python
if not VALID_UUID_RE.match(qid):
    return f"Invalid query ID passed to '{command_name}' command: {qid}"
```

### 🐛 Entire flow reproducing the bug
Async query execution
```
snow sql  -q "create table foo;>"
scheduled query ID: 8243530113286
```
Async query results polling
```
snow sql  -q '!result 8243530113286
Invalid query ID passed to 'result' command: 8243530113286
Error: SQL rendering error
```

### 🛠️ Proposed solution
Create pseudo Uuid from query_id::i64, return it to the client and expect it back.
For compatibility with Embucket original i64 query_id should work as well when specified as input.

So following should work:
Submit query:
```
snow sql -q "select 1;>"
select 1
+--------------------------------------+
| scheduled query ID                   |
|--------------------------------------|
| 00000000-0000-0000-0000-077f4243a656 |
+--------------------------------------+
```

Get result:
```
snow sql -q '!result 00000000-0000-0000-0000-077f4243a656'
+----------+
| Int64(1) |
|----------|
| 1        |
+----------+
```

With error handling.
Note: None is retuned in error message instead of query id:
```
snow sql -c local-embucket -q '!result 00000000-0000-0000-0000-077f4243a656'
╭─ Error ──────────────────────────────────────────────────────────────────────╮
│ 010001 (02000): None: Query History error:   │
│ Can't locate query record by query_id: 00000000-0000-0000-0000-077f4243a656  │
│ (8243153970774)                                                              │
╰──────────────────────────────────────────────────────────────────────────────╯
```

### LINKED ISSUE #1638: Add CANCEL QUERY function
### 🪣  Embucket: Add function for cancelling running queries from SQL interface
`CANCEL_QUERY` function should support both uuid, integer ids 

###  ❄️ Snowflake: [CANCEL_QUERY](https://docs.snowflake.com/en/sql-reference/functions/system_cancel_query)

Example:
```
SELECT SYSTEM$CANCEL_QUERY('d5493e36-5e38-48c9-a47c-c476f2111ce5');
+-------------------------------------------------------------+
| SYSTEM$CANCEL_QUERY('D5493E36-5E38-48C9-A47C-C476F2111CE5') |
|-------------------------------------------------------------|
| query [d5493e36-5e38-48c9-a47c-c476f2111ce5] terminated.    |
+-------------------------------------------------------------+
```

YaroslavLitvinov:
closes #1609
closes #1636 
closes #1638

Known issue: due to milliseconds granularity on query_id, ids clash with each other, reproducible by snowflake-rest-api tests

PR includes:
* ExecutionService supporting async queries
  -  add query execution statuses: Canceled, TimedOut
  - queries concurrency limit approach has changed, instead of Semaphore based permit
  - proper tracing spans for spawned tasks (like `submit_query`)
  - Async Query lifecycle management: submit, abort, wait, etc..
* Changes in snowflake rest:
  - Add `/queries/{queryId}/result` handler for `snow sql "!result <query_id>"`
  - Implement aborting queries with request-abort handler (by request_id)
  - test suite for quick tests build / execution without dependency on server:  `cargo test-rest`
  - snow_sql client api supporting snow_sql extensions:
    - `!result _UUID_` - waits for query_id=_UUID_ query completion and return its result
    - `!abort _UUID_` - substituted by `SELECT SYSTEM$CANCEL_QUERY` udf
    - `_QUERY_;>` - runs _QUERY_ asynchronously
  - custom gzip support replaced by http tower layer
* QueryRecord refactored in QueryHistory
* RunningQueries API for canceling queries
* Affected READMEs updated
* Other things

# Performance notes:
In comparison to old version of synchronous execution, the newer version that uses asynchronous execution shows performance degrade about 30% as now `submit_query` function used for every execution, hits twice `QueryHistory` by saving intermediate query status: **Running** when it started and updating its status upon finish. Regardless if query executed synchronously (api-ui, tests) or asynchronously. Following tests take `63.32s` in old version vs `94.80s` in version introduced in this PR. These Embucket-seed tests really indicative since they are running many sqls in loop.
```
running 3 tests
test tests::test_seed_templates_parseable ... ok
test tests::test_seed_roundtrip ... ok
test tests::test_seed_client has been running for over 60 seconds
test tests::test_seed_client ... ok
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 94.80s
```


crates/core-executor/src/lib.rs
```diff
@@ -3,21 +3,77 @@ use std::sync::Arc;
...
+///
+/// This will block the executing thread for 5 seconds before returning the value `5`.
+///
+/// # Notes
+/// - This UDF performs blocking (`thread::sleep`), and should **not** be used in production settings,
+///   as it will block the execution thread and may reduce system throughput.
+/// - The function is marked as `Volatile`, indicating that it has side effects (i.e., sleeping)
+///   and cannot be optimized away by the query planner.
+#[allow(clippy::unwrap_used)]
+fn sleep_udf() -> ScalarUDF {
```
osipovartem: let`s move this func to embucket-fucntions crate if possible

osipovartem (approved):

crates/core-executor/src/service.rs
```diff
@@ -313,67 +424,242 @@ impl ExecutionService for CoreExecutionService {
         query: &str,
         query_context: QueryContext,
     ) -> Result<QueryResult> {
-        // Attempt to acquire a concurrency permit without waiting.
-        // This immediately returns an error if the concurrency limit has been reached.
-        // If you want the task to wait until a permit becomes available, use `.acquire().await` instead.
+        let query_handle = self.submit_query(session_id, query, query_context).await?;
+        self.wait_submitted_query_result(query_handle).await
+    }
```
DanCodedThis, line 429: @YaroslavLitvinov Does this mean that we always now use channels/async exec?

crates/core-history/src/history_store.rs
```diff
@@ -432,7 +410,7 @@ mod tests {
             db.add_query(item).await.expect("Failed adding query");
         }
 
-        let cursor = <QueryRecord as IterableEntity>::Cursor::min_cursor();
+        let cursor = QueryRecordId(<QueryRecord as IterableEntity>::Cursor::min_cursor());
         eprintln!("cursor: {cursor}");
```
DanCodedThis, line 414: Trace instead?

DanCodedThis (approved):
@YaroslavLitvinov LGTM

crates/core-history/src/history_store.rs
YaroslavLitvinov: This is in tests. Think it's intentionally when assert happen you already have a context

crates/core-executor/src/service.rs
YaroslavLitvinov: Yes, QueryEngine has evolved and  under the hood it has an async nature. In case of SYNC execution it uses just one `tokio::oneshot::channel`.
Edited: sync query execution left for compatibility with existing tests, also api-ui uses sync version.

PR #1642, opened 2025-08-31
dbt snowplow workflow change

osmazur:
Merged run.sh and run_snowplow_web.sh into one fille.
Added dbt snowplow web incremental run as a separate workflow.
Changed the old snowplow workflow to use data generation and copy into.
Updated the readme file:
image


osipovartem (approved):

test/dbt_integration_tests/dbt-snowplow-web/incremental.sh
```diff
@@ -10,16 +10,25 @@ echo "Setting up Docker container"
...
 
+# install requirements
+
+python3 -m venv env
+source env/bin/activate
+
+python3 -m pip install --upgrade pip >/dev/null 2>&1
+
+pip install -r requirements.txt >/dev/null 2>&1
+
```
JanKaul: This will create a separate venv in addition to the one in `dbt-snowplow-web/dbt-snowplow-web/env`. I think it would be better to just use one.

DanCodedThis (approved):

test/dbt_integration_tests/dbt-snowplow-web/incremental.sh
osmazur: you are right, made an update to keep it in one place, as well as merged run.sh and run_snowplow_web.sh to keep dbt related part in one file.

DanCodedThis (approved):

PR #1646, opened 2025-09-01
Fix type_of, is_decimal

osipovartem:
- Fixed type_of for already specified types
- Added missing is_decimal
- Fixed incorrect 'null' hardcoded value

related to #1645


crates/embucket-functions/src/json.rs
```diff
@@ -681,14 +682,19 @@ pub fn encode_struct_array(array: ArrayRef) -> Result<JsonValue, ArrowError> {
 
 /// Encodes a Decimal128 Arrow array into a JSON array
 pub fn encode_decimal128_array(array: ArrayRef) -> Result<JsonValue, ArrowError> {
-    let array = array.as_primitive::<Decimal128Type>();
+    let array = cast(&array, &DataType::Float64)?;
```
Vedin, line 688: I'm not quite understanding this logic. Why we cast to Float64 here instead of Decimal128? 
If it's required, then please comment it out. Why can't we reuse `encode_float64_array`?

Vedin (approved):
lgtm, just small question about decimals

crates/embucket-functions/src/json.rs
osipovartem: In the code below we converted decimal to string, Decimal(314, 3, 2) was converted as 314 instead of 3.14
json parser can work with numbers as floats and integers.

encode_float64_array can work only with Float64 array type, not Decimal(_, _)
So we have to cast Decimal to Float64
But you are right, why we should return JsonValue::String instead of JsonValue::Number
I'll fix

Vedin (approved):

PR #1654, opened 2025-09-02
Make flatten non blocking

osipovartem:


crates/embucket-functions/src/table/flatten/provider.rs
```diff
@@ -206,147 +208,150 @@ impl ExecutionPlan for FlattenExec {
...
-                    self.args.is_outer,
-                    flatten_func.row_id.load(Ordering::Acquire),
-                )],
-                self.schema.clone(),
-                self.projection.clone(),
-            )?));
-        }
+        })
+        .try_flatten()
+        .map_err(|e| e);
```
DanCodedThis, line 316: What is the point of mapping an error to itself? Do we accidentally wrap it in our error type? `.context()` is preferable :)

crates/embucket-functions/src/table/flatten/provider.rs
```diff
@@ -206,147 +208,150 @@ impl ExecutionPlan for FlattenExec {
...
+                        schema.clone(),
+                        projection.clone(),
+                    )?
+                    .boxed(),
+                )
+            } else {
+                Ok::<_, DataFusionError>(
+                    MemoryStream::try_new(all_batches, schema.clone(), projection.clone())?
+                        .with_fetch(limit)
+                        .boxed(),
```
DanCodedThis, line 311: Can this fail with memory?

DanCodedThis (approved):
@osipovartem LGTM

crates/embucket-functions/src/table/flatten/provider.rs
osipovartem: we use memory here to apply projection and limit for raw RecordBatch

crates/embucket-functions/src/table/flatten/provider.rs
osipovartem: try_flatten doesn't return an error directly

PR #1658, opened 2025-09-03
Add timezone for JSON type

### LINKED ISSUE #1634: Python connector fails to deserialize a response
It also causes snowplow model `snowplow_web_base_sessions_lifecycle_manifest` to fail, but here is the most simple query to reproduce (make sure format is json, it works as usual with arrow):

```sh
 > select convert_timezone('UTC', current_timestamp());
Failed to convert: field convert_timezone(Utf8("UTC"),now()): TIMESTAMP_TZ::1756424096.0, Error: not enough values to unpack (expected 2, got 1)
Traceback (most recent call last):
  File "/Users/ramp/.asdf/installs/python/3.12.2/lib/python3.12/site-packages/snowflake/connector/result_batch.py", line 521, in _parse
    row_result[idx] = v if c is None or v is None else c(v)
                                                       ^^^^
  File "/Users/ramp/.asdf/installs/python/3.12.2/lib/python3.12/site-packages/snowflake/connector/converter.py", line 252, in conv
    value, tz = encoded_value.split()
    ^^^^^^^^^
ValueError: not enough values to unpack (expected 2, got 1)

Error occurred: 252005: 252005: Failed to convert: field convert_timezone(Utf8("UTC"),now()): TIMESTAMP_TZ::1756424096.0, Error: not enough values to unpack (expected 2, got 1)
```

osipovartem:
TIMESTAMP_TZ format is parsed by snowflake python connector by the func below
```python
def _TIMESTAMP_TZ_to_python(self, ctx: dict[str, Any]) -> Callable:
    """Converts TIMESTAMP TZ to datetime.

    The timezone offset is piggybacked.
    """
    scale = ctx["scale"]

    def conv(encoded_value: str) -> datetime:
        value, tz = encoded_value.split()
        tzinfo = _generate_tzinfo_from_tzoffset(int(tz) - 1440)
        return SnowflakeConverter.create_timestamp_from_string(
            value=value, scale=scale, tz=tzinfo
        )

    return conv
```
So we have to add possitive offset in minutes for JSON data serialization format
Example,
1_627_846_261_233_222_111 nanoseconds with America/Los_Angeles timezone should be parsed as `1627846261.233222111 1020` where **1020-1440 = -420 (PDT)**

Closes #1634

### DIFF
--- a/crates/core-executor/src/utils.rs
+++ b/crates/core-executor/src/utils.rs
@@ -25,6 +25,7 @@ use datafusion::arrow::record_batch::RecordBatch;
 use datafusion::common::ScalarValue;
 use datafusion_common::TableReference;
 use datafusion_expr::{Expr, LogicalPlan};
+use embucket_functions::conversion::to_timestamp::parse_timezone;
 use indexmap::IndexMap;
 use serde_json::Value;
 use snafu::{OptionExt, ResultExt};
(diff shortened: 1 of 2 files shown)


DanCodedThis (approved):
@osipovartem LGTM

rampage644 (approved):

PR #1673, opened 2025-09-05
[WIP] Add snowflake time benchmark for tpcds data

YevheniiNiestierov:
Added time (execution + compilation) benchmark for tpc-ds dataset. Will update some queries, for some reason return empty result, will investigate (queries: 32, 37, 58, 82, 92, 94, 95)
output example:
Screenshot 2025-09-05 at 23 14 26


andheroe:
@YevheniiNiestierov having screenshot of the output would be so much better for understanding your change

andheroe (approved):

PR #1691, opened 2025-09-15
Update replace func with missing params

### LINKED ISSUE #1690: Error during planning: Function 'replace' expects 3 arguments but received 2 No function matches
Extend replace func by default empty string as replacement

osipovartem:
Closes #1690


DanCodedThis (approved):
@osipovartem LGTM

PR #1696, opened 2025-09-15
Update data loading paths and environment variable names in Embucket …

YevheniiNiestierov:
…integration, update readme

### DIFF
--- a/benchmark/infrastructure/env.tpl
+++ b/benchmark/infrastructure/env.tpl
@@ -9,8 +9,8 @@ SERVER_ADDRESS=${server_address}
 VITE_API_URL=${vite_api_url}
 
 # Database initialization configuration
-DATABASE_NAME=benchmark_db
-SCHEMA_NAME=benchmark_schema
+DATABASE_NAME=embucket
+SCHEMA_NAME=public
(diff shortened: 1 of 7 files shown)


benchmark/infrastructure/env.tpl
```diff
@@ -9,8 +9,8 @@ SERVER_ADDRESS=${server_address}
 VITE_API_URL=${vite_api_url}
 
 # Database initialization configuration
-DATABASE_NAME=benchmark_db
-SCHEMA_NAME=benchmark_schema
```
andheroe, line 13: You can't just change this. In the init script there is a bunch of logic that creates this db and schema, then validates. In your benchmark script you need to add these names to construct the full table name. If you don't use the full qualifying name of the table, just the name for the table only - then by default it uses "embucket" and "public". That's why it works.

andheroe (changes requested):

benchmark/infrastructure/env.tpl
YevheniiNiestierov: Good to know, didn't pay attention to that, thanks

andheroe (approved):

PR #1697, opened 2025-09-16
fix merge single partition execution

JanKaul:
Fixes the bug where Merge Into takes only one partition into account. It now coalesces the partitions before performing the Merge operation.


osipovartem (approved):

osipovartem (approved):

DanCodedThis (approved):
@JanKaul LGTM

PR #1699, opened 2025-09-16
Try to cast to utf8 for if the type is numeric in lenght func

### LINKED ISSUE #1698: Function 'length' failed to match any signature
```Error during planning: Internal error: Function 'length' failed to match any signature, errors: Internal error: Expect TypeSignatureClass::Native(LogicalType(Native(String), String))```

### PR DESCRIPTION
**Pull request by osipovartem**
Closes #1698

### DIFF
--- a/crates/embucket-functions/src/string-binary/length.rs
+++ b/crates/embucket-functions/src/string-binary/length.rs
@@ -140,6 +140,7 @@ impl ScalarUDFImpl for LengthFunc {
                     .collect::<UInt64Array>();
                 Arc::new(new_array)
             }
+            DataType::Null => Arc::new((0..arr.len()).map(|_| None).collect::<UInt64Array>()),
             _ => {
                 return exec_err!("Invalid datatype");
             }
--- a/crates/embucket-functions/src/string-binary/length.rs
+++ b/crates/embucket-functions/src/string-binary/length.rs
@@ -1,16 +1,13 @@
 use datafusion::arrow::array::{Array, AsArray};
+use datafusion::arrow::compute::cast;
 use datafusion::arrow::{array::UInt64Array, datatypes::DataType};
 use datafusion::error::Result as DFResult;
-use datafusion::logical_expr::{
-    Coercion, Signature, TypeSignature, TypeSignatureClass, Volatility,
-};
+use datafusion::logical_expr::{Signature, TypeSignature, Volatility};
 use datafusion_common::cast::{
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: crates/embucket-functions/src/string-binary/length.rs [hunk 1]
```diff
@@ -44,17 +41,10 @@ impl LengthFunc {
...
-                        TypeSignatureClass::Native(logical_string()),
-                    )]),
-                    TypeSignature::Coercible(vec![Coercion::new_exact(
-                        TypeSignatureClass::Native(logical_binary()),
-                    )]),
-                ],
-                Volatility::Immutable,
-            ),
+            signature: Signature {
+                type_signature: TypeSignature::Any(1),
```
Vedin, line 45: Probably, not important for this function but in general I prefer using `Signature::user_defined(Immutable)` and then implement `coerce_types` itself.

Vedin (approved):

crates/embucket-functions/src/string-binary/length.rs
```diff
@@ -141,6 +131,16 @@ impl ScalarUDFImpl for LengthFunc {
                     .collect::<UInt64Array>();
                 Arc::new(new_array)
             }
+            v if v.is_numeric() => {
+                let casted = cast(&arr, &DataType::Utf8)?;
+                let strs = as_string_array(&casted)?;
+                let new_array = strs
+                    .iter()
+                    .map(|array_elem| array_elem.map(|value| value.chars().count() as u64))
```
DanCodedThis: Why not just `value.len()`? Chars count, means we make an iterator and collect the length by using next until we get to None. Essentially, one is constant time O(1) and the second one is O(n).

crates/embucket-functions/src/string-binary/length.rs
```diff
@@ -141,6 +131,16 @@ impl ScalarUDFImpl for LengthFunc {
...
+                    .collect::<UInt64Array>();
+                Arc::new(new_array)
+            }
+            DataType::Null => Arc::new((0..arr.len()).map(|_| None).collect::<UInt64Array>()),
```
DanCodedThis: Is there `UInt64Array::new_nulls`? It's probably faster, since it wouldn't use an iterator, I think it would just write out the buffer.

DanCodedThis (approved):
@osipovartem LGTM

DanCodedThis (approved):

PR #1713, opened 2025-09-18
Substitute aliases in selection filter

### LINKED ISSUE #1712: Pass aliases to selection filter

osipovartem:
- fix aliases scope priority (take into account only aliases which already added to the map)
- added logic to substitute aliases for selection filter (having, order_by and group_by  can be added in future)
- fixed error when column cannot be found if this is an alias

**gitlab 93.4% success rate**
Closes #1712


DanCodedThis (approved):
@osipovartem LGTM

JanKaul (approved):

PR #1721, opened 2025-09-19
Update df 50

Vedin:
Known cavets: 
1. `ObjectNamePart::Function` does nothing for now. functions with https://docs.snowflake.com/en/sql-reference/identifier-literal
2. Named arguments for Literals and TableFunctions is not used. 
3. Some functions have their own implementation for `Eq` or `Hash` traits. These traits are now required by datafusion but some of our functions can support them automatically. There is probably a more elegant solution for it. I just implement these traits where needed.


crates/core-executor/src/snowflake_error.rs
```diff
@@ -327,7 +327,7 @@ fn datafusion_error(df_error: &DataFusionError, subtext: &[&str]) -> SnowflakeEr
     let message = df_error.to_string();
     match df_error {
         DataFusionError::ArrowError(arrow_error, ..) => {
-            match arrow_error {
+            match &**arrow_error {
```
osipovartem: maybe some clone here

osipovartem (approved):

crates/core-executor/src/snowflake_error.rs
Vedin: `clone` didn't work for me here because enum is expected. So, to make it less ugly I replace dereferencing with `as_ref()`? Sounds good?

rampage644 (approved):

DanCodedThis (approved):
@Vedin Let's merge and fix in subsequent prs if the errors are not crazy enough

DanCodedThis (approved):

osipovartem (approved):

PR #1725, opened 2025-09-19
Refactor benchmark preparation and execution logic for TPCH queries

YevheniiNiestierov:
Next step is to add option to load dataset from s3 to snowflake. Removed old logic for loading data to snowflake from local file and logic for generation dataset using duckdb.
Will update result display logic soon.
Also need to rename schemas in SF to unify logic for env variables.
example of output structure for Embucket results:
image

### DIFF
--- a/benchmark/calculate_average.py
+++ b/benchmark/calculate_average.py
@@ -72,25 +72,21 @@ def calculate_benchmark_averages(schema, warehouse, is_embucket=False):
                 total_value = total_row['Total (ms)'].values[0]
                 total_values.append((i, total_value))
 
-        if len(total_values) >= 3:  # Only remove if we have enough files with TOTAL rows
-            # Sort by total values
-            total_values.sort(key=lambda x: x[1])
+        # Sort by total values
+        total_values.sort(key=lambda x: x[1])
 
-            # Get indices of min and max files
-            min_idx = total_values[0][0]
-            max_idx = total_values[-1][0]
+        # Get indices of min and max files
+        min_idx = total_values[0][0]
+        max_idx = total_values[-1][0]
 
-            print(f"Removing file with lowest TOTAL: {csv_files[min_idx]} (value: {total_values[0][1]})")
-            print(f"Removing file with highest TOTAL: {csv_files[max_idx]} (value: {total_values[-1][1]})")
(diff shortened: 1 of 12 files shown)


benchmark/utils.py
```diff
@@ -29,6 +29,7 @@ def create_embucket_connection():
         "host": host,
         "protocol": protocol,
         "port": int(port) if port else 3000,
+        "socket_timeout": 1200, # connector restarts query if timeout is reached
```
andheroe: Is this seconds? Please add to the comment

benchmark/calculate_average.py
```diff
@@ -0,0 +1,133 @@
...
+
+            print(f"Removing file with lowest TOTAL: {csv_files[min_idx]} (value: {total_values[0][1]})")
+            print(f"Removing file with highest TOTAL: {csv_files[max_idx]} (value: {total_values[-1][1]})")
+
+            # Filter out min and max files
+            filtered_indices = [i for i in range(len(all_dfs)) if i != min_idx and i != max_idx]
+            dfs = [all_dfs[i] for i in filtered_indices]
+            used_files = [csv_files[i] for i in filtered_indices]
+            print(f"Using {len(dfs)} files for averaging: {used_files}")
+        else:
```
andheroe: why do we need this else? In what scenario will we have not enough rows? if this happens we should fail, because we didn't agree on this behavior in benchmarking methodology

benchmark/calculate_average.py
```diff
@@ -0,0 +1,133 @@
...
+            print(f"Not enough files with TOTAL rows, using all {len(dfs)} files")
+
+        # Sort each DataFrame by 'Query' to align rows
+        dfs = [df.sort_values('Query').reset_index(drop=True) for df in dfs]
+
+        # Concatenate DataFrames along a new axis
+        stacked = pd.concat(dfs, axis=0, keys=range(len(dfs)))
+
+        # Determine columns based on pattern type
+        if pattern_type == 'embucket':
```
andheroe: This is overcomplication that will shoot you in the foot. Let's collect only total ms. We are not using compilation and execution times anyway. ALso this pattern matching where you check `if "embucket"` will be a disaster if we decide to add another database to a benchmark (which is a reasonable scenario)

andheroe (changes requested):

andheroe (approved):

PR #1748, opened 2025-09-24
Refactor benchmarking logic to improve modularity, avg calculation an…

YevheniiNiestierov:
…d add command line argument parsing


benchmark/calculate_average.py
```diff
@@ -5,30 +5,37 @@
...
 
-def natural_sort_queries(df):
-    """Sort queries numerically."""
-    df['sort_key'] = df['Query'].apply(lambda x:
-                                       int(re.search(r'q(\d+)', x).group(1))
-                                       if re.search(r'q(\d+)', x) else 0)
-    df = df.sort_values('sort_key')
-    df = df.drop('sort_key', axis=1)
+
+def add_query_index(df):
```
andheroe: THe function is called "add_query_index" and it's last line removes the index. WAT

benchmark/calculate_average.py
```diff
@@ -5,30 +5,37 @@
...
+    """Add index column to dataframe based on query number."""
+    # Add index column
+    df['query_index'] = df['Query'].apply(lambda x:
+                                          int(re.search(r'q(\d+)', x).group(1))
+                                          if re.search(r'q(\d+)',
+                                                       x) else 9999)  # High value for TOTAL or other special rows
```
andheroe: There is a constant for infinity value - better to use it, it's not unreasonable to have 10000 queries in some benchmark

benchmark/calculate_average.py
```diff
@@ -77,7 +84,7 @@ def calculate_benchmark_averages(schema, warehouse, is_embucket=False):
         averaged = stacked.groupby(level=1)[[COLUMN_TO_AVERAGE]].mean().reset_index(drop=True)
         averaged['Query'] = dfs[0]['Query']
         averaged = averaged[['Query', COLUMN_TO_AVERAGE]]
-        averaged = natural_sort_queries(averaged)
+        averaged = add_query_index(averaged)
```
andheroe: So this function sorts rows, not adds index

benchmark/env_example
```diff
@@ -1,16 +1,22 @@
...
-SNOWFLAKE_PASSWORD=
+SNOWFLAKE_ACCOUNT=your_snowflake_account
+SNOWFLAKE_USER=your_snowflake_user
+SNOWFLAKE_PASSWORD=your_snowflake_password
 SNOWFLAKE_DATABASE=benchmark_db
-SNOWFLAKE_SCHEMA=tpcds
-SNOWFLAKE_WAREHOUSE=BENCHMARK_WH
+SNOWFLAKE_WAREHOUSE=BENCHMARK_WH_XS
 
+EMBUCKET_INSTANCE=c7i_4xlarge
```
andheroe: WHy this instance? Pretty big for a default, better to add smaller as a default

benchmark/utils.py
```diff
@@ -10,14 +10,14 @@ def create_embucket_connection():
...
     # Connection config with defaults
-    host = os.getenv("EMBUCKET_HOST", "localhost")
-    port = os.getenv("EMBUCKET_PORT", "3000")
-    protocol = os.getenv("EMBUCKET_PROTOCOL", "http")
-    user = os.getenv("EMBUCKET_USER", "embucket")
-    password = os.getenv("EMBUCKET_PASSWORD", "embucket")
-    account = os.getenv("EMBUCKET_ACCOUNT") or f"acc_{uuid.uuid4().hex[:10]}"
-    database = os.getenv("EMBUCKET_DATABASE", "embucket")
-    schema = os.getenv("EMBUCKET_SCHEMA", "public")
+    host = os.environ.get("EMBUCKET_HOST", "localhost")
```
andheroe: WHy change to `enviorn` here?
Also, why do we need default values here? We have default values in env example already

andheroe (changes requested):

benchmark/calculate_average.py
```diff
@@ -5,30 +5,36 @@
...
 
-def natural_sort_queries(df):
-    """Sort queries numerically."""
-    df['sort_key'] = df['Query'].apply(lambda x:
-                                       int(re.search(r'q(\d+)', x).group(1))
-                                       if re.search(r'q(\d+)', x) else 0)
-    df = df.sort_values('sort_key')
-    df = df.drop('sort_key', axis=1)
+
+def sort_by_query_index(df):
```
andheroe, line 11: Much better, much easier to read

benchmark/calculate_average.py
```diff
@@ -5,30 +5,36 @@
...
+    """Add index column to dataframe based on query number."""
+    # Add index column
```
andheroe, line 16: Ideally, you need to add an example of the string that you parse here, so it's easier to understand what the regexp is doing.

benchmark/calculate_average.py
```diff
@@ -5,30 +5,36 @@
...
+    # Sort by the new index
+    df = df.sort_values('query_index')
+
+    # Remove the index column
+    df = df.drop('query_index', axis=1)
     return df
 
 
-def calculate_benchmark_averages(schema, warehouse, is_embucket=False):
+def calculate_benchmark_averages(schema, warehouse, database, benchmark_type):
```
andheroe: Let's rename `database` to `system` or something else. We use `database` for an entity inside both systems, so it is confusing.

benchmark/calculate_average.py
```diff
@@ -5,30 +5,36 @@
...
     Args:
         schema: The schema/dataset used in the benchmark
         warehouse: The warehouse/instance used in the benchmark
-        is_embucket: Flag to indicate if processing Embucket results (True) or Snowflake results (False)
+        database: Database used
     """
-    # Directory where benchmark results are stored
-    if is_embucket:
-        search_dir = f'embucket_tpch_results/{schema}/{warehouse}'
+    if database == 'embucket':
```
andheroe: You should never use strings for constants like this. You should use Enums. You won't believe the pain in the ass this is to debug later.

benchmark/env_example
```diff
@@ -1,16 +1,22 @@
...
-EMBUCKET_ACCOUNT=embucket
-EMBUCKET_DATABASE=embucket
-EMBUCKET_SCHEMA=public
\ No newline at end of file
+EMBUCKET_DATABASE=benchmark_database
+EMBUCKET_SCHEMA=benchmark_schema
+
+EMBUCKET_ENABLED=true
+EMBUCKET_PROTOCOL=http
+EMBUCKET_HOST=inctance_ip_address
```
andheroe:
```suggestion
EMBUCKET_HOST=instance_ip_address
```

andheroe (changes requested):

benchmark/calculate_average.py
YevheniiNiestierov: what about "platform"?

andheroe (approved):

Issue #1006, opened 2025-06-04
api: UI Search, ordering, etc for GET `/ui/worksheets` handler (via SQL)


DanCodedThis:
- Connect the DF View from History Schema for Worksheets
- Add `SearchParameters`
- Add tests

PR #1749, opened 2025-09-24
api: UI Queries history search

### LINKED ISSUE #1007: api: UI Search, ordering, etc for GET `/ui/queries` handler (via SQL)
- Connect the DF View from History Schema for Query history
- Add `SearchParameters`
- Add tests

DanCodedThis:
Closes #1007

- [x] Connect the DF View from History Schema for Query history
- [x] Use SQL under the hood
- [x] Fixed tests
- [x] Reusing old parameters (no breaking changes)
- [x] Additional tests

P.S. Rebase on main after merge of #1006 issue

### DIFF
--- a/crates/api-ui/src/queries/handlers.rs
+++ b/crates/api-ui/src/queries/handlers.rs
@@ -1,9 +1,6 @@
 #![allow(clippy::needless_for_each)]
-use crate::queries::error::{GetQueryRecordSnafu, QueryRecordResult, StoreSnafu};
-use crate::queries::models::{
-    GetQueriesParams, QueriesResponse, QueryCreatePayload, QueryCreateResponse, QueryGetResponse,
-    QueryRecord, QueryRecordId,
-};
+use crate::queries::error::{DatetimeSnafu, ExecutionSnafu, GetQueryRecordSnafu, QueriesSnafu, QueryRecordResult, StoreSnafu};
+use crate::queries::models::{GetQueriesParams, QueriesResponse, QueryCreatePayload, QueryCreateResponse, QueryGetResponse, QueryRecord, QueryRecordId, QueryStatus, ResultSet};
 use crate::state::AppState;
 use crate::{apply_parameters, downcast_int64_column, downcast_string_column, error::ErrorResponse, error::Result, queries::error::{self as queries_errors, QueryError}, SearchParameters};
 use api_sessions::DFSessionId;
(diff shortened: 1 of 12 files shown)


osipovartem (approved):

crates/api-ui/src/queries/handlers.rs
```diff
@@ -197,45 +206,107 @@ pub async fn get_query(
...
+            .context(ExecutionSnafu)
+            .context(QueriesSnafu)?;
+        let worksheet_ids = downcast_int64_column(&record, "worksheet_id")
+            .context(ExecutionSnafu)
+            .context(QueriesSnafu)?;
+        let queries = downcast_string_column(&record, "query")
+            .context(ExecutionSnafu)
+            .context(QueriesSnafu)?;
+        let start_times = downcast_string_column(&record, "start_time")
+            .context(ExecutionSnafu)
```
YaroslavLitvinov, line 253: Would be much better to create another pair of wrapper functions for downcast_string_column, downcast_int64_column to  hide context calls, 
`w_downcast_string_column(&record, "start_time")?`

crates/api-ui/src/queries/handlers.rs
```diff
@@ -197,45 +206,107 @@ pub async fn get_query(
...
+        OrderDirection::DESC,
+    );
+    let QueryResult { records, .. } = state
+        .execution_svc
+        .query(&session_id, sql_string.as_str(), context)
+        .await
+        .context(ExecutionSnafu)
+        .context(QueriesSnafu)?;
+    let mut items = Vec::new();
+    for record in records {
```
YaroslavLitvinov, line 242: Would be good having  comment on that block, of what we do here

YaroslavLitvinov: For sure such conversion should be done separately implementing some QueryRecord::try_from

YaroslavLitvinov (commented):
Left comments regarding implemetation details, #1007

DanCodedThis:
> Left comments regarding implemetation details

I'll fix in a subsequent pr, makes sense

crates/api-ui/src/queries/handlers.rs
DanCodedThis: not possible since conversion is done for arrays, can't pass the record batch and also array index, and doing so with tyr_from `QueriesResponse` seems redundant

PR #1751, opened 2025-09-24
Skip an inline CTE that contains itself

### LINKED ISSUE #1750: Fix flatten CTE visitor recursion
Some dbt-gitlab queries contain recursive deps
```sql
...
recursive_namespace_lineage(namespace_id, parent_id, valid_to, valid_from, valid_to_list, valid_from_list,
                             upstream_lineage) AS (
  SELECT
    root.namespace_id,
  FROM parent_change AS root
  WHERE parent_id IS NULL
  UNION ALL
  SELECT
    iter.namespace_id,
  FROM recursive_namespace_lineage AS anchor
  INNER JOIN parent_change AS iter
    ON iter.parent_id = anchor.namespace_id
    AND NOT ARRAY_CONTAINS(iter.namespace_id::VARIANT, anchor.upstream_lineage)
    AND (CASE
           WHEN iter.valid_from BETWEEN anchor.valid_from AND anchor.valid_to THEN TRUE
           WHEN iter.valid_to BETWEEN anchor.valid_from AND anchor.valid_to THEN TRUE
           WHEN anchor.valid_from BETWEEN iter.valid_from AND iter.valid_to THEN TRUE
           ELSE FALSE
         END) = TRUE
)
```

osipovartem:
Closes #1750 
Fixed visitor memory leak

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,7 @@ venv
 __pycache__
 
 .env
+alloc.log
 
 slatedb-prefix/
 warehouse-prefix/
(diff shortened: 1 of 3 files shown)


DanCodedThis (approved):
@osipovartem LGTM

PR #1757, opened 2025-09-25
Refactor benchmark code to use SystemType enum for system identification

YevheniiNiestierov:


benchmark/benchmark.py
```diff
@@ -24,48 +24,43 @@
 logger = logging.getLogger(__name__)
 
 
-class SystemType(Enum):
-    EMBUCKET = "embucket"
-    SNOWFLAKE = "snowflake"
-
-
-def get_results_path(platform: str, benchmark_type: str, scale_factor: str,
+def get_results_path(system: str, benchmark_type: str, scale_factor: str,
```
andheroe: You need to pass enum value like `SystemType.EMBUCKET` and then compare it with Enum value in the code. The whole point of Enum is to make sure you don't use strings, where you can have a typo.

andheroe (changes requested):

benchmark/benchmark.py
YevheniiNiestierov: ops, didnt see that

andheroe (approved):

PR #1759, opened 2025-09-25
Add Snowflake data loading functionality to benchmarking logic, updat…

YevheniiNiestierov:
…e env example

### DIFF
--- a/benchmark/data_preparation.py
+++ b/benchmark/data_preparation.py
@@ -26,15 +26,12 @@ def create_tables(cursor, system):
 def upload_parquet_to_snowflake_tables(cursor, dataset, dataset_scale_factor):
     """Upload parquet files to Snowflake tables from S3 stage."""
     table_names = get_table_names(fully_qualified_names_for_embucket=False)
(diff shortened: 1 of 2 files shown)


benchmark/data_preparation.py
```diff
@@ -1,22 +1,48 @@
...
         print(f"Creating table: {table_name}")
         cursor.execute(ddl_sql.strip())
 
 
+def upload_parquet_to_snowflake_tables(cursor, dataset, dataset_scale_factor):
+    """Upload parquet files to Snowflake tables from S3 stage."""
+    table_names = get_table_names(fully_qualified_names_for_embucket=False)
+    for placeholder, qualified_table_name in table_names.items():
+        # Extract bare table name for the S3 path
+        bare_table_name = qualified_table_name.split('.')[-1]
```
andheroe: But we don't use fully qualified names for snowflake, why do this split?

andheroe (changes requested):

andheroe (approved):

PR #1762, opened 2025-09-25
engine: Remove `dedicated_executor` experiment

### LINKED ISSUE #1760: engine: Vanilla one tokio runtime
- Remove `dedicated_executor`
- Use only one vanilla tokio runtime, no thread pools, just the basics (for now)
- See if it increases performance (query execution speed)
- Investigate if it relaxes the memory problems
- Check if it's the issue of multiple runtimes that cause the memory pools to work incorrectly
- Gather data of how much should io-runtime vs cpu-runtime have threads for their pools is it 1/3 vs 2/3 or 1/4 / 3/4 (or others) for full saturation
- After decide if we should introduce a second runtime and divide io vs cpu tasks

DanCodedThis:
Closes #1760

- [x] **Experiment**: removed dedicated executor in favor of one vanilla tokio runtime
- [x] This is a test to determine and work towards if two global runtimes (io & cpu) would be feasible
- [ ] Also this may fix the memory issues (since the `DedicatedExecutor` seems to not use memory pools)
- [ ] And by removing extra runtimes increase performance
  - By removing the overhead of many runtime switches
  - Removing the tokio threads > physical threads problem
- [ ] May require some tinkering (what is the default worker thread pool of tokio runtime?)


crates/core-executor/src/query.rs
```diff
@@ -2201,50 +2201,44 @@ impl UserQuery {
...
-                let mut schema = df.schema().as_arrow().clone();
-                let records = df.collect().await.context(ex_error::DataFusionSnafu)?;
-                if !records.is_empty() {
-                    schema = records[0].schema().as_ref().clone();
-                }
-                Ok::<QueryResult, Error>(QueryResult::new(records, Arc::new(schema), query_id))
-            })
-            .await
-            .context(ex_error::JobSnafu)??;
+        let stream = tokio::task::spawn(async move {
```
osipovartem, line 2204: why even call tokio::task::spawn(async move?
already async

DanCodedThis: tokio::task::spawn should schedule it on one of the worker threads

Or are you talking about the scope/block itself being async? If so, I just removed the dedicated executor, and haven't changed anything else

PR #1767, opened 2025-09-26
Update Slatedb 0.8.2, eliminate writer latency for query history items 

### LINKED ISSUE #1752: Draft: Decrease query latency caused by puts into query history
### Issue
With adding of asynchronous query execution it was added one more put to query history. 
Before we were just saving QueryRecord containing result, but now we also save intermediate `QueryStatus::Runing`.
Saving intermediate query status enables our WebUI to see information about running queries, in a natural way just fetching a history.
But this approach impacts significantly on a query execution latency. 

### Proposal
Add caching layer in SlateDBHistoryStore, so:
* When put `QueryRecord` it added to the cache, then spawned async task and result returned without waiting completion of the put;
* So no wait when saving items to the history; 
* Cache  should work transparently, so `HistoryStore` interface behave like it access items as usually. Cached data mixed with the real data loaded from slatedb in proper way;  
* Saver task also uses `ExecutionService::wait_historical_query_result` to be notified when to remove item from the cache.

### Thoughts
Implementing this would require some effort, maybe it's not a better timing for this to come up.

YaroslavLitvinov:
Closes #1752

### PR includes

* slatedb version update to 0.8.2
  - It brings less granularity on error, by providing ErrorKind
* Eliminate writer latency for puts coming from query history, by using less durable WriteOptions

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 4
 
 [[package]]
 name = "addr2line"
-version = "0.24.2"
+version = "0.25.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
+checksum = "1b5d307320b3181d6d7954e663bd7c774a838b8220fe0593c86d9fb09f498b4b"
 dependencies = [
  "gimli",
 ]
@@ -391,15 +391,6 @@ name = "array-init"
 version = "2.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3d62b7694a562cdf5a74227903507c56ab2cc8bdd1f781ed5cb4cf9c9f810bfc"
-
-[[package]]
-name = "array-util"
-version = "1.0.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7e509844de8f09b90a2c3444684a2b6695f4071360e13d2fda0af9f749cc2ed6"
-dependencies = [
- "arrayvec",
-]
 
 [[package]]
 name = "arrayref"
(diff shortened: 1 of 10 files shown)


osipovartem (approved):

YaroslavLitvinov:
Changing writer options significantly improved performance, though some query history related tests fails: 

```
failures:
    tests::worksheets::test_ui_worksheets
    tests::worksheets::test_ui_worksheets_search
    tests::worksheets::test_ui_worksheets_sort

```

JanKaul:
~~Weird, does the UI get its values from SlateDB or is it directly accessing the objectstore?~~
Stupid question, it has to go through SlateDB. Weird, because it should be in SlateDBs cache.

YaroslavLitvinov:
Actually tests with that option are flaky, as once there failing 2 of them, then 3. And in CI/CD none is failing

DanCodedThis:
> ~Weird, does the UI get its values from SlateDB or is it directly accessing the objectstore?~ Stupid question, it has to go through SlateDB. Weird, because it should be in SlateDBs cache.

What about `PutOptions` **TTL** could this affect it somehow (a shot in the dark)?

DanCodedThis:
Also `ObjectStoreCacheOptions` on SlateDb creation

DanCodedThis:
> Also `ObjectStoreCacheOptions` on SlateDb creation

It seems default cache options don't activate this

DanCodedThis:
It seems default options doesn't set up anything or at all if you check out the `DbBuilder` **build**? I think we should definitely double check all options, settings, etc for our SlateDb instance.

DanCodedThis:
By default they provide `Moka` & `Foyer` caches. Is `Foyer` worse in our case? We use `Moka`, however we can even build our own cache if necessary with their `DbCache` trait.

cc @osipovartem @rampage644 @Vedin

DanCodedThis:
There are also `DBReaderOptions` which set up their own cache (`Foyer` or `Moka` also). But I didn't see where those options are used yet.

PR #1781, opened 2025-09-27
feat: add caching enabling/disabling commands support for benchmarks …

YevheniiNiestierov:
…and update result paths

### DIFF
--- a/benchmark/benchmark.py
+++ b/benchmark/benchmark.py
@@ -25,13 +25,16 @@ logger = logging.getLogger(__name__)
 
 
 def get_results_path(system: SystemType, benchmark_type: str, dataset_path: str,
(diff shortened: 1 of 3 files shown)


benchmark/benchmark.py
```diff
@@ -87,7 +90,7 @@ def save_results_to_csv(results, filename="query_results.csv", system=None):
                 writer.writerow(["TOTAL", "", total_time, ""])
 
 
-def run_on_sf(cursor, warehouse, tpch_queries):
+def run_on_sf(cursor, warehouse, tpch_queries, cache=False):
```
andheroe:
```suggestion
def run_on_sf(cursor, warehouse, tpch_queries, cache=True):
```

andheroe (changes requested):

andheroe (approved):

PR #1786, opened 2025-09-29
feat: add option to disable result caching for Snowflake benchmarks, …

YevheniiNiestierov:
…update logging

### DIFF
--- a/benchmark/benchmark.py
+++ b/benchmark/benchmark.py
@@ -27,9 +27,12 @@ logger = logging.getLogger(__name__)
 
 def get_results_path(system: SystemType, benchmark_type: str, dataset_path: str,
                      instance: str, warehouse_size: str = None, run_number: Optional[int] = None,
-                     cached: bool = False) -> str:
+                     cached: bool = False, disable_result_cache: bool = False) -> str:
     """Generate path for storing benchmark results."""
-    cache_folder = "cached" if cached else "no_cache"
+    if disable_result_cache and system == SystemType.SNOWFLAKE:
+        cache_folder = "no_result_cache"
+    else:
+        cache_folder = "cached" if cached else "no_cache"
 
     if system == SystemType.SNOWFLAKE:
         # Use warehouse size in the path instead of warehouse name
@@ -42,6 +45,7 @@ def get_results_path(system: SystemType, benchmark_type: str, dataset_path: str,
     if run_number is not None:
         return f"{base_path}/{system.value}_results_run_{run_number}.csv"
     return base_path
+
 
 
 def save_results_to_csv(results, filename="query_results.csv", system=None):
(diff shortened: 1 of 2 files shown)


benchmark/benchmark.py
```diff
@@ -313,17 +317,18 @@ def run_snowflake_benchmark(run_number: int, cache: bool = False):
     sf_cursor = sf_connection.cursor()
 
     # Control query result caching for benchmark
-    if cache:
+    if disable_result_cache or not cache:
```
andheroe:
```suggestion
    if disable_result_cache:
```

benchmark/benchmark.py
```diff
@@ -27,9 +27,12 @@
...
 def get_results_path(system: SystemType, benchmark_type: str, dataset_path: str,
                      instance: str, warehouse_size: str = None, run_number: Optional[int] = None,
-                     cached: bool = False) -> str:
+                     cached: bool = False, disable_result_cache: bool = False) -> str:
     """Generate path for storing benchmark results."""
-    cache_folder = "cached" if cached else "no_cache"
+    if disable_result_cache and system == SystemType.SNOWFLAKE:
+        cache_folder = "no_result_cache"
+    else:
+        cache_folder = "cached" if cached else "no_cache"
```
andheroe: Let's not modify the structure of folders. LEt's use simple folder structure for everything. But add parameters to the name of csv files with results

andheroe: Also cached vs no_cache has nothing to do with "result cache". This if else is super confusing and it mixes up different things

benchmark/benchmark.py
```diff
@@ -421,12 +428,34 @@ def display_comparison(sf_results, emb_results):
...
 
-def run_benchmark(run_number: int, system_enum: Optional[SystemType], no_cache: bool = True):
+def run_benchmark(run_number: int, system_enum: Optional[SystemType], no_cache: bool = True,
+                  disable_result_cache: bool = False):
     """Run benchmarks on the specified system."""
+    # Log cache configuration for better clarity
+    if system_enum == SystemType.SNOWFLAKE:
+        if no_cache:
+            logger.info(
+                "SNOWFLAKE RUN CONFIGURATION: Cold run - warehouse will be suspended between queries, both query compilation cache and result cache are disabled (USE_CACHED_RESULT=FALSE)")
```
andheroe: We should not disable result cache.

andheroe (changes requested):

benchmark/benchmark.py
```diff
@@ -27,15 +27,21 @@
...
                      instance: str, warehouse_size: str = None, run_number: Optional[int] = None,
-                     cached: bool = False) -> str:
+                     cached: bool = False, disable_result_cache: bool = False) -> str:
     """Generate path for storing benchmark results."""
-    cache_folder = "cached" if cached else "no_cache"
+    # Clearer naming: warm = warehouse kept active, cold = warehouse suspended
+    run_mode = "warm" if cached else "cold"
+
+    # Then handle result cache setting separately
+    if system == SystemType.SNOWFLAKE and disable_result_cache:
```
andheroe:
```suggestion
    if disable_result_cache:
```

andheroe (changes requested):

andheroe (approved):

andheroe (approved):

PR #1801, opened 2025-09-30
Limit amount of data saving in the history

### LINKED ISSUE #1796: Limit amount of data we're saving in the history
Limit the number of rows from query result we're saving in the history.
* Using hard limit by batch size: Upper limit is 4GB - 512MB, just do not break slatedb (should be < 4GB)
  It uses the bytes limit and not rows limit, as rows limit is unpredictable as huge rows can easily break such limit logic
* rows limit that directly limits amount of rows

YaroslavLitvinov:
Closes #1796


crates/core-history/src/entities/result_set.rs
```diff
@@ -6,6 +6,9 @@ use serde_json::Value;
 use snafu::{OptionExt, ResultExt};
 use std::fmt;
 
+// ResultSet exceeded limit of 4GB - 512MB
```
rampage644, line 9: Does that mean allowing to save 3gb results? I would start with allowing to save 10-50 rows maximum (this is exact what snowflake allows).

YaroslavLitvinov: batch size  is kind of a hard limit, that prevent breaking slatedb. soft limit is added as well.

crates/core-history/src/entities/result_set.rs
```diff
@@ -78,6 +81,88 @@ pub struct ResultSet {
...
+                self.rows.len() * 90 / 100 // shrink 90 % of rows
+            } else {
+                self.rows.len() * 50 / 100 // shrink 50 % of rows
+            };
+
+            let hard_rows_limit: usize = if self.rows.len() > shrink_on_count {
+                self.rows.len() - shrink_on_count
+            } else {
+                self.rows.len()
+            };
```
DanCodedThis, line 111: What is this part for? If we shrinked it already, wouldn't it always be smaller? Maybe for like small numbers like 0, 1 it wouldn't is it for that?

rampage644 (approved):

DanCodedThis (approved):
@YaroslavLitvinov LGTM, one question

crates/core-history/src/entities/result_set.rs
YaroslavLitvinov: shrink on means how many rows to remove from original batch, so it's just an intermediate value. This is actually about getting hard limit in rows from hard limit in bytes.

PR #1814, opened 2025-09-30
experimental: Main conditional compilation

DanCodedThis:
- This is pr to **main** from experimental, to be able to use it's features using conditional compilation
- There are foure feature flags:
  - vanilla-tokio-runtime - to deactivate dedicated executor
  - none-durable-history-write
  - sort-merge-join (could be a .env var)
  - experimental (which actiavtes all features)
- In practice you should be able to run: `cargo run --features experimental` and it will work and activate everything
- This pr is done to ease up using experimtnal fetures, and be able to use them with the main changes/fixes right away

P.S. Please check if I didn't accidentally rebase wrong (included everything).


DanCodedThis:
Need to fix cargo test --all-features to exclude experimental ones, so our tests would stay green.

PR #1832, opened 2025-10-02
Allow environment variable override for minimum_parallel_output_files

JanKaul:
## Summary

Allows the `DATAFUSION_EXECUTION_MINIMUM_PARALLEL_OUTPUT_FILES` environment variable to override the default value. Previously, the config was loaded from environment but then unconditionally overwritten with the hardcoded `MINIMUM_PARALLEL_OUTPUT_FILES` constant.

## Changes

- Modified `UserSession::new()` to check if the environment variable is set before applying the default value
- The hardcoded value is now only used as a fallback when the environment variable is not set

### DIFF
--- a/crates/core-executor/src/session.rs
+++ b/crates/core-executor/src/session.rs
@@ -79,7 +79,13 @@ impl UserSession {
 
         let session_params = SessionParams::default();
         let session_params_arc = Arc::new(session_params.clone());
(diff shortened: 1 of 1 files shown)


DanCodedThis (approved):
@JanKaul LGTM

osipovartem (approved):

rampage644 (approved):

PR #1845, opened 2025-10-07
Add ability to execute the full SQL query directly using DuckDB in-memory engine

osipovartem:
- added ability to bypass DataFusion entirely and execute the full SQL query directly using DuckDB in-memory engine
- added duckdb flag to benchmark
- **added cli flag to use duckdb by default (this will work only for SELECT, not for any other statement)**
- fixed metastore metadata location issue after any data update (wrong link was stored in this field)
- added setup queries to duckdb connector to allow s3 access if the volume has s3 type

Some links
https://duckdb.org/docs/stable/core_extensions/iceberg/iceberg_rest_catalogs
https://duckdb.org/docs/stable/core_extensions/iceberg/overview.html
https://blog.greybeam.ai/query-snowflake-in-duckdb/#querying-time

```
cd crates/benchmarks
RESULTS_NAME=embucket USE_DUCKDB=false ./bench.sh run tpch10
RESULTS_NAME=duckdb USE_DUCKDB=true ./bench.sh run tpch10
./bench.sh compare duckdb embucket
```


crates/core-executor/src/query.rs
```diff
@@ -317,6 +322,24 @@ impl UserQuery {
         // 4. Single place to rewrite-optimize-adjust logical plan
         // etc
         if let DFStatement::Statement(s) = statement {
+            if self.session.get_session_variable("use_duck_db").is_some() {
```
osipovartem: cc @rampage644

osipovartem: Cc @camuel

DanCodedThis (approved):
@osipovartem LGTM, should we later add async query on top of this? What do you think?

crates/core-executor/src/query.rs
```diff
@@ -459,6 +471,49 @@ impl UserQuery {
...
+        let duckdb_pool = Arc::new(
+            DuckDbConnectionPool::new_memory().context(ex_error::DuckdbConnectionPoolSnafu)?,
+        );
+        let stream = get_stream(duckdb_pool, self.query.clone(), schema.clone())
+            .await
+            .context(ex_error::DataFusionSnafu)?;
+        let records = stream
+            .try_collect::<Vec<_>>()
+            .await
+            .context(ex_error::DataFusionSnafu)?;
```
DanCodedThis, line 507: Should this be a DF error? Or DuckDB error?

crates/benchmarks/src/tpch/run.rs
```diff
@@ -134,6 +135,11 @@ impl RunOpt {
         )
         .await?;
 
+        if self.common.use_duckdb {
+            let mut query = session.query("SET use_duck_db = true", query_context());
```
rampage644: Let's use something more aligned with how datafusion names those, like "datafusion.exection.something"?

Perhaphs "embucket.execution.acceleration" or smth?

crates/benchmarks/src/util/options.rs
```diff
@@ -71,6 +71,11 @@ pub struct CommonOpt {
     /// * `reorder_filters = true`
     #[structopt(long = "pushdown")]
     pub pushdown: bool,
+
+    /// If the session variable `use_duckdb` is set, we bypass `DataFusion` entirely
+    /// and execute the full SQL query directly using `DuckDB` in-memory engine.
+    #[structopt(long = "use_duckdb")]
+    pub use_duckdb: bool,
```
rampage644, line 78: I would also add it in our embucketd main cliOpts as well to be able to set it once in config if needed, default to false

crates/core-executor/Cargo.toml
```diff
@@ -17,13 +17,16 @@ aws-credential-types = { workspace = true }
 bytes = { workspace = true }
 chrono = { workspace = true }
 dashmap = { workspace = true }
+datafusion-federation = {version = "0.4.10"}
+datafusion-table-providers = { version = "0.8.1", features = ["duckdb"] }
```
rampage644, line 20: Double check if "duckdb-federation" feature is also needed here

crates/core-executor/Cargo.toml
```diff
@@ -17,13 +17,16 @@ aws-credential-types = { workspace = true }
...
 datafusion = { workspace = true }
 datafusion-common = { workspace = true }
 datafusion-doc = { workspace = true }
 datafusion-expr = { workspace = true }
 datafusion-functions-json = { workspace = true }
 datafusion-physical-plan = { workspace = true }
 datafusion_iceberg = { workspace = true }
+duckdb = { version = "=1.3.1", package = "spiceai_duckdb_fork" } # Forked to add support for duckdb_scan_arrow, pending: https://github.com/duckdb/duckdb-rs/pull/488
```
rampage644: nitpick - this is being blocked on deprecated DuckDB C API, version v.1.4 (that is out) should have more stable API. Once merged can switch to v1.4

crates/core-executor/src/query.rs
```diff
@@ -459,6 +471,49 @@ impl UserQuery {
...
+            .await?;
+        let plan = if let DFStatement::Statement(s) = statement.clone() {
+            self.sql_statement_to_plan(*s).await
+        } else {
+            ex_error::OnlySQLStatementsSnafu.fail()
+        }?;
+        let schema = plan.schema().inner().clone();
+
+        // Convert already resolved table references to iceberg_scan function call
+        self.update_statement_references(&mut statement, true)
```
rampage644: I would use dedicated function to update references instead of adding a flag for existing function.

crates/core-executor/src/query.rs
```diff
@@ -459,6 +471,49 @@ impl UserQuery {
...
+            .await?;
+        self.query = statement.to_string();
+
+        let duckdb_pool = Arc::new(
+            DuckDbConnectionPool::new_memory().context(ex_error::DuckdbConnectionPoolSnafu)?,
```
rampage644: Eventually, we need to decide if this is memory or file based connection pool. Maybe also switch to read only mode.

rampage644 (approved):

rampage644:
I also noticed this branch causes some weird regressions:

```
> use database s1;

Error occurred: 010001 (02000): 00000000-0000-0000-0000-077e6e679ecc: SQL compilation error: unsupported feature: Unsupported SQL statement: USE DATABASE s1
```

Also, when I execute the sql `SET use_duck_db = true` in the UI I get following error:
```
8239599687770: SQL compilation error: Failed to execute query.\nParser Error: syntax error at or near \"SET\"\n\nLINE 1: WITH fetch_schema AS (SET use_duck_db = 1) SELECT * FROM fetch_schema LIMIT 0\n 
```

crates/core-executor/src/query.rs
```diff
@@ -305,6 +308,15 @@ impl UserQuery {
     )]
     pub async fn execute(&mut self) -> Result<QueryResult> {
         let statement = self.parse_query().context(ex_error::DataFusionSnafu)?;
+
+        // If the session variable "use_duck_db" is set, we bypass DataFusion entirely
+        // and execute the full SQL query directly using DuckDB in-memory engine.
+        // This is typically used for heavy or complex queries that DataFusion handles poorly,
+        // such as large joins or subqueries, while preserving the same schema and query context.
+        if self.session.get_session_variable("use_duck_db").is_some() {
+            return self.execute_duck_db(statement).await;
```
Vedin, line 322: What about continue with DF execution on error? Do you plan to handle it in follow-up PR's?

crates/core-executor/src/query.rs
osipovartem: this is already datafusion error since we use DF SendableRecordBatch here

crates/core-executor/src/query.rs
osipovartem: We'll handle fallback behavior in a follow-up PR.
Currently, when use_duck_db is enabled, we don't expect any fallback to DataFusion (particularly for TPC-H queries).

crates/core-executor/Cargo.toml
osipovartem: removed

crates/core-executor/Cargo.toml
osipovartem: removed

crates/benchmarks/src/util/options.rs
osipovartem: added

crates/benchmarks/src/tpch/run.rs
osipovartem: renamed

crates/core-executor/src/query.rs
osipovartem: let's do this on the next PR after some discussion

DanCodedThis (approved):

rampage644 (approved):

DanCodedThis (approved):

DanCodedThis (approved):

Issue #1853, opened 2025-10-13
Bug: Number ranges (BETWEEN, >=/<=) include values outside the range


YevheniiNiestierov:
In Embucket, checking prices with a range like [23, 32] sometimes includes values that are clearly outside that range (e.g., 22.64 or 32.31).
This lets extra ITEM rows pass filters and changes the final result set (we get 39 rows instead of the correct 33 seen on Snowflake). Range filters return the wrong rows, s downstream joins/aggregates are wrong

### How to reproduce:
```
-- Expected: all FALSE
-- Embucket: TRUE, TRUE, TRUE, FALSE
select
  22.64 between 23 and 32                          as btw_2264,
  22.64 >= 23 and 22.64 <= 32                      as ge_le_2264,
  cast(22.64 as decimal(7,2)) between 23 and 32    as cast_btw_2264,
  round(22.64,2) between 23 and 32                 as round_btw_2264;
```
With [TPC-DS (scale factor=10) dataset](https://us-east-2.console.aws.amazon.com/s3/buckets/embucket-testdata?region=us-east-2&bucketType=general&prefix=tpcds/10/&showversions=false):
```
select
  count(case when i_current_price between 23 and 32 then 1 end) as betw_23_32,
  count(case when round(i_current_price,2) between 23 and 32 then 1 end) as betw_round2_23_32
from ITEM
where i_color in ('burlywood','coral','dark','puff','smoke','peach');
```
	Embucket: betw_23_32 = 88, betw_round2_23_32 = 81
	Snowflake: both = 81
So Embucket admits +7 extra items unless we ROUND(...,2)

**Looks like a numeric comparison/coercion bug in the predicate path (possibly FLOAT vs DECIMAL handling or constant folding**

DanCodedThis:
Should be closed by #1856

PR #1856, opened 2025-10-14
engine: Decimal cast fix

### LINKED ISSUE #1850: Bug: TPC-DS query with AND on joined tables produces incorrect row count
The issue occurs when a WHERE clause uses an AND operator to combine filters on columns from different tables in a JOIN. The engine appears to incorrectly evaluate the combined logical condition, leading to more rows being included in the result set than should be.

All tests were performed on an identical [TPC-DS (scale factor=10) dataset](https://us-east-2.console.aws.amazon.com/s3/buckets/embucket-testdata?region=us-east-2&bucketType=general&prefix=tpcds/10/&showversions=false), with Snowflake's results used as the correct baseline. 

### How to reproduce:
1 Go to infrastructure folder and follow [instructions](https://github.com/Embucket/embucket/tree/main/benchmark/infrastructure) to create instance for Embucket
2 Go to benchmark folder and upload needed dataset to Embucket and Snowflake
```
cd benchmark
python data_preparation.py --benchmark-type tpcds --system embucket --dataset-path tpcds/10
python data_preparation.py --benchmark-type tpcds --system snowflake --dataset-path tpcds/10
```
### Investigation:
Noticed the difference between Embucket and Snowflake outputs in [query13](https://github.com/Embucket/embucket/blob/b5c408a1e773540b8d562b4a85f2f8b02e48acd5/benchmark/tpcds/tpcds_queries.py#L783) After that isolated problematic logic to investigate. The following query returns 13681 rows in Embucket, whereas the correct result (from Snowflake) is 13564:
```
#"isolating faulty query"
select count(*)
from STORE_SALES, CUSTOMER_ADDRESS, DATE_DIM
where ss_sold_date_sk = d_date_sk and d_year = 2001
  and ss_addr_sk = ca_address_sk
  and ca_country = 'United States' and ca_state in ('MI', 'MN', 'NC')
  and ss_net_profit between 100 and 200;
```

Identifying a Specific Incorrect Row: found a specific row that Embucket includes but should be excluded. The primary key is (ss_item_sk= 69047, ss_ticket_number= 83794)
```
select
    s.ss_item_sk,
    s.ss_ticket_number,
    c.ca_state,
    s.ss_net_profit
from
    STORE_SALES s
join
    CUSTOMER_ADDRESS c on s.ss_addr_sk = c.ca_address_sk
where
    s.ss_item_sk = 69047
    and s.ss_ticket_number = 83794;
```
This query returns:
ss_item_sk: 69047, ss_ticket_number: 83794, ca_state: 'MN', ss_net_profit: 99.63

The logic from the "isolating faulty query" is ca_state IN ('MI', 'MN', 'NC') AND ss_net_profit BETWEEN 100 AND 200. 
Therefore: 
ca_state = 'MN' evaluates to TRUE.
ss_net_profit = 99.63 evaluates to FALSE.

**The combined expression TRUE AND FALSE must evaluate to FALSE. Therefore, this row should be excluded from the result set. The correct count should be 13564.**
**Looks like Embucket incorrectly evaluates the AND condition as true for this row (and 118 others like it) and includes it in the result set. This leads to an incorrect total count of 13681.**

LLM's comment:
_"The root cause appears to be a bug in the evaluation of the AND logical operator. The bug specifically manifests when the predicates connected by AND apply to columns from different tables that are part of a JOIN (in this case, STORE_SALES.ss_net_profit and CUSTOMER_ADDRESS.ca_state).The engine seems to fail to correctly apply the second condition after the first one has been evaluated."_

DanCodedThis:
Closes #1850 & #1853

- Fixed incorrect truncation in decimal to decimal cast
- Using `to_decimal` internally and then casting to the correct decimal precision and scale resulted in all decimal values truncated to int like values
- Using `to_decimal` with the correct precision and scale fixes it, also no need for additional casting
- Added insta tests
- Related to tps ds q13
- Regression testing:
  - [x] **dbt-gitlab** works (93.4%)
  - [x] **dbt-snowplow** works (100%)

P.S. cc @YevheniiNiestierov

### DIFF
--- a/crates/core-executor/src/datafusion/logical_analyzer/cast_analyzer.rs
+++ b/crates/core-executor/src/datafusion/logical_analyzer/cast_analyzer.rs
@@ -105,7 +105,7 @@ impl CastAnalyzer {
                     args: vec![expr.clone()],
                 })))
             }
-            data_type @ (DataType::Decimal128(_, _) | DataType::Decimal256(_, _)) => {
+            data_type @ DataType::Decimal128(_, _) => {
                 Self::rewrite_numeric_cast(expr, data_type, try_mode)
             }
             data_type @ (DataType::Int32 | DataType::Int64)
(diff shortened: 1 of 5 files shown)


crates/core-executor/src/datafusion/logical_analyzer/cast_analyzer.rs
```diff
@@ -105,7 +105,7 @@ impl CastAnalyzer {
                     args: vec![expr.clone()],
                 })))
             }
-            data_type @ (DataType::Decimal128(_, _) | DataType::Decimal256(_, _)) => {
+            data_type @ DataType::Decimal128(_, _) => {
```
osipovartem, line 108: why Decimal256 skipped?

DanCodedThis: Decided to remove since we don't support such values anyway. Better to add back?

DanCodedThis: I originally added it as a precaution, but I think it's not the place for that. And it didn't & shouldn't do anything anyway.

YaroslavLitvinov (approved):

PR #1860, opened 2025-10-15
Check query by EXPLAIN before execution

### LINKED ISSUE #1859: EXPLAIN query before calling duckdb

osipovartem:
- added check to pass on SELECT queries to duckdb
- added to logic to call EXPLAIN before calling duckdb, covered by session variable 
Closes #1859

### DIFF
--- a/crates/core-executor/src/query.rs
+++ b/crates/core-executor/src/query.rs
@@ -169,6 +169,13 @@ impl UserQuery {
         Ok(statement)
     }
 
+    pub fn statement(&self) -> std::result::Result<DFStatement, DataFusionError> {
+        let state = self.session.ctx.state();
+        let dialect = state.config().options().sql_parser.dialect.as_str();
+        let statement = state.sql_to_statement(&self.raw_query, dialect)?;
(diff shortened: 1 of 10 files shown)


crates/core-executor/src/query.rs
```diff
@@ -486,21 +500,31 @@ impl UserQuery {
     pub async fn execute_duck_db(&mut self, mut statement: DFStatement) -> Result<QueryResult> {
         // Fully qualify all table references
         self.update_statement_references(&mut statement)?;
-        let plan = self.statement_to_plan(&statement).await?;
-        let schema = plan.schema().inner().clone();
```
DanCodedThis, line 490: Why we don't use the schema anymore? Does explain produce an accurate schema?

DanCodedThis (approved):
@osipovartem LGTM

crates/core-executor/src/query.rs
osipovartem: this schema is not used in query_arrow
we allways call conn.prepare to get real schema.
So the schema is not required there

PR #1872, opened 2025-10-22
Sqlite for query history

YaroslavLitvinov:
### :hammer_and_wrench: This PR includes
* For Sqlite (requires clang 18) it enables journal_mode=wal, busy_timeout=2000 (Not sure last is needed)
* SlateDBHistoryStore based on Sqlite, enabled by "sqlite" feature in `core-history` crate 
  - queries and results are separate databases
  - QueryRecord doesn't include result data anymore
  - ResultSet used for saving, reading result data from database.
* Added query result handler + updated openapi spec for api-ui. It returns serialized ResultSet
* query history mostly not used directly in api-ui, instead it used exisitng approach "aka" SlateDBView, when handling api-ui requests
* For some tests enabled multi_threaded runtime as these test there flaking
* Increased granularity of QueryRecordId from milis to microseconds
* traces.log added for tests in `api-snowflake-rest`

### :memo: Also worth noting 
* Sqlite writes query history just to the relative dir `sqlite_data/` in filesystem
* tests, checks workflows have additional clang-18 deps
* Sqlite Vfs is gated by "vfs" feature (not used currently)
  * It includes sqlite-plugin, and derived implementation of [sqlite vfs - s3qlite](https://github.com/maxmcd/s3qlite)
  * LIBSQLITE3_FLAGS in .cargo/config.toml affects only Sqlite with "vfs" enabled
  * Do not enable vfs feature
* Embucket seed build disabled as it takes more time to compile (Let me know if anybody uses it)


crates/core-sqlite/src/vfs/mod.rs
```diff
@@ -0,0 +1,797 @@
+mod handle;
```
rampage644, line 1: Why vfs is still needed? Still running over slatedb?

crates/api-ui/src/tests/queries.rs
```diff
@@ -401,13 +416,18 @@ async fn test_ui_async_query_infer_default_exec_mode() {
...
+    http_req::<()>(
+        &client,
+        Method::GET,
+        &format!("http://{addr}/ui/queries/{}/result", query_record.id),
+        String::new(),
+    )
+    .await
+    .expect_err("Get query error");
+
+    std::thread::sleep(std::time::Duration::from_millis(1000));
```
rampage644, line 430: I realize this was in the test previosly, but is intended to use std::sleep vs tokio::sleep? To emulate blocking?

crates/core-sqlite/src/vfs/mod.rs
YaroslavLitvinov: this feature is gated and currently off, I don't know maybe we will find this helpful in future.

crates/api-ui/src/tests/queries.rs
YaroslavLitvinov: > I realize this was in the test previosly, but is intended to use std::sleep vs tokio::sleep? To emulate blocking?

This is just a way to wait certain period of time, though std::thread::sleep is a blocking call, but it is not harmfull for test that uses tokio flavor=multi_thread and just PASS. Yes may use ` tokio::sleep` in this particular case.

crates/core-history/src/entities/query.rs
```diff
@@ -27,6 +27,32 @@ impl Display for QueryStatus {
...
+}
+
+impl std::error::Error for ParseQueryStatusError {}
+
+impl FromStr for QueryStatus {
+    type Err = ParseQueryStatusError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "Running" => Ok(Self::Running),
```
DanCodedThis, line 46: There is no "Not started" right?

crates/core-history/src/entities/result_set.rs
```diff
@@ -123,10 +149,9 @@ impl ResultSet {
         level = "debug",
         fields(batch_size_bytes=self.batch_size_bytes, rows_count=self.rows.len(), serialized_size, rows_count_limit),
     )]
-    pub fn serialize_with_limit(
-        &self,
-        max_rows_limit: usize,
-    ) -> (std::result::Result<String, serde_json::Error>, usize) {
+    pub fn serialize_with_limit(&self) -> (std::result::Result<String, serde_json::Error>, usize) {
+        let max_rows_limit = self.configured_rows_limit.unwrap_or(usize::MAX);
```
DanCodedThis, line 153: Reasonable default?

crates/core-history/src/sqlite_history_store.rs
```diff
@@ -0,0 +1,786 @@
...
+
+const WORKSHEET_ADD: &str = "
+INSERT INTO worksheets (id, name, content, created_at, updated_at)
+    VALUES (:id, :name, :content, :created_at, :updated_at);
+";
+
+pub struct SlateDBHistoryStore {
+    pub queries_db: SqliteDb,
+    pub results_db: SqliteDb,
+}
```
DanCodedThis, line 59: SqliteHistoryStore?

crates/core-sqlite/src/vfs/sqlite-plugin/build.rs
```diff
@@ -0,0 +1,32 @@
+extern crate bindgen;
+
+use std::env;
+use std::path::PathBuf;
+
+fn main() {
```
DanCodedThis, line 6: Plugin? Second main?

crates/core-history/src/sqlite_history_store.rs
```diff
@@ -0,0 +1,786 @@
...
+                        status,
+                        error,
+                        diagnostic_error
+                    FROM queries
+                    WHERE {} id > :cursor
+                    ORDER BY start_time DESC
+                    LIMIT :limit",
+                    worksheet_id
+                        .map_or_else(String::new, |wid| format!("worksheet_id = {wid} AND"))
+                ))?;
```
DanCodedThis, line 651: We should definitely connect this in the df catalog later to allow seamless querying from `snow-sql`, and from `ui/query` endpoint

DanCodedThis (approved):
@YaroslavLitvinov Overall LGTM, didn't look into vfs that much since I presumed it is/will be disabled for now

crates/core-history/src/entities/query.rs
YaroslavLitvinov: We have no "NotStarted" state, but this code was just moved to the `from` trait.

crates/core-history/src/entities/result_set.rs
YaroslavLitvinov: Max here means that there are no actual limit

DanCodedThis: I meant perhaps we should set/find a reasonable default. Maybe in the future

crates/core-history/src/sqlite_history_store.rs
YaroslavLitvinov: yes, here we have not very flexible name for metastore object used across the project. This PR just adds new trait implementation.

crates/core-sqlite/src/vfs/sqlite-plugin/build.rs
YaroslavLitvinov: Yep dependency has own build script

crates/core-history/src/sqlite_history_store.rs
YaroslavLitvinov: @DanCodedThis It is still queryable, as before, from snow and from ui. Or what do you mean about connect this in df?

DanCodedThis: I assume we can't query it from the execution service `.query()`, since it needs explicit connection in the df-catalog (we need to create views which are based on the sqlite). However, we can use the History trait methods since they call sqlite inside. I meant if we query from snow sql we can't query history store, since it's won't be connected like that. Not sure if it's an important feature just a thought for the future.

PR #1874, opened 2025-10-22
Register some DF UDF functions in duckdb

osipovartem:
- added logic to register the following UDF funcs in duckdb
  - String binary
  - Regexp
  - Contitional
  - Crypto
  - Datetime
  - Numeric
  - System
- ```Some UDFs were not registered/overloaded in DuckDB: [\"length\", \"lower\", \"split\", \"substr\", \"replace\", \"regexp_replace\", \"md5\", \"datediff\", \"dayname\", \"last_day\", \"monthname\", \"year\", \"day\", \"dayofmonth\", \"dayofweek\", \"dayofyear\", \"week\", \"weekofyear\", \"month\", \"quarter\", \"hour\", \"minute\", \"second\"]```
- Removed float16 from parse_ip since this type is not supported by duckdb types convertor

**Next steps is to** 
- try to fix funs overloading
- check slt tests
- add session udfs like current_ip, etc.

### DIFF
--- a/crates/core-executor/src/duckdb/functions.rs
+++ b/crates/core-executor/src/duckdb/functions.rs
@@ -2,30 +2,76 @@ use crate::error::{self as ex_error, Result as CoreResult};
 use arrow_schema::{DataType, Field, FieldRef};
 use datafusion::arrow::array::Array;
 use datafusion::arrow::record_batch::RecordBatch;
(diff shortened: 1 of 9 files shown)


DanCodedThis (approved):
@osipovartem Preliminary apporve, if we decided to merge it later

PR #1877, opened 2025-10-23
Metastore on sqlite

YaroslavLitvinov:
### Partial implementation of Metastore on sqlite
* Volumes, databases, schemas are migrated on SQLITE (no vfs used)
* We use diesel orm to map mentioned ^^ entities to sql.
* Tables remained on slatedb, at least temporary.
* `ExecutionService::abort_query` turned into async function as now it waits for abort confirmation

# Known issues
* UI should support some minor API changes in responses. Also didn't fix openapi spec, yet.
* Slatedb sql view we use in execution engine renamed to sqlite view, and maybe need some fixes.
* List of failing tests, probably due to some minor bugs in new Metastore implementation. Fixing them is a matter of time.
    tests::test_schemas
    tests::service::test_max_concurrency_level
    tests::service::test_query_recording
    tests::sql::commands::show::query_show_databases
    tests::sql::ddl::database::query_drop_database
    tests::sql::ddl::database::query_drop_database_error_in_use
    tests::test_generic_sqls::snowflake_generic::submit_ok_query_with_concurrent_limit
    tests::test_rest_quick_sqls::snowflake_compatibility::abort_ok_query
    tests::test_rest_quick_sqls::snowflake_compatibility::cancel_ok_query
    tests::test_rest_quick_sqls::snowflake_compatibility::cancel_ok_sleeping_query


crates/core-metastore/src/sqlite/diesel_gen.rs
```diff
@@ -0,0 +1,47 @@
+// @generated automatically by Diesel CLI.
+diesel::table! {
+    databases (id) {
+        id -> Text,
+        ident -> Text,
+        properties -> Nullable<Text>,
+        volume_ident -> Text,
```
rampage644: I think this can also be migrated to volume_id? Or better postpone such changes to the next PR?

crates/core-metastore/src/sqlite_metastore.rs
```diff
@@ -0,0 +1,1035 @@
...
+    Table,
+}
+
+///
+/// tbl/<db>/<schema> -> List of tables for <schema> in <db>
+/// tbl/<db>/<schema>/<table> -> `Table`
+///
+const KEY_TABLE: &str = "tbl";
+
+pub struct SlateDBMetastore {
```
rampage644, line 66: Sqlite or SlateDB metastore?

YaroslavLitvinov: Slatedb still used for tables

crates/api-ui/src/volumes/models.rs
```diff
@@ -117,20 +118,23 @@ pub struct VolumeResponse(pub Volume);
 #[derive(Debug, Clone, Serialize, Deserialize, ToSchema)]
 #[serde(rename_all = "camelCase")]
 pub struct Volume {
+    pub id: i64,
```
YaroslavLitvinov, line 121: this `id: i64` should be removed, as `id` is exposed at RwObject

YaroslavLitvinov: No, since this is api-ui, RwObject is not applicable. So that is ok.

PR #1891, opened 2025-10-30
engine: Explain fix

### LINKED ISSUE #1890: engine: `EXPLAIN` doesn't use visitors or references updating
- Explain is a special Datafusion statement not part of the normal DF statement
- Need to make sure that all vistors are used in the query from explain
- Need to make sure that reference updating works and resolves to correct tables
- Insta tests

DanCodedThis:
Closes #1890

- Fixed explain to use visitors  
- Explain should be able to see tables which are part of the query context
- Added insta tests

P.S. So our toolchain clippy was on 1.91, and I was on 1.90, even though I made no changes there, with 1.90 or 1.91 it failed clippy for me, so I fixed all of the clippy issues and also fmt diff issue (very strange behavior) we had.


rampage644 (approved):

DanCodedThis:
Flaky tests with different paths to parquet files

DanCodedThis:
So the tests failed, because there is a difference on how many "-" symbols for pretty batch print it prints :(

DanCodedThis:
There is a test with record batch conversion, specifically timestaps - where it fails, although I didn't touch it. I checked on recent **main** the test with record batch conversion fails on timestamps there also, so our test suite doesn't run on them, or I don't know

DanCodedThis:
> There is a test with record batch conversion, specifically timestaps - where it fails, although I didn't touch it. I checked on recent **main** the test with record batch conversion fails on timestamps there also, so our test suite doesn't run on them, or I don't know

If no other test fails, we can merge this

osipovartem (approved):

PR #1868, opened 2025-10-20
WIP udf duckdb func registration

osipovartem:
### **Summary**

Integrates native DuckDB UDF registration using the latest raw `duckdb` crate (≥ 1.4.x), without relying on `df-table-providers`.

### **Details**

* Added `DfUdfWrapper<T>` — a generic adapter exposing any DataFusion `ScalarUDFImpl` as a DuckDB `VArrowScalar` function.
* Implemented `register_all_udfs(conn)` to register core Embucket UDFs (e.g. `length_test`).
* Switched to DuckDB’s new Arrow streaming interface (`query_arrow`, `stream_arrow`) for result retrieval.
* Enabled `EXPLAIN (format html)` execution via `spawn_blocking` for thread-safe DuckDB calls.
* Removed dependency on `datafusion-table-providers` for UDF interoperability (duckdb from spiceai fork 1.3.1).

### **Impact**

Allows running Embucket’s custom UDFs natively inside DuckDB’s in-memory engine, improving performance and simplifying the accelerated execution path.

### **Next**

* Try to register UDF functions

Example implemented (registered length_test func to call DF length func)
```
select length_test('fffff')
   ;
+----------------------+
| length_test('fffff') |
|----------------------|
| 5                    |
+----------------------+
```

PR #1905, opened 2025-11-05
Revert duckdb related code

osipovartem:
- **Revert "WIP udf duckdb func registration (#1868)"**
- **Revert "Add ability to execute the full SQL query directly using DuckDB in-memory engine (#1845)"**


crates/core-executor/src/query.rs
```diff
@@ -172,13 +165,6 @@ impl UserQuery {
         Ok(statement)
     }
 
-    pub fn statement(&self) -> std::result::Result<DFStatement, DataFusionError> {
```
rampage644, line 175: I speculate this is not used method, am I right?

rampage644 (approved):

crates/core-executor/src/query.rs
osipovartem: yes, it was added for duckdb

PR #1911, opened 2025-11-06
Fix variant_element double quotes for strings 

osipovartem:
The issue was caused by using Value::to_string() on JSON String values.
When `Value::String("test")` is serialized via `to_string()`, it becomes `"\"test\""`, because the JSON serializer escapes the inner string.
For flatten = true, we now detect Value::String and extract the raw inner string instead of serializing it as JSON.


DanCodedThis (approved):
@osipovartem LGTM

rampage644 (approved):

PR #1912, opened 2025-11-06
Add bearer token authorization check

JanKaul:
## Summary

Add bearer token authorization for the Iceberg REST API endpoint. When configured, incoming requests are validated against the bearer token. Requests without proper authorization return 401 Unauthorized. If no token is configured, all requests are allowed through.

## Changes

- Add `axum-auth` dependency for middleware support
- Implement `require_auth` middleware to validate Bearer tokens
- Add `bearer_token` field to the Iceberg REST API state
- Add `iceberg_rest_bearer_token` CLI option (supports ICEBERG_REST_BEARER_TOKEN env var)
- Integrate auth middleware into the Iceberg REST router

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -228,6 +228,7 @@ name = "api-iceberg-rest"
 version = "0.1.0"
 dependencies = [
  "axum 0.8.6",
+ "axum-auth",
  "core-metastore",
  "core-utils",
  "error-stack",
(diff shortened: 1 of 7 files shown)


DanCodedThis (approved):
@JanKaul How do you authorize? Or do you just send the token with each request manually, no log in? Otherwise, LGTM

PR #1914, opened 2025-11-07
core embucket

YaroslavLitvinov:
Metastore on sqlite + inmemory for tables
No core-utils, No Slatedb


YaroslavLitvinov:
Tests stuck, I think here is a deadlock.
test tests::dashboard::test_ui_dashboard has been running for over 60 seconds
test tests::navigation_trees::test_ui_databases_navigation has been running for over 60 seconds
test tests::tables::test_ui_tables has been running for over 60 seconds

PR #1917, opened 2025-11-08
engine: Incorrect status message fix

### LINKED ISSUE #1916: engine: Incorrect status response with `arrow` format
- When arrow format used, snowflake cli seg-faults
- When using `USE` or `SET` statements, or any other statements that produce an empty null array in **arrow** format.
- Should send a status column with a message ` Statement executed successfully.`
- With `json` format, snowflake cli just got `No data`. Because it sent an empty array.

DanCodedThis:
Closes #1916

- Before for `arrow` we sent an empty null array. Which resulted in a seg-fault of the snowflake cli
- For `json` we sent `No data` - an empty array.
- Both were incorrect, changed to one `status` column with a successful execution message
- Added insta tests
- Checked other statements which use the status response and fixed their snapshots also

P.S. `Alter` & `Use` don't check if the variable exists (only the type I assume), so it always says it is successful, it did that before, but just didn't send the message. `Drop` should check. `Set` is probably infallible, either, then sql "comp" errors.


Vedin (approved):

crates/core-executor/src/query.rs
```diff
@@ -2754,7 +2754,15 @@ impl UserQuery {
             false,
         )]));
         Ok(QueryResult::new(
-            vec![RecordBatch::new_empty(schema.clone())],
+            vec![
+                RecordBatch::try_new(
```
YaroslavLitvinov, line 2758: I actually think that this should not be fixed query executor modifying QueryResult. This  "Statement executed ..." message that was added looks like belongs to a rest api. It naturally belongs to the snowflake-rest.

YaroslavLitvinov: Also I believe that such change should be covered by a separate test in snowlake-rest as well

DanCodedThis: We should probably do that later, I agree

PR #1919, opened 2025-11-11
engine: Service timeout shutdown

### LINKED ISSUE #1896: Exit the process after X minutes of inactivity
Add a setting that monitors if there is any usage and after X minutes/seconds of inactivity gracefully finishes the application.

DanCodedThis:
Closes #1896

- Adds new cli option: `IDLE_TIMEOUT_SECONDS` with default value of `18000` (5 hours)
- Adds new `timeout_signal` fn is a execution service
- The function is done similar to the session expiration and other shutdown signals
- The function is then used in tokio select of graceful shutdown of the `axum::serve` in main
- Adds an expiration time test
- Interval and idle timeout now are different numbers, to avoid bugs written below

### DIFF
--- a/crates/core-executor/src/service.rs
+++ b/crates/core-executor/src/service.rs
@@ -19,7 +19,7 @@ use std::sync::atomic::Ordering;
 use std::vec;
 use std::{collections::HashMap, sync::Arc};
 use time::{Duration as DateTimeDuration, OffsetDateTime};
-
+use tokio::io::AsyncWriteExt;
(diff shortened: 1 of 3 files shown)


DanCodedThis:
The timer check interval and the idle timeout should be two different numbers, where if one queries some time ago and then on the second looses his last session when the check happens it shouldn't shut down since the use was recent

osipovartem (approved):

